PR #26, opened 2017-05-23
Fix hookup

david-deboer:


hera_mc/args_test.py
```diff
@@ -0,0 +1,53 @@
+#! /usr/bin/env python
+# -*- mode: python; coding: utf-8 -*-
+# Copyright 2016 the HERA Collaboration
+# Licensed under the 2-clause BSD license.
+
+"""This is meant to hold utility scripts for handling parts and connections
+
+"""
```
bhazelton: Can you explain what this file does a little better. It says it's supposed to hold "utility scripts" but I just see a bunch of arg parser stuff. I can see how that could be reused but you don't seem to be using it in any of the other changes on this branch.

hera_mc/data/HERA_350.txt
```diff
@@ -318,31 +318,31 @@ HH8	541018.40	6601070.74	1052.64
 HH9	541033.00	6601070.74	1052.52
 HH10	541047.60	6601070.74	1052.61
 HH22	541062.20	6601079.17	1052.45
-HA320	540762.90	6601530.14	1048.22
-HA321	540908.90	6601530.14	1048.44
-HA322	541054.90	6601530.14	1048.79
-HA323	541200.90	6601530.14	1049.28
-HA324	540689.90	6601412.13	1049.05
+HB320	540762.90	6601530.14	1048.22
```
bhazelton: What is the significance of switching from HA to HB here? I don't actually know what either mean.

hera_mc/mc.py
```diff
@@ -166,6 +166,7 @@ def __init__(self, db_url):
 
     def create_tables(self):
         """Create all M&C tables"""
+        print(self.engine)
```
bhazelton: Is this just for debugging, or is it generally useful to print this?

scripts/parts.py
```diff
@@ -30,8 +30,9 @@
                         dest='mapr_cols', default='all')
     parser.add_argument('--levels-testing', help="Set to test filename if correlator levels not accessible [levels.tst]", 
                         dest='levels_testing', default='levels.tst')
-    parser.add_argument('--date', help="MM/DD/YY or now [now]", default='now')
-    parser.add_argument('--time', help="hh:mm or now [now]", default='now')
+    parser.add_argument('--date', help="UTC YYYY/M/D or now [now]", default='now')
+    parser.add_argument('--time', help="UTC hh:mm or now [now]", default='now')
```
bhazelton: This is the 3rd place these exact lines are added. I also see them in args_test.py. I freely admit that I'm not sure how this all works, but can you reduce the code replication by using args_test here?

hera_mc/args_test.py
david-deboer: I didn't change the comment - it is just meant to be an import module to allow testing of cm_handling and cm_hookup within python

hera_mc/data/HERA_350.txt
david-deboer: The inner ring of outliers is ring A.  The outer ring is B

hera_mc/mc.py
david-deboer: debugging - could be deleted

scripts/parts.py
david-deboer: argtest is just meant to be for testing.  every script will get this

scripts/parts.py
```diff
@@ -50,10 +51,11 @@
     args.revision = args.revision.upper()
     if args.revision == 'ALL':
         args.active = False
-    if args.levels_testing.lower() == 'none' or args.levels_testing.lower() == 'false':
-        args.levels_testing = False
-    elif args.levels_testing == 'levels.tst':
-        args.levels_testing = os.path.join(mc.test_data_path, 'levels.tst')
+    if type(args.levels_testing) == str:
```
pkgw: Why is this `type()` test needed? Is `args.levels_testing` equal to `None` if it's not specified? If so the test should be more specific: `if args.levels_testing is not None:`

hera_mc/args_test.py
pkgw: It is not clear to me what this module is for either. Please elaborate in the docstring.

hera_mc/mc.py
pkgw: Please delete it.

scripts/parts.py
pkgw: If many scripts are going to have these kinds of similar/identical arguments, it might be helpful to add a utility module that has functions like `add_datetime_args(parser)` to reduce the code duplication.

PR #30, opened 2017-06-02
Args cleanup

david-deboer:
This is an unknown, trivially different branch that exists because of my transition to doing things this way.  I'd like to just get it rebased with master to confidently march into the future with a clean slate...


bhazelton:
This now looks very reasonable. I'll merge it.

PR #32, opened 2017-06-04
Add RTP & Librarian tables

bhazelton:
These tables all have add_* and get_* methods and are fully covered by tests. The tables are also documented in docs/mc_definition.tex.

This PR also has some significant code reworking including:
- move MCSession object to it's own file
- define add_* methods on MCSession so add & get methods are on the same object
- define standard time filtering method used by get_* methods on MCSession
- define standard __eq__ and __repr__ methods on MCDeclarativeBase
- __eq__ method supports precision information per column for numeric types. these are defined in the table definition
- changed DOUBLE_PRECISION column types to Float because Float is mapped to DOUBLE PRECISION by default for postgresql
- defined all times in new tables in gps seconds rather than datetimes (this should be propagated to other tables)

### DIFF
--- a/hera_mc/librarian.py
+++ b/hera_mc/librarian.py
@@ -8,9 +8,10 @@ Librarian tables
 The columns in this module are documented in docs/mc_definition.tex,
 the documentation needs to be kept up to date with any changes.
 """
+from math import floor
 from astropy.time import Time
 from sqlalchemy import Column, ForeignKey, Integer, BigInteger, String, Text, Float
-from . import MCDeclarativeBase, DEFAULT_GPS_TOL, DEFAULT_MIN_TOL
+from . import MCDeclarativeBase, DEFAULT_MIN_TOL
 from .server_status import ServerStatus
 
 
@@ -22,7 +23,7 @@ class LibStatus(MCDeclarativeBase):
     """
     Definition of lib_status table.
 
-    time: time of this status in gps seconds (double). Primary_key
+    time: time of this status in floor(gps seconds) (BigInteger). Primary_key
     num_files: number of files in librarian (BigInteger)
     data_volume_gb: data volume in GB (Float)
     free_space_gb: free space in GB (Float)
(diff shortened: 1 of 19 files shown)


hera_mc/server_status.py
```diff
@@ -0,0 +1,91 @@
...
+        if not isinstance(system_time, Time):
+            raise ValueError('system_time must be an astropy Time object')
+        system_time = system_time.utc.gps
+
+        return cls(hostname=hostname, mc_time=mc_time, ip_address=ip_address,
+                   system_time=system_time, num_cores=num_cores,
+                   cpu_load_pct=cpu_load_pct, uptime_days=uptime_days,
+                   memory_used_pct=memory_used_pct, memory_size_gb=memory_size_gb,
+                   disk_space_pct=disk_space_pct, disk_size_gb=disk_size_gb,
+                   network_bandwidth_mbs=network_bandwidth_mbs)
```
pkgw: It will probably end up being helpful to provide a function that can calculate all of this information for the machine that the code is running on. I think that can wait until the future, though.

hera_mc/librarian.py
```diff
@@ -0,0 +1,244 @@
...
+
+    time: time of this status in gps seconds (double). Primary_key
+    num_files: number of files in librarian (BigInteger)
+    data_volume_gb: data volume in GB (Float)
+    free_space_gb: free space in GB (Float)
+    upload_min_elapsed: minutes elapsed since last file upload (Float)
+    num_processes: number of background tasks running (Integer)
+    git_version: librarian git version (String)
+    git_hash: librarian git hash (String)
+    """
```
bhazelton: I added this in the comment at the top of the file.

pkgw:
@dannyjacobs Do you have any (further?) comments on this? I think we should go ahead and merge if not.

bhazelton:
@pkgw there are a couple of your points above that I wanted to discuss in a group.

PR #33, opened 2017-06-14
Centralize code to define some common CM-related command-line arguments

pkgw:
Namely `--verbosity`, `--date`, and `--time`. The `--show-active` and `--show-all` options might also make sense to centralize but it might be worth thinking about defining them in a different way.


pkgw:
@bhazelton or @david-deboer Can we get a look at this when someone has the chance? This has been languishing even though it's quite simple.

PR #39, opened 2017-06-19
Cm convert2gps

david-deboer:
Although bugs still persist, it impacts things so fundamentally that I'd like to merge it into master and start a clean up branch that will also fully incorporate the args changes.  This will allow the new cm gps schema.


bhazelton:
It looks like the tests aren't passing.

david-deboer:
Recall that it needs the new schema for the database, and so checks won't pass unless you blast and rebuild the db (and then you've blasted and rebuilt the db...)

bhazelton:
For the testing that runs on Travis, the database is deleted and re-made every time. So we should be able to get testing to work on Travis.

david-deboer:
It now passes nosetests.

hera_mc/cm_utils.py
```diff
@@ -9,15 +9,17 @@
...
 
 from hera_mc import mc
-import datetime
+from astropy.time import Time
+from astropy.time import TimeDelta
 import os.path
-import pytz
 
+PAST_DATE = '2000-01-01'
+FUTURE_DATE = '2025-12-31'
```
bhazelton: I'm not sure I'm following all this logic, but are you switching from using NANs for times in the future (i.e. if they don't have stop times) to using this Dec 31, 2025 date?

hera_mc/geo_location.py
```diff
@@ -69,9 +70,12 @@ class GeoLocation(MCDeclarativeBase):
     elevation = Column(Float)
     "Elevation in m"
 
-    created_date = NotNull(DateTime(timezone=True))
+    created_gpstime = NotNull(BigInteger)
     "The date when the station assigned by project."
 
+    def gps2Time(self):
```
bhazelton: If this method is just making a new convenience attribute on the class, why not call it in the __init__ method? You can write a custom __init__ that calls the parent class __init__ like this `super(UVData, self).__init__()` and then add anything you'd like for your custom class.

hera_mc/geo_location.py
```diff
@@ -208,7 +212,7 @@ def is_in_connections(args, station_name, check_if_active=False):
     ------------
     args:  needed arguments to open database and set date/time
     station_name:  string name of station
-    check_if_active:  boolean flag to check if active
+    check_if_active:  either False to not check/care if active, or astropy Time to check
```
bhazelton: the name of this parameter is confusing and the mixed type (boolean or time) is weird. Why not use something that has time in its name and if it's none, not care if it's active?

hera_mc/part_connect.py
```diff
@@ -43,15 +45,22 @@ class Parts(MCDeclarativeBase):
...
     "The date when the part was installed (or otherwise assigned by project)."
 
-    stop_date = Column(DateTime(timezone=True))
+    stop_gpstime = Column(BigInteger)
     "The date when the part was removed (or otherwise de-assigned by project)."
 
     def __repr__(self):
         return '<heraPartNumber id={self.hpn}{self.hpn_rev} type={self.hptype}>'.format(self=self)
 
+    def gps2Time(self):
```
bhazelton: again, think about adding this to the init

hera_mc/part_connect.py
```diff
@@ -212,6 +222,9 @@ class PartInfo(MCDeclarativeBase):
     def __repr__(self):
         return '<heraPartNumber id = {self.hpn} comment = {self.comment}>'.format(self=self)
 
+    def gps2Time():
```
bhazelton: again, think about adding this to the init

hera_mc/part_connect.py
```diff
@@ -247,17 +260,24 @@ class Connections(MCDeclarativeBase):
...
-    stop_date = Column(DateTime(timezone=True))
+    stop_gpstime = Column(BigInteger)
     "stop_time is the time that the connection is removed"
 
     def __repr__(self):
         up = '{self.upstream_part}:{self.up_part_rev}'.format(self=self)
         down = '{self.downstream_part}:{self.down_part_rev}'.format(self=self)
         return '<{}<{self.upstream_output_port}$|${self.downstream_input_port}>{}>'.format(up, down, self=self)
 
+    def gps2Time(self):
```
bhazelton: again, think about adding this to the init

hera_mc/tests/test_connections.py
```diff
@@ -40,6 +40,5 @@ def test_commit_part(self):
         print(part)
         self.test_session.add(part)
         self.test_session.commit()
+        #self.assertTrue
```
bhazelton: should this be commented? If so, delete it, otherwise uncomment it.

hera_mc/geo_handling.py
```diff
@@ -229,7 +238,7 @@ def plot_stations(args, stations_to_plot, fignm, marker_color='g', marker_shape=
                                      xytext=(pt[coord[args.xgraph]] + 2, pt[coord[args.ygraph]]))
 
 
-def plot_station_types(args, label_station=False):
+def plot_station_types(args, label_station=False, query_date=False):
```
bhazelton: again, I suggest None rather than False

hera_mc/geo_handling.py
```diff
@@ -168,19 +177,19 @@ def get_all_locations(args):
...
-        for a in session.query(geo_location.GeoLocation).filter(geo_location.GeoLocation.created_date >= dt):
+        for a in session.query(geo_location.GeoLocation).filter(geo_location.GeoLocation.created_gpstime >= dt):
             found_stations.append(a.station_name)
     return found_stations
 
 coord = {'E': 'easting', 'N': 'northing', 'Z': 'elevation'}
 
 
-def plot_stations(args, stations_to_plot, fignm, marker_color='g', marker_shape='o', marker_size='8', label_station=False):
+def plot_stations(args, stations_to_plot, fignm, query_date=False, marker_color='g', marker_shape='o', marker_size='8', label_station=False):
```
bhazelton: I think you should use None for an undefined query date rather than False so you don't end up with mixed types.

david-deboer:
Thanks Bryna,

For gps2Time, basically I’d like that everytime is pulls a record from the db that it converts gps to Time.  I wasn’t sure how to do that.  If easy, can you give me specific code.  If hard, can you videocon?

I’ll make the other changes or comment in git/GitHub

Thanks,
Dave

hera_mc/cm_utils.py
david-deboer: Correct - this was just a simple way to make sure you bracket the project with dates.

hera_mc/tests/test_connections.py
david-deboer: I have no idea on this one  -- not sure when/why/where it happened.  I'll uncomment it for now, but will be fixing/adding tests in the future and will address more then.

hera_mc/geo_location.py
david-deboer: I've done a major overhaul in another branch, which handles some of this (and moves modules in other files...)

bhazelton:
For the gps2time, I'd try adding the following method to your geo_location object (and similar code to the other ones):

    def __init__(self):
        super(GeoLocation, self).__init__()
        self.created_date = Time(self.created_gpstime, format='gps')

Then created_date should exist on your objects but not be columns in the database. I don't promise this will work, I haven't tried it on sqlalchemy objects, but I have used it on other objects with inherited init methods and I think it might work. There may also be a more elegant way to do it, I'll poke around on google a bit.

PR #41, opened 2017-07-12
Cm args test

david-deboer:
I'd like to merge this in because it changes the structure of the geo_location calls (makes a lot of things cleaner), and also it seems to have a conflict with master and I'd like to fix that up.  I wonder if these conflicts are related to the issue that Peter had to fix, since I don't think the listed conflicts should be conflicts...  I did add some testing!  I plan to go through the same process with parts/connections and include more testing.

As you can tell, I'm a fan of more frequently merging with master.


pkgw:
Closing this PR in favor of #42, which rebases against master and fixes the Travis failure.

PR #42, opened 2017-07-12
"cm_args_test", rebuilt on master

pkgw:
This rebuilds #41 against the master branch, to the best of my ability.


pkgw:
Confirmed that the two different branches end up in the same place:
```lambda 16:25 hera_mc $ git diff master..cm_args_test |sha256sum
9f0a6f547ed21812e2bd627612381b35fbddc2b0c8267a7ef64ddda9bee22838  -
lambda 16:25 hera_mc $ git diff master..pr-cm_args_test-rebase  |sha256sum
9f0a6f547ed21812e2bd627612381b35fbddc2b0c8267a7ef64ddda9bee22838  -
```

pkgw:
CC @bhazelton @dannyjacobs.

pkgw:
OK, I think this is OK to merge, but the diffs are so complex that it is very hard to evaluate the changes. I have added back one piece of code that seems to have been deleted erroneously. A lot of the changes are just in the indentation, so it is possible to get a clearer view by checking out the PR branch and doing something like:

```
git diff -w master..HEAD
```

which will hide whitespace-only changes.

@david-deboer you really need to start dividing your work into smaller, focused commits. Longer, more carefully-composed commit messages would also be very helpful. Otherwise we're just not able to understand what your changes are doing and review them properly. Review your diffs before committing with `git diff` and use `git add` and `git add -p` to group your code changes into smaller chunks for committing. This will also help because there are several instances in which clear typos or merge-conflict markers were committed, and reviewing the changes before committing would probably have caught those cases.

hera_mc/geo_handling.py
```diff
@@ -19,298 +21,397 @@
...
+        checks to see if the station_name is in the connections database (which means it is also in parts)
 
-    Parameters:
-    ------------
-    args:  needed arguments to open database
-    station_to_fine:  station name to find
-    query_date:  astropy Time for contemporary antenna
-    show_location:   if True, it will print the information.
-    """
+        return True/False unless check_if_active flag is set, when it returns the antenna number at that location
```
bhazelton: Is the check_if_active flag meant to be passed into this method? If not, and since it doesn't appear to be an attribute on the class, the output of the method should not depend on it.

Actually, this documentation appears to be wrong because I don't see this flag used anywhere in the function.

hera_mc/geo_handling.py
```diff
@@ -19,298 +21,397 @@
...
+                print("Error:  more than one active connection for", station_name)
+                station_connected = False
+        return station_connected
+
+
+    def find_station_of_antenna(self, antenna, query_date):
+        """
+        checks to see at which station an antenna is located
+
+        Returns False or the active station_name (must be an active station for the query_date)
```
bhazelton: I'm uncomfortable with the use of a boolean as opposed to a string if the station isn't available. I suggest using None instead.

hera_mc/geo_handling.py
```diff
@@ -19,298 +21,397 @@
...
+
+        Parameters:
+        ------------
+
+        antenna:  antenna number as float (why?), int, or string. If needed, it prepends the 'A'
+        query_date:  is the astropy Time for contemporary antenna
```
bhazelton: why indeed? if a float doesn't make sense, you should test for it and raise an error.

hera_mc/geo_handling.py
```diff
@@ -19,298 +21,397 @@
...
+        """
+
+        if type(antenna) == float or type(antenna) == int or antenna[0] != 'A':
+            antenna = 'A' + str(antenna).strip('0')
+        connected_antenna = self.session.query(part_connect.Connections).filter( (part_connect.Connections.downstream_part == antenna)  &
+                                                                                 (query_date.gps >= part_connect.Connections.start_gpstime) )
```
bhazelton: this isn't pep8 compliant. Make sure you check the rest of the code too.

hera_mc/geo_handling.py
```diff
@@ -19,298 +21,397 @@
...
+        found_location = []
+        for L in to_find:
+            station_name = False
+            try:
+                antenna_number = int(L)
+                station_name = self.find_station_of_antenna(antenna_number, query_date)
+            except ValueError:
+                station_name = L.upper()
+            found_it = False
+            if station_name:
```
bhazelton: Can this if statement ever fail? If so, it looks like the returned list will just be missing entries for those stations. I think this might be very confusing for a user, because they wouldn't know which ones were missing. I strongly suggest adding a None to the returned list in the location of the station that wasn't found.

hera_mc/geo_handling.py
```diff
@@ -19,298 +21,397 @@
...
+                            print('\tactive:  ', active)
+                            print('\tcreated:  ', cm_utils._get_displayTime(a.created_date))
+                        elif verbosity == 'l':
+                            print(a, this_station)
+            if show_location:
+                if not found_it and verbosity == 'm' or verbosity == 'h':
+                    print(L, ' not found.')
+        return found_location
+
+    def get_all_everconnected_locations(self):
```
bhazelton: This method needs a docstring.

hera_mc/geo_handling.py
```diff
@@ -19,298 +21,397 @@
...
+        station_name:  string name of station
+        active_date:  astropy Time to check if active, ignores if not an instance of Time
+        """
+
+        connected_station = self.session.query(part_connect.Connections).filter(part_connect.Connections.upstream_part == station_name)
+        if connected_station.count() > 0:
+            station_connected = True
+        else:
+            station_connected = False
+        if station_connected and isinstance(active_date,Time):
```
bhazelton: what happens if active_date is not a Time object? I notice below (line 272) that you call this method and pass '>' for the active_date, but it looks like that will just get ignored here. I really think you should test that active_date is the type you expect and raise an error if it's not. Silently ignoring it leads to hard to diagnose problems in the future.

hera_mc/geo_handling.py
```diff
@@ -19,298 +21,397 @@
...
+
+    def get_all_everconnected_locations(self):
+        stations = self.session.query(geo_location.GeoLocation).all()
+        connections = self.session.query(part_connect.Connections).all()
+        stations_conn= []
         for stn in stations:
             hera_proj = Proj(proj='utm', zone=stn.tile, ellps=stn.datum, south=True)
             stn.lon, stn.lat = hera_proj(stn.easting, stn.northing, inverse=True)
-            ever_connected = geo_location.is_in_connections(args, stn.station_name,'>')
+            ever_connected = self.is_in_connections(stn.station_name,'>')
```
bhazelton: Does passing '>' for active_date work? It looks like it gets ignored if it's not an astropy Time object.

hera_mc/geo_handling.py
```diff
@@ -19,298 +21,397 @@
...
+                                          'longitude': stn.lon,
+                                          'latitude': stn.lat,
+                                          'elevation': stn.elevation,
+                                          'antenna_number': ant_num,
+                                          'start_date': start_date,
+                                          'stop_date': stop_date})
+        return stations_conn
+
+
+    def get_since_date(self,query_date,station_types_to_check):
```
bhazelton: This method needs a docstring and maybe a better name. Get what since date?

hera_mc/geo_handling.py
```diff
@@ -19,298 +21,397 @@
...
+
+
+    def plot_stations(self, stations_to_plot, query_date, state_args):
+        """
+        Plot a list of stations.
+
+        Parameters:
+        ------------
+        stations_to_plot:  list containing station_names (note:  NOT antenna_numbers)
+        query_date:  date for active-only plot
```
bhazelton: this description suggests that query_date is optional if you don't want an active-only plot, but since it's positional it's actually required.

hera_mc/geo_handling.py
```diff
@@ -19,298 +21,397 @@
...
+        state_args:  dictionary with state arguments (fig_num, marker_color, marker_shape, marker_size, show_label)
+        """
```
bhazelton: the docstring should note that the figure number is returned.

hera_mc/geo_handling.py
```diff
@@ -19,298 +21,397 @@
...
 
-       Parameters:
-       ------------
-       args:  needed arguments to open database and set date/time
-       label_station:  flag to either label the station or not.  If not false, give label type.
-    """
+        Parameters:
+        ------------
+        label_station:  flag to either label the station or not.  If not false, give label type.
+        query_date:
```
bhazelton: docstring is incomplete.

hera_mc/geo_handling.py
```diff
@@ -19,298 +21,397 @@
...
+        return state_args['fig_num']
+
+
+    def overplot(self, located_stations, state_args):
+        """Overplot station on an existing plot.  It sets specific symbols/colors for active, connected, etc
+
+           Parameters:
+           ------------
+           located:  geo class of station to plot
+           state_args:
```
bhazelton: docstring is incomplete.

hera_mc/geo_handling.py
```diff
@@ -19,298 +21,397 @@
...
-    located = locate_station(args, args.locate, '>', show_location=False)
+    args = parser.parse_args([])
+    query_date = cm_utils._get_datetime(query_date)
+    h = Handling(args)
+    located = h.get_location(location_names, query_date, show_location=show_location, verbosity=verbosity)
+    h.close()
     return located
 
+def show_it_now(fignm):
+    if fignm is not False and fignm is not None:
```
david-deboer: added - it is used in scripts, so outside the file so leaving off underscore.

hera_mc/geo_handling.py
david-deboer: Documentation was wrong.

hera_mc/geo_handling.py
david-deboer: Fixed this by sending via cm_utils._get_datetime and fixing up around that.

hera_mc/geo_handling.py
david-deboer: Made None and sent query_date via _get_datetime also as above.

hera_mc/geo_handling.py
david-deboer: Float is fine, I just don't remember why I explicitly called out float and am unwilling to just ignore that...

hera_mc/geo_handling.py
david-deboer: Ran through pep8-ifier

hera_mc/geo_handling.py
bhazelton: That's the purpose of tests :). If it needs to be a float to support some specific use case, that case needs to be a test so that you can ensure that it always works (or verify that it can change to int if other code changes).

hera_mc/geo_handling.py
david-deboer: It can.  I've added None -- good idea.

hera_mc/geo_handling.py
david-deboer: Thanks for catching this - I fixed per above.

hera_mc/geo_handling.py
david-deboer: Did both.

hera_mc/geo_handling.py
david-deboer: Sort of true, but I've removed the comment in the docstring.

hera_mc/geo_handling.py
david-deboer: It does.

PR #48, opened 2017-07-14
Cm handling hookup fix

david-deboer:
Small changes to make cm parts work for now.

### DIFF
--- a/hera_mc/cm_handling.py
+++ b/hera_mc/cm_handling.py
@@ -39,6 +39,14 @@ class Handling:
 
     def __init__(self, args):
         self.args = args
+    #------START-TO-DO------#
+        #-#db = mc.connect_to_mc_db(self.args)
+        #-#self.session = db.sessionmaker()
+        #-#self.station_types = None
+
+    #-#def close(self):
+        #-#self.session.close()
+    #------END-TO-DO------#
 
     def is_in_connections(self, hpn_query, rev_query='ACTIVE', return_active=False):
         """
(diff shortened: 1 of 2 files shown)


hera_mc/cm_handling.py
```diff
@@ -39,6 +39,14 @@ class Handling:
...
     def __init__(self, args):
         self.args = args
+    #------START-TO-DO------#
+        #-#db = mc.connect_to_mc_db(self.args)
+        #-#self.session = db.sessionmaker()
+        #-#self.station_types = None
+
+    #-#def close(self):
+        #-#self.session.close()
+    #------END-TO-DO------#
```
pkgw: As I mentioned in my comment earlier, it would be good to be more selective when committing so as not to add these kinds of changes to `master`. But to get this turned around quickly, I am OK with this being here this time.

hera_mc/cm_hookup.py
```diff
@@ -37,11 +37,13 @@ def __get_connection(self, direction, part, rev, port, session):
         if direction == 'up':      # Going upstream
             for conn in session.query(part_connect.Connections).filter((part_connect.Connections.downstream_part == part) &
                                                                        (part_connect.Connections.down_part_rev == rev)):
+                conn.gps2Time()
```
pkgw: Looking at what the `gps2Time` routines do, I think there is a better solution. We could add methods to the classes like this:

```
@property
def start_date(self):
    return Time(self.start_gpstime, format='gps')
```

to create the `Time` objects on demand without needing to remember to add these conversion calls: it all happens transparently if you access `thingie.start_date`. But again, in the interests of rapid turnaround, I think it is OK to consider that possibility in a separate PR/issue.

pkgw:
Hmm, I guess GitHub doesn't let you rebase in PRs without them being merged against the latest version of `master`. That's not an inherent limitation in Git but it does eliminate the possibility of merge conflicts.

bhazelton:
Give me a moment to play with settings, @pkgw, this may be a setting I set too stringently

PR #49, opened 2017-07-15
Cm update db

david-deboer:
Only updated cm csv files (and added a short-cut flag -e for --exact-match).  This is the last update before THE BIG RENUMBERING exercise.


david-deboer:
Thanks.  I don't know why they became floats -- it was written by the to_csv method under pandas.  I put in a code fix subsequently.  I'll rebase and merge.

PR #50, opened 2017-07-17
Cm renumber

david-deboer:
This has the new cm tables, with all of the renumbering, stoppages etc.  In writing the (one-time) scripts to do this I tweaked/fixed other things that got exposed (I know....).

I suspect the "Can't automatically merge" is because I started this off of a branch other than master, since master didn't have the previous latest tables.  So the csv files might conflict?  Obviously I want these new ones.


pkgw:
@david-deboer OK — because of the way that we are rebasing branches into master right now, you **cannot** safely start new branches off of old branches! If we were merging branches into master it would be fine, but with the current approach it isn't. This is one of the reasons that I would prefer a merge-based methodology.

We can use Git's "cherry-pick" functionality to pull the changes from this branch into new branches that are based on the correct history. Cherry-pick slurps the actual changes and log message from a commit but sticks them on top of a different history — which is exactly what we want in this situation.

The latest commit on `master` is 4ef0220, "updated base files". You have 12 commits on this PR that are subsequent to that, but Git thinks you have other commits because of the rebasing.

The magic commands to rebuild this branch on the proper master branch are, I think:

```
git fetch origin # ensure we're up-to-date
git checkout -b new-cm-renumber origin/master # new branch based on latest master
git cherry-pick 1db8481..100105e # cherry-pick the commits we want
git branch -m cm-renumber old-cm-renumber # rename first branch to "old"
git branch -m new-cm-renumber cm-renumber # rename this branch to "new"
git push -f origin # force-push revised branch to GitHub
```

Here the magic numbers are the commit ID of what *this* branch erroneously thinks is the "update base files" commit (1db8481) and the last commit on this branch (100105e). The problem is that with the rebase onto master, "update base files" is now a new different commit.

In the future I would *really* like to see you keep the CSV updates and the code updates in separate commits.

hera_mc/cm_handling.py
```diff
@@ -153,7 +153,7 @@ def __find_ports(self, connections_dict, current):
                 elif ':down:' in k and v.upstream_output_port not in output_ports:
                     output_ports.append(v.upstream_output_port)
                 else:
-                    print("cm_handling[146]: ERROR SHOULD BE UP or DOWN ", k)
+                    print("cm_handling[156]: WARNING SHOULD BE UP or DOWN ", k)
```
pkgw: After you print this warning the code will just keep on running along merrily. Is that the best behavior? Should it raise an exception instead?

hera_mc/cm_transfer.py
```diff
@@ -183,5 +183,7 @@ def __initialization(args):
                         for i, r in enumerate(row):
                             if r == '':
                                 r = None
+                            elif 'gpstime' in field_name[i]:
```
pkgw: Please add a comment here explaining why this is necessary — namely, the fact that the stoptime columns get written out as floats because they're nullable, and Pandas has to represent those null values as NaNs, and there's no such thing as a NaN integer.

hera_mc/cm_utils.py
```diff
@@ -14,7 +14,8 @@
 import os.path
 
 PAST_DATE = '2000-01-01'
-FUTURE_DATE = '2025-12-31'
+def _future_date():
+    return Time.now() + TimeDelta(100, format='jd')
```
pkgw: What's the motivation for this change? Not obvious to me that there was anything wrong with the previous version.

hera_mc/geo_handling.py
```diff
@@ -144,38 +144,44 @@ def is_in_geo_location(self, station_name):
...
-        return True/False unless active_date is a Time, when it returns the antenna number at that location
+        return True/False unless (1) active_date is provided and (2) return_antrev is True
 
         Parameters:
         ------------
-        args:  needed arguments to open database and set date/time
         station_name:  string name of station
-        active_date:  astropy Time to check if active, ignores if not an instance of Time
+        active_date:  astropy Time to check if active, default is None
+        return_antrev:  boolean lag to return True/False or antrev tuple, default is True
```
pkgw: Typo: "flag", I think.

hera_mc/geo_handling.py
```diff
@@ -144,38 +144,44 @@ def is_in_geo_location(self, station_name):
...
         """
 
         active_date = cm_utils._get_datetime(active_date)
-        connected_station = self.session.query(part_connect.Connections).filter(part_connect.Connections.upstream_part == station_name)
-        if connected_station.count() > 0:
-            station_connected = True
+        connected_antenna = self.session.query(part_connect.Connections).filter(part_connect.Connections.upstream_part == station_name)
+        if connected_antenna.count() > 0:
+            antenna_connected = True
```
pkgw: It might be a bit clearer if `antenna_connected` is renamed to `retval` or `result` or something. First, the name is very similar to `connected_antenna`. Second, the name is misleading when the `return_antrev` option is active.

hera_mc/geo_handling.py
```diff
@@ -238,8 +244,8 @@ def get_location(self, to_find, query_date, show_location=False, verbosity='m'):
                             this_station = 'No station type data.'
                     a.gps2Time()
                     desc = self.station_types[this_station]['Description']
-                    ever_connected = self.is_in_connections(a.station_name, '<')
-                    active = self.is_in_connections(a.station_name, query_date)
+                    ever_connected = self.is_in_connections(a.station_name)
+                    active = self.is_in_connections(a.station_name, query_date, True)
```
pkgw: Please write out `return_antrev=True` in the function call to clarify that kind of data is being returned.

hera_mc/geo_handling.py
```diff
@@ -144,38 +144,44 @@ def is_in_geo_location(self, station_name):
             station_present = False
         return station_present
 
-    def is_in_connections(self, station_name, active_date=None):
+    def is_in_connections(self, station_name, active_date=None, return_antrev=True):
```
pkgw: If the function is going to stay named `is_in_connections`, the default value of `return_antrev` should be False — otherwise the function name makes it sound like it returns a boolean, but the default behavior is to not do this.

This function is getting a bit Swiss-Army-knifey — it might be simpler to break it into multiple functions, each with more focused semantics.

hera_mc/geo_handling.py
```diff
@@ -340,12 +346,12 @@ def plot_stations(self, stations_to_plot, query_date, state_args):
...
                         else:
-                            antrev = self.is_in_connections(station, query_date)
+                            antrev = self.is_in_connections(station, query_date, True)
                             if antrev is False:
                                 labeling = 'NA'
                             else:
-                                ant = antrev.split(':')[0]
-                                rev = antrev.split(':')[1]
+                                ant = antrev[0]
+                                rev = antrev[1]
```
pkgw: This can be simplified to `ant, rev = antrev`.

hera_mc/geo_handling.py
```diff
@@ -329,7 +335,7 @@ def plot_stations(self, stations_to_plot, query_date, state_args):
             for a in self.session.query(geo_location.GeoLocation).filter(geo_location.GeoLocation.station_name == station):
                 show_it = True
                 if state_args['show_state'].lower() == 'active':
-                    show_it = self.is_in_connections(station, query_date)
+                    show_it = self.is_in_connections(station, query_date, False)
```
pkgw: If the default setting for `return_antrev` is changed to be False, in this case the False argument should just be removed.

scripts/move_paper_feed.py
```diff
@@ -49,7 +49,7 @@ def OK_to_add(args, connect, handling):
         print('Error: ', connect.upstream_part, "already connected.")
         is_OK_to_add = False
     # 3 - check to see if antenna is already connected (should be, but isn't necesarily active)
-    is_connected = handling.is_in_connections(connect.downstream_part, 'A', return_active=True)
+    is_connected = handling.is_in_connections(connect.downstream_part, 'P', return_active=True)
```
pkgw: Are these magic letters 'P', 'T', etc., documented anywhere?

scripts/renumber_ants.py
```diff
@@ -0,0 +1,330 @@
...
+
+    if args.actually_do_it:
+        part_connect.update_part(args, data)
+    else:
+        print(data)
+
+def get_connection_key(c,p):
+    for ck in c.keys():
+        if p[0] in ck:
+            break
```
pkgw: Not important since this is a one-time script that has presumably already served its purpose, but you could just to `return ck` here, and `return None` if the `for` loop exits.

hera_mc/cm_handling.py
david-deboer: I changed it to a Warning from an Error, because it reflects something fishy in the database, but the ending result is (generally) fine.  I updated the text in the warning a bit.  (I've already spent too much time trying to debug that one, since it is a funny interaction of things...)

hera_mc/cm_transfer.py
david-deboer: Thanks for figuring this out!  Done.

hera_mc/cm_utils.py
david-deboer: I did this because astropy.Time gives a warning for future dates over a certain UT1/UTC transition (?) or some reason.  So, I wanted to end the warnings that get printed out.  If I made a FUTURE_DATE close enough to not, it might not always be the FUTURE.

hera_mc/geo_handling.py
david-deboer: I sort of agree.  I did modify it from the trickiness of return boolean if active_date was none otherwise return value (that was not a good thing...).  I have been cleaning up geo_handling and cm_handling (I've done a bit already on geo_handling), so I'd like to defer this to the future as I continue cleaning...

hera_mc/geo_handling.py
david-deboer: Same reply as above -- if I change how these work it permeates the code and I'd best do that intelligently under a separate branch.

hera_mc/geo_handling.py
david-deboer: I added the explicit return_antrev=False a al earlier.  I like keeping needed values in there, in case I foolishly change a default...

scripts/move_paper_feed.py
david-deboer: Yes - I have an unpushed branch where I'm updating Docs.  This is one of those to-be-deprecated scripts after this PR.

hera_mc/cm_utils.py
pkgw: Ah, this makes sense. An explanatory comment in the code would be helpful.

scripts/renumber_ants.py
david-deboer: I actually don't like having returns up in the function, but rather one at the end.  It gets hard to figure out things if you return in multiple places (I think).

pkgw: Strongly disagree! Returns within a function give you a very flexible way to control the flow of execution that can't be reproduced with the standard set of if's and loops. I find them helpful as a way to analyze a sequence of possible states that are slowly removed from consideration.

PR #51, opened 2017-07-18
scripts/mc_server_status_daemon.py: add server status reporter

pkgw:
Various HERA machines are supposed to periodically report their status to the M&C system. Most of the information can be gathered fairly straightforwardly, but we do want to average load values over time to beat down the noise somewhat. As best I (PKGW) can tell, the most straightforward way to do this is to write a small daemon script that maintains and M&C database connection and does the averaging. Lo and behold, here we are.

We also fix a mistake in mc_session.py.

### DIFF
--- a/hera_mc/host_status.py
+++ b/hera_mc/host_status.py
@@ -4,6 +4,8 @@
 
 """M&C logging of the uptime, load, etc., on our computers out in the Karoo.
 
+NOTE: this is superseded by the server_status set of tables!
+
 """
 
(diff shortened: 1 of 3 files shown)


pkgw:
I think the most practical thing would be to manually set it up to launch on boot, on all of the relevant machines. Because you need this long-running daemon I don't think it makes a ton of sense to try to orchestrate its launch through services like Librarian or RTP.

PR #52, opened 2017-07-19
only make/move Archived scripts dir

david-deboer:
This just makes the Archived directory under scripts to handle the one-time/timed-out scripts to keep track of part movement etc.


pkgw:
The CI is failing because the `setup.py` installation step ends up trying to install `Archived` like a script file. The code that triggers this [is here](https://github.com/HERA-Team/hera_mc/blob/master/hera_mc/version.py#L64) — the setup is a bit circuitous.

I'd suggest putting these scripts in a different toplevel directory named `oneoffs`. Maybe also add a date code to the script names (e.g. `oneoffs/201707_renumber_ants.py`) to give a bit more context.

PR #58, opened 2017-07-26
Fix geo_handling and cm_transfer to take sessions rather than args

bhazelton:
Fixes many tests that break otherwise if the default_db is not in testing mode.

Also fixed a few bugs to make write_antenna_location_file.py work.

### DIFF
--- a/Deprecated/stop_all_C7.py
+++ b/Deprecated/stop_all_C7.py
@@ -9,7 +9,7 @@ Script to handle the ONE TIME disconnection of all 150m cables
 
 from __future__ import absolute_import, division, print_function
 
-from hera_mc import mc, cm_utils, part_connect, cm_handling, geo_handling, cm_hookup, cm_part_revisions
+from hera_mc import mc, cm_utils, part_connect, cm_handling, cm_hookup, cm_part_revisions
 import sys
 
 
--- a/hera_mc/cm_transfer.py
+++ b/hera_mc/cm_transfer.py
@@ -15,37 +15,40 @@ import os.path
 import csv
 
 
-def package_db_to_csv(args):
+def package_db_to_csv(session=None, tables='all', base=False, maindb=False):
     """This will get the configuration management tables from the database
        and package them to csv files to be read by initialize_db_from_csv"""
+    if session is None:
+        db = mc.connect_to_mc_db()
+        session = db.sessionmaker()
+
     tmp_filename = '__tmp_initfile.tmp'
 
-    if args.base:
+    if base:
         data_prefix = cm_table_info.base_data_prefix
     else:
         data_prefix = cm_table_info.data_prefix
     cm_tables = cm_table_info.cm_tables
 
-    db = mc.connect_to_mc_db(args)
-    if args.tables == 'all':
+    if tables == 'all':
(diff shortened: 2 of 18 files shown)


Deprecated/stop_all_C7.py
```diff
@@ -9,7 +9,7 @@
 
 from __future__ import absolute_import, division, print_function
 
-from hera_mc import mc, cm_utils, part_connect, cm_handling, geo_handling, cm_hookup, cm_part_revisions
+from hera_mc import mc, cm_utils, part_connect, cm_handling, cm_hookup, cm_part_revisions
```
pkgw: Might make sense not to touch the scripts in `Deprecated/` once they've served their purpose? Obviously not a big deal though.

hera_mc/cm_transfer.py
```diff
@@ -15,37 +15,40 @@
 import csv
 
 
-def package_db_to_csv(args):
+def package_db_to_csv(session=None, tables='all', base=False, maindb=False):
     """This will get the configuration management tables from the database
        and package them to csv files to be read by initialize_db_from_csv"""
```
pkgw: Docstring descriptions of the new arguments would be useful here.

hera_mc/cm_transfer.py
```diff
@@ -54,14 +57,14 @@ def package_db_to_csv(args):
             os.remove(tmp_filename)
 
 
-def initialize_db_from_csv(args):
-    """This entry module provides a double-check entry point to read the csv files and 
+def initialize_db_from_csv(session=None, tables='all', base=False, maindb=False):
+    """This entry module provides a double-check entry point to read the csv files and
        repopulate the configuration management database.  It destroys all current entries,
        hence the double-check"""
```
pkgw: Docstrings again.

hera_mc/cm_transfer.py
```diff
@@ -136,8 +140,8 @@ def __initialization(args):
             print('Initialization for %s not found' % (table))
             use_table.remove(table)
         else:
-            if args.maindb:
-                if dbkey != args.maindb:
+            if maindb:
+                if dbkey != maindb:
```
pkgw: Oh, seeing what this `maindb` variable is ... I do not think it has a very good name. But that's an issue for another PR.

Deprecated/stop_all_C7.py
bhazelton: Yeah, I fixed this when it was not in deprecated, but when I rebased on master it had been moved. I can remove it if really desired, but it's importing something that's not used, so it doesn't really change anything.

hera_mc/cm_transfer.py
bhazelton: yes, I'll add them.

hera_mc/cm_transfer.py
bhazelton: yeah, it took me a while to figure it out and it's another one of those times where it can have different types, either boolean or a string which is weird. but I agree, it should be changed in a different PR.

david-deboer: Right - this is already an issue for the fix maindb thing.

PR #62, opened 2017-07-27
Add subsystem_error table

bhazelton:
Added a subsystem_error table to track errors generated by subsystems. Added appropriate methods, tests, documentation and alembic version file.


hera_mc/mc_session.py
```diff
@@ -252,6 +252,56 @@ def get_server_status(self, subsystem, starttime, stoptime=None, hostname=None):
...
+        """
+        Add a new subsystem subsystem_error to the M&C database.
+
+        Parameters:
+        ------------
+        subsystem: string
+            name of subsystem.
+        time: astropy time object
+            time of this error report
+        log: string
```
pkgw: Missing docstring for "severity", which is important since we should document that lower numbers are more severe.

hera_mc/mc_session.py
```diff
@@ -252,6 +252,56 @@ def get_server_status(self, subsystem, starttime, stoptime=None, hostname=None):
...
+        subsystem: string
+            subsystem to get records for. If none, all subsystems will be included.
+
+        Returns:
+        --------
+        list of SubsystemError objects
+        """
+        from .subsystem_error import SubsystemError
+
+        error_list = self._time_filter(SubsystemError, 'time', starttime,
```
pkgw: Style comment: could just return this value instead of saving it in a variable that is immediately returned.

PR #63, opened 2017-07-27
Set up Travis to use alembic

bhazelton:
Make the Travis default database in production mode and use Alembic to create the schema.

Added a test that fails if the alembic revisions don't make the database match the code schema.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,8 +20,9 @@ install:
   - conda info -a
 
   # create environment and install dependencies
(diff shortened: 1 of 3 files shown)


bhazelton:
I thought about adding a test that downgrades the database by one revision and checks that a similar test fails, but I didn't because I didn't want it to be unsafe to run tests onsite.

I did do that test on my own machine and the test failed, as expected.

PR #64, opened 2017-07-27
update lib_raid_errors alembic revision to properly add autoincrement

bhazelton:
This is a hotfix for the alembic revision that modified the lib_raid_errors table. I'd like to merge it before the alembic update happens on site.


pkgw:
BTW, I'm guessing that this technique is derived from the discussion in [this Google Group thread](https://groups.google.com/forum/#!topic/sqlalchemy-alembic/DaBpaca0Bug). It would be good to include a link to the discussion in the commit message, along with perhaps a more extended discussion of how the chosen approach was arrived at, to get the contextual information logged in the Git history.

bhazelton:
Good point about the commit message, I'll try to do that in the future. It looks like I can't change it now without a force push to the master branch which would be very bad. I could add a comment in the alembic revision file if you think I should.

PR #65, opened 2017-07-27
Remove inappropriate auto-increments in hera_obs, lib_status and rtp_status

bhazelton:


alembic/versions/5b141f32ea38_remove_bad_autoincrements.py
```diff
@@ -0,0 +1,42 @@
...
+
+# revision identifiers, used by Alembic.
+revision = '5b141f32ea38'
+down_revision = '4653b4902dc0'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
```
pkgw: Should ditch the `###` comments in these files if/when the commands have been vetted by a human.

pkgw:
If I'm understanding correctly this should now be good to merge, but I'll hold off just in case.

PR #68, opened 2017-07-30
Cm pam change

david-deboer:
This became a fairly comprehensive update to de-argify stuff, as well as put in the pam script, with a number of improvements on scripting stuff.  It is up to date with the current master.

I also started doing stuff for the new cm db workflow.  But then realized I should do that separately...


david-deboer:
Uhh - sorry, I see tests aren't passing (I was over-eager on the PR)...

hera_mc/part_connect.py
```diff
@@ -369,6 +509,8 @@ def format_check_update_connection_request(request):
     valueN:  corresponding new value
     """
```
bhazelton: Docstring has wrong parameters

hera_mc/tests/test_geo_location.py
```diff
@@ -17,18 +17,66 @@
...
+        gl.station_name = prefix+'_null'
+        gl.station_type_name = stn
+        gl.datum = 'WGS84'
+        gl.tile = '34J'
+        gl.northing = 6601181.0
+        gl.easting = 541007.0
+        gl.elevation = 1051.69
+        gl.created_gpstime = 1172530000
+        self.test_session.add(gl)
+        self.test_session.commit()
```
bhazelton: This cofa adding code looks like it's copied and pasted a bunch of places. You might consider making a function in `tests/__init__.py` that contains this code and could be called quickly from the appropriate tests.

hera_mc/tests/test_geo_location.py
```diff
@@ -17,18 +17,66 @@
 
 class TestGeo(TestHERAMC):
 
-    def test_add_part(self):
-        self.test_session.add(geo_location.GeoLocation())
+
+    def test_add_loc(self):
```
bhazelton: I'm not sure this test is adding anything, given that your 3rd test does exactly the same thing and then retrieves the value and checks that it is as expected.

scripts/pam_swap.py
```diff
@@ -0,0 +1,126 @@
+#! /usr/bin/env python
+# -*- mode: python; coding: utf-8 -*-
+# Copyright 2016 the HERA Collaboration
+# Licensed under the 2-clause BSD license.
+
+"""
+Script to handle swapping of PAMs.
+"""
```
bhazelton: I worry about how much logic is in this script and therefore very difficult to test. Could you make it a very thin (almost no if/else logic) wrapper around a function that could be tested?

hera_mc/mc.py
```diff
@@ -110,6 +110,20 @@ def get_mc_argument_parser():
                    help='Name of the database to connect to. The default is used if unspecified.')
     return p
 
+def get_cm_csv_path(args):
```
david-deboer: I'll finish this shortly.  It does not currently exist.  I plan to do this tomorrow.

scripts/pam_swap.py
david-deboer: These scripts are going to need a bit of separate/specific logic.  I have pulled out some stuff that is more generic.

bhazelton: It's not just about pulling out generic code, even specific code can be in a module that can be tested. The problem is if you have a lot of logic in scripts, there's no good way to actually test the logical branches.

hera_mc/geo_handling.py
```diff
@@ -51,10 +49,8 @@ def get_location(location_names, query_date='now', show_location=False, verbosit
     location_name:  location name, may be either a station (geo_location key) or an antenna
```
bhazelton: location_name -> location_names to be consistent with function definition.

scripts/pam_swap.py
bhazelton: per Slack conversation, this script is fairly temporary to implement a specific change, so doesn't need the same level of testing.

hera_mc/geo_handling.py
```diff
@@ -31,15 +31,13 @@ def cofa(show_cofa=False):
     show_cofa:  boolean to print out cofa info or just return class
```
bhazelton: resolved.

hera_mc/geo_handling.py
```diff
@@ -95,9 +91,16 @@ def close(self):
     def cofa(self, show_cofa=False):
```
bhazelton: resolved.

hera_mc/cm_handling.py
```diff
@@ -35,293 +36,265 @@ class Handling:
...
-                self.show_other_connections(connection_dict, already_shown)
-        if return_dictionary:
-            return connection_dict
-
-    def show_active_connections(self, connection_dict):
+        #-#nc = self.no_connection_designator
+        #-#connection_dossier[nc] = PC.Connections(upstream_part=nc, upstream_output_port=nc, up_part_rev=nc,
+        #-#                                               downstream_part=nc, downstream_input_port=nc, down_part_rev=nc,
+        #-#                                               start_gpstime=cm_utils._get_datetime('<', '<').gps,
+        #-#                                               stop_gpstime=cm_utils._get_datetime('>', '>').gps)
```
david-deboer: Removed.

bhazelton: I think they're still there

hera_mc/tests/test_geo_location.py
david-deboer: I'm not sure.  It is nice to control from within your module.  I'd like to defer this one to think about it.

david-deboer:
The csv stuff is handled in the next PR.

PR #70, opened 2017-07-31
Cm new csv

david-deboer:
This provides the small changes to implement the new cm csv work flow.
You'll need the new repo hera_cm_db_updates and to add the path where it is installed to the config json file.
It is built off of cm-pam-change.


david-deboer:
Only the last commit should be checked, since the others are under cm-pam-change

pkgw:
You need to add a suitable CSV file path to the example config file in `ci` in order for the tests to pass.

david-deboer:
The new PR has more tests.

pkgw:
I'm going to temporarily close this as part of an experiment.

pkgw:
Oh, now I realize what's going on: taking out the `_initialization` call in the test harness means that those code paths aren't covered anymore, dropping the code coverage too much for Coveralls. So I think you need to keep the call and add a fake CSV path to the testing config file.

bhazelton:
Or there's another solution: drop that but add tests that cover the code in other ways

david-deboer:
The next PR adds tests, so that's good at least.  I think we should remove it from __init__, but I can try and add in another place.  Can we override this time?

pkgw:
Wait, can this PR just be closed? It's looking to me that #71 contains all of its changes.

david-deboer:
Well, I can make sure that 71 has them...

pkgw:
The only commit in this branch not present in cm-pc-tests is `888fdd7`, `fixed _initialization error`, which I believe is redundant with `2345523`, `Added test_parts. [...]`.

PR #71, opened 2017-07-31
Cm pc tests

david-deboer:
This adds some tests for parts and connections.  It is built off of cm-pam-change -> cm-new-csv


david-deboer:
I don't understand how the coverage didn't increase relative to the previous one, since I explicitly added tests (maybe it just made it more comprehensive, but of same stuff?)

david-deboer:
I guess I'll try to recover the _initialization test stuff...

pkgw:
It looks like the same thing is happening here as in #70: now `cm_transfer.py` is completely uncovered, which outweighs the gains of the new test. I can't see why it didn't think this was a problem before, because it still does seem to me that this PR had an equivalent change as in #70 with regards to getting rid of the `_initialization()` call.

bhazelton:
If you want to see what's going on with coverage, you really need to spend some time on the coveralls page. If you follow the 'details' link of the coveralls check below, you can see that you did add some new covered lines, but you also added a bunch of lines that aren't covered

bhazelton:
I think it's much better to add new tests to cover the code in a reasonable way, which we need anyway.

pkgw:
@bhazelton I think most of that churn is an effect of GitHub thinking that this PR contains a lot more new commits than it really does. The actual differences between this branch and master are pretty small. I don't know why GitHub isn't realizing that; I think it normally does.

david-deboer:
I agree - but I really need to get into data entry and data integrity.

So, this is the downside of basing new stuff off of unmerged old stuff.

bhazelton:
Coveralls shouldn't care about that -- It just asks what lines are covered by testing or not. You can run similar reports locally on the master branch and on this branch and compare the coverage. I'll do that now and see if I get similar results to coveralls.

bhazelton:
nosetest's coverage report is a little less accurate than coveralls, but it's also seeing a decrease in test coverage of about 3%

bhazelton:
That decrease is when I run the coverage report separately on master and on the cm-pc-tests branch, so it has nothing to do with the PR logic

pkgw:
Here's a Coveralls report for the changes in this branch without all of the erroneous churn that GitHub thinks is there: https://coveralls.io/builds/12630918

bhazelton:
So coveralls doesn't really care how you get the coverage up, you can cover the lines that are now not covered or you can write other tests that cover other areas of the code. There are a lot of uncovered lines of code right now, so there's a lot of low-hanging fruit.

david-deboer:
OK - we have success on the cm-pc-tests PR!

PR #73, opened 2017-07-31
Add QM tables

adampbeardsley:
This branch creates the tables and interface for quality metrics. There are three tables: `ant_metrics`: Stores metric info on a per antenna/pol basis
`array_metrics`: Stores array-wide metrics
`metric_list`: Stores descriptions for each of the metrics
The add methods are pretty straightforward. The get methods allow the user to provide any combination of keys, including lists of keys. For example, one could query all antennas and metrics 'MeanVij' and 'redCorr' in one call.
I have not run the alembic revision script yet -- figured it would be better to get some eyes on the code and table structure first.


hera_mc/mc_session.py
```diff
@@ -723,3 +724,178 @@ def get_paper_temps(self, starttime, stoptime=None):
...
+    def add_ant_metric(self, obsid, ant, pol, metric, val):
+        """
+        Add a new antenna metric to the M&C database.
+
+        Parameters:
+        ------------
+        obsid: long integer
+            observation identification number
+        ant: integer
+            antenna number
```
bhazelton: I think we need to think a little about what we actually want here, it turns out that 'antenna number' isn't a very precise term in M&C :). I do think we want something that is tied to the physical antenna, not the correlator input number, but I'm not quite sure whether it should be a 'station' or something else. @david-deboer?

I also wonder whether we want it to be a ForeignKey into one of the cm tables.

hera_mc/qm.py
```diff
@@ -0,0 +1,175 @@
...
+
+
+class ant_metrics(MCDeclarativeBase):
+    """
+    Definition of ant_metrics table.
+
+    obsid:      observation identification number, generally equal to the floor
+                of the start time in gps seconds (long integer)
+    ant:        Antenna number (int >= 0)
+    pol:        Polarization ('x' or 'y')
```
bhazelton: Do we want 'x' and 'y' or 'N' and 'E' here?

hera_mc/qm.py
```diff
@@ -0,0 +1,175 @@
...
+    mc_time = Column(BigInteger, nullable=False)
+    val = Column(Float, nullable=False)
+
+    # tolerances set to 1ms
+    tols = {'mc_time': DEFAULT_GPS_TOL}
+
+    @classmethod
+    def create(cls, obsid, metric, db_time, val):
+        """
+        Create a new ant_metric object using Astropy to compute the LST.
```
bhazelton: ant_metric -> array_metric

hera_mc/tests/test_qm.py
```diff
@@ -0,0 +1,132 @@
...
+
+class TestQM(TestHERAMC):
+
+    def setUp(self):
+        super(TestQM, self).setUp()
+        t1 = Time('2016-01-10 01:15:23', scale='utc')
+        t2 = t1 + TimeDelta(120.0, format='sec')
+        self.obsid = utils.calculate_obsid(t1)
+        # Create obs to satifsy foreign key constraints
+        self.test_session.add_obs(t1, t2, self.obsid)
```
bhazelton: We recently learned that if you have things that fail in the setup that it makes nosetests hang rather than just returning an error. So I suggest copying the parts where you actually add things to the session into the tests. It's annoying because it duplicates code, but nosetests hanging is really awful.

hera_mc/mc_session.py
adampbeardsley: I thought about this too. I ended up thinking that because the user will be entering this info as a result of some data analysis, it makes sense to use the antenna numbers in the data file. Of course we can always back out other info given the config management at the time.

hera_mc/tests/test_qm.py
adampbeardsley: Ah, I did experience the hanging a lot, couldn't figure out why it was doing that instead of just erroring. I will change this.

hera_mc/qm.py
adampbeardsley: Good catch.

hera_mc/qm.py
adampbeardsley: Thought about this too. Right now 'x' and 'y' are how things are treated in hera_qm, and I haven't seen much progress toward 'N' and 'E', other than the 'x_orientation' parameter in pyuvdata. But I can be persuaded.

hera_mc/mc_session.py
bhazelton: I think this is resolved -- antenna number will be tied to the physical antenna and will be written into the data files by the correlator in the future.

hera_mc/qm.py
bhazelton: let's leave it for now, we can change later if we need to.

hera_mc/tests/__init__.py
```diff
@@ -37,3 +40,53 @@ def setUp(self):
...
+        if reg:
+            reg.clear()
+
+
+def checkWarnings(func, func_args=[], func_kwargs={},
+                  category=UserWarning,
+                  nwarnings=1, message=None, known_warning=None):
+    """Function to check expected warnings."""
+    category = get_iterable(category)
+    message = get_iterable(message)
```
bhazelton: Lets use the updates that I just did on pyuvdata here, so that category and message get expanded if needed.

bhazelton:
If this is ready to go, take off the "DO NOT MERGE" prefix

adampbeardsley:
Thanks Bryna! Still not quite ready though - I need some things in hera_qm to settle before we make the interface official. I'll let you know when it's ready for a final sign-off (hopefully minimal changes before then).

adampbeardsley:
Ok @bhazelton I think this PR is ready for an official review. I made a few organizational changes since you last looked - moved code from scripts into modules so it can be tested, and some changes were made on the hand-off between hera_qm and hera_mc.
I had to add a few extra installs to the .travis.yml file because of hera_qm's dependency on pyuvdata. Hopefully this is ok.

hera_mc/mc_session.py
```diff
@@ -728,3 +730,240 @@ def get_paper_temps(self, starttime, stoptime=None):
...
+        pol: string ('x' or 'y')
```
pkgw: Aren't we supposed to be doing N and E for pols?

hera_mc/mc_session.py
pkgw: Would it be better to log both physical antenna number and correlator input number in this table?

hera_mc/mc_session.py
```diff
@@ -728,3 +730,240 @@ def get_paper_temps(self, starttime, stoptime=None):
...
+        if ant is not None:
+            args.append(ant_metrics.ant.in_(get_iterable(ant)))
+        if pol is not None:
+            args.append(ant_metrics.pol.in_(get_iterable(pol)))
+        if metric is not None:
+            args.append(ant_metrics.metric.in_(get_iterable(metric)))
+        if starttime is None:
+            starttime = 0
+        else:
+            if isinstance(starttime, Time):
```
pkgw: These could be condensed into `elif:` lines.

hera_mc/mc_session.py
```diff
@@ -728,3 +730,240 @@ def get_paper_temps(self, starttime, stoptime=None):
...
+            description of metric
+        """
+        from .qm import metric_list
+
+        self.add(metric_list.create(metric, desc))
+
+    def update_metric_desc(self, metric, desc):
+        """
+        Update the description of a metric in the M&C database.
+        This will be required when replace an RTP auto-generated description for
```
pkgw: Grammar-o.

hera_mc/mc_session.py
```diff
@@ -728,3 +730,240 @@ def get_paper_temps(self, starttime, stoptime=None):
...
+        for metric, desc in metric_list.items():
+            # Check if metric is already in db.
+            r = self.get_metric_desc(metric=metric)
+            if len(r) == 0:
+                self.add_metric_desc(metric, desc)
+            else:
+                self.update_metric_desc(metric, desc)
+        self.commit()
+
+    def add_metrics_file(self, filename, ftype):
```
pkgw: I'd suggest naming this `ingest_metrics_file` rather than `add_metrics_file`.

hera_mc/qm.py
```diff
@@ -0,0 +1,177 @@
...
+from math import floor
+from sqlalchemy import (Column, Integer, BigInteger, Float, ForeignKey,
+                        String, ForeignKeyConstraint)
+from sqlalchemy.ext.hybrid import hybrid_property
+
+from hera_mc import geo_handling
+from . import MCDeclarativeBase, DEFAULT_GPS_TOL, DEFAULT_DAY_TOL, DEFAULT_HOUR_TOL
+
+
+class ant_metrics(MCDeclarativeBase):
```
pkgw: I'd much rather have these class name written in the standard `CamelCase` style. You can then put:

```
__tablename__ = 'ant_metrics'
```

To have the database table name be written in `snake_case`.

hera_mc/utils.py
```diff
@@ -27,3 +27,14 @@ def calculate_obsid(starttime):
...
 
     return int(floor(starttime.gps))
+
+
+def get_iterable(x):
+    """Helper function to ensure iterability."""
+    if isinstance(x, str):
+        return (x,)
+    else:
+        if isinstance(x, collections.Iterable):
```
pkgw: This will work for the core built-in collection types but probably not possible user-defined iterable types. If we care about that, the more to-the-point approach would be to do something like:

```
try:
    iter(x)
except TypeError:
    return (x,)
return x
```

scripts/add_qm_metrics.py
```diff
@@ -0,0 +1,32 @@
...
+                    help='json files to read and enter into db.')
+parser.add_argument('--type', dest='type', type=str, default=None,
+                    help='File type to add to db. Options = ["ant", "firstcal", "omnical"]')
+args = parser.parse_args()
+db = mc.connect_to_mc_db(args)
+session = db.sessionmaker()
+
+files = args.files
+if len(files) == 0:
+    raise AssertionError('Please provide a list of quality metric files.')
```
pkgw: Since this is a standalone script, we know that an exception here is going to end up printing distracting backtrace information. Not a huge deal but you'd get cleaner output with:

```
import sys
print >>sys.stderr, 'Please provide...' # since this module doesn't use the print_function __future__
sys.exit(1)
```

scripts/update_qm_list.py
```diff
@@ -0,0 +1,19 @@
...
+# Licensed under the 2-clause BSD license.
+
+import hera_mc.mc as mc
+
+parser = mc.get_mc_argument_parser()
+args = parser.parse_args()
+db = mc.connect_to_mc_db(args)
+session = db.sessionmaker()
+
+metric_list = copy(ant_metric_list)
```
pkgw: I don't see how this script can work since nothing defines the variables `copy`, `ant_metric_list`, `firstcal_metric_list`, etc.

adampbeardsley: Woops, those lines were actually from an old version. The metric lists are now imported in the  `update_qm_list` method of mc_session. I will fix this.

hera_mc/mc_session.py
adampbeardsley: I can add that if we want it. Is there existing code to get it from obsid + antenna number?

hera_mc/mc_session.py
adampbeardsley: Yeah, does that notation exist anywhere else? Right now the data files map to x and y, and the metrics get stored according to that. I can easily map them back to N and E, but I didn't want to do that if everywhere is still x and y.

adampbeardsley:
Thanks for the review @pkgw. I've addressed almost all your suggestions in the latest commit - the exceptions being the correlator number and the grand N/S vs y/x debate (see specific responses).

hera_mc/mc_session.py
bhazelton: So there's going to be a correlator input string (not number) in the files the correlator writes in the future and I do think we want to persist it (this will require some development work on pyuvdata, but it should be pretty easy). There will also soon be code to get the correlator input string for an obsid + antenna, but it's still being written. So soonish we'll want to add that, but I'll leave it up to you @adampbeardsley whether to do it in this PR or add it later.

PR #76, opened 2017-08-01
scripts/mc_megaraid_monitor.py: script to monitor RAID statuses

pkgw:
Part of M&C's job is to check on the status of our RAID arrays. This script uses `storcli64` to assess the RAID status and log it to the database. There is all sorts of information that we *could* log, but hopefully this analysis covers the most important things by watching the event log for errors.

### DIFF
--- a/scripts/mc_megaraid_monitor.py
+++ b/scripts/mc_megaraid_monitor.py
@@ -92,7 +92,7 @@ args = parser.parse_args()
 db = mc.connect_to_mc_db(args)
 
 
-# Get the seqNum of the last even that we've noticed
+# Get the seqNum of the last event that we've noticed
 
 try:
     ticker_file = open(event_ticker, 'r')
@@ -130,7 +130,7 @@ for line in show_all.stdout:
 
 if show_all.wait() != 0:
     print('error: storcli exited with an error code', file=sys.stderr)
-    print('unfortunately this script may have swallwed its error message', file=sys.stderr)
+    print('unfortunately this script may have swallowed its error message', file=sys.stderr)
     sys.exit(1)
 
 num_disks = int(item_values.pop('Physical Drives', 0))
--- a/scripts/mc_megaraid_monitor.py
+++ b/scripts/mc_megaraid_monitor.py
@@ -14,6 +14,8 @@ client requires.
 """
 from __future__ import absolute_import, division, print_function
 
+import datetime
+import dateutil.tz
 import errno
 import json
 import os.path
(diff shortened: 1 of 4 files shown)


scripts/mc_megaraid_monitor.py
```diff
@@ -0,0 +1,244 @@
...
+
+
+# Before running anything, make sure we can connect to the DB.
+
+parser = mc.get_mc_argument_parser()
+args = parser.parse_args()
+db = mc.connect_to_mc_db(args)
+
+
+# Get the seqNum of the last even that we've noticed
```
bhazelton: even -> event

scripts/mc_megaraid_monitor.py
```diff
@@ -0,0 +1,244 @@
...
+    #----------------------------------------------------------------
+    #0/0   RAID60 Optl  RW     Yes     RWBD  -   ON  196.475 TB
+    #----------------------------------------------------------------
+
+    if line.startswith('0/0'):
+        item_values['VD 0/0 State'] = line.split()[2]
+
+if show_all.wait() != 0:
+    print('error: storcli exited with an error code', file=sys.stderr)
+    print('unfortunately this script may have swallwed its error message', file=sys.stderr)
```
bhazelton: swallwed -> swallowed

scripts/mc_megaraid_monitor.py
```diff
@@ -0,0 +1,244 @@
...
+    TBD: I think these times are in system local time (and not, say, UTC), but
+    it's not clear!
+
+    """
+    dow, month, dom, hhmmss, year = text.split()
+    month = _months[month]
+    dom = int(dom)
+    year = int(year)
+    iso_format = '%04d-%02d-%02d %s' % (year, month, dom, hhmmss)
+    return Time(iso_format, format='iso')
```
bhazelton: What time zone gets attached when you call this? The system local time? Or do you need to specify it somehow?

pkgw: Good call, needed to be a bit more careful here. Addressed in a80bbcc.

scripts/mc_megaraid_monitor.py
pkgw: Addressed in 67f2584.

scripts/mc_megaraid_monitor.py
pkgw: Addressed in 67f2584.

PR #77, opened 2017-08-01
Details for Mac OS X installation

adampbeardsley:
They may be really simple, but these are the notes I wish I had when I first installed.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -31,7 +31,7 @@ this, or follow the OS-X-specific notes below.
 
 
 Configure hera_mc to talk to the db:
-After setting up the database, you need to fill in the configuration file
+After setting up the database (see below), you need to fill in the configuration file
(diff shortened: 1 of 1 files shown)


INSTALL.md
```diff
@@ -62,9 +62,29 @@ If desired, populate the configuration management tables by running the `cm_init
 
 ### Basic OS X PostgreSQL installation
 
-Follow the directions under `Interactive installer by EnterpriseDB` here: https://www.postgresql.org/download/macosx/
+Install the app found here: http://postgresapp.com/, and follow steps 1 and 2 on that page.
```
pkgw: I will leave it to the OSX users to decide if this app is the best option for getting things installed on OSX. But, we should be forward-looking with where the Web is going: the link here should be an `https://` URL, not unencrypted `http://`.

PR #78, opened 2017-08-01
Cm add part

david-deboer:
This mostly adds a generic add_part and add_connection script, but it does clean up a few uglies in the connection stuff.  I added some tests as well.

### DIFF
--- a/hera_mc/cm_handling.py
+++ b/hera_mc/cm_handling.py
@@ -17,19 +17,6 @@ from hera_mc import mc, geo_handling, correlator_levels, cm_utils
 from hera_mc import part_connect as PC
 from hera_mc import cm_revisions as cmpr
 
-
-def _make_part_key(hpn, rev):
-    return ":".join([hpn, rev])
-def _split_part_key(key):
-    return key.split(':')[0], key.split(':')[1]
-
-def _make_connection_key(hpn, rev, port, start_gps):
-    return ":".join([hpn, rev, port, start_gps])
-def _split_connection_key(key):
-    ks = key.split(':')
-    return ks[0], ks[1], ks[2], ks[3]
-
-
 class Handling:
     """
     Class to allow various manipulations of parts and their properties etc.  Things are
@@ -121,7 +108,7 @@ class Handling:
                 elif part_cnt == 1:
                     part = copy.copy(part_query.all()[0])
                     part.gps2Time()
-                    pr_key = _make_part_key(part.hpn, part.hpn_rev)
+                    pr_key = cm_utils._make_part_key(part.hpn, part.hpn_rev)
                     part_dossier[pr_key] = {'Time': at_date, 'part': part, 'part_info': None,
                                             'input_ports': [], 'output_ports': [],
                                             'connections': None, 'geo': None}
@@ -247,7 +234,7 @@ class Handling:
                                                                       (PC.Connections.up_part_rev == this_rev)):
                     if port.lower() == 'all' or conn.upstream_output_port.lower() == port.lower():
                         conn.gps2Time()
-                        ckey = _make_connection_key(conn.downstream_part,       conn.down_part_rev, 
+                        ckey = cm_utils._make_connection_key(conn.downstream_part,       conn.down_part_rev, 
                                                     conn.downstream_input_port, conn.start_gpstime)
                         connection_dossier['connections'][ckey] = copy.copy(conn)
                         down_parts.append(ckey)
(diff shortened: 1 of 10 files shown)


hera_mc/cm_handling.py
```diff
@@ -160,13 +140,11 @@ def find_ports(self, connection_dossier):
...
-                continue
-            if ':up:' in k and v.downstream_input_port not in input_ports:
-                input_ports.append(v.downstream_input_port)
-            elif ':down:' in k and v.upstream_output_port not in output_ports:
-                output_ports.append(v.upstream_output_port)
+        for k, c in connection_dossier['connections'].iteritems():
+            if c.downstream_input_port not in input_ports:
+                input_ports.append(c.downstream_input_port)
+            elif c.upstream_output_port not in output_ports:
+                output_ports.append(c.upstream_output_port)
```
pkgw: If I'm interpreting this correctly, you're trying to make sorted lists of unique items here. An easier and more efficient way to do this is to use the `set()` type, which acts like a mathematical set and has an `.add()` method that adds item. Then you can write `input_ports = sorted(input_ports)` to convert the set into a sorted list.

hera_mc/cm_handling.py
```diff
@@ -388,10 +365,8 @@ def show_other_connections(self, connection_dossier, already_shown):
...
         """
 
-        for k, v in connection_dossier.iteritems():
-            if k in self.non_class_connection_dossier_entries:
-                continue
-            elif k in already_shown:
+        for k, v in connection_dossier['connections'].iteritems():
+            if k in already_shown:
                 continue
             else:
```
pkgw: The `else` isn't really needed since the non-`else` branch does a `continue`. Simplest is

```
for k, v in connection_dossier['connections'].iteritems():
   if k not in already_shown:
       print(v, v.start_date, v.stop_date)
```

scripts/add_connection.py
```diff
@@ -0,0 +1,75 @@
+#! /usr/bin/env python
+# -*- mode: python; coding: utf-8 -*-
+# Copyright 2016 the HERA Collaboration
+# Licensed under the 2-clause BSD license.
+
+"""
+Script to handle swapping of PAMs.
```
pkgw: This docstring needs to be updated.

scripts/add_connection.py
```diff
@@ -0,0 +1,75 @@
+#! /usr/bin/env python
+# -*- mode: python; coding: utf-8 -*-
+# Copyright 2016 the HERA Collaboration
```
pkgw: The year is 2017.

scripts/add_part.py
```diff
@@ -0,0 +1,67 @@
+#! /usr/bin/env python
+# -*- mode: python; coding: utf-8 -*-
+# Copyright 2016 the HERA Collaboration
```
pkgw: The year is 2017.

scripts/add_part.py
```diff
@@ -0,0 +1,67 @@
...
+# Licensed under the 2-clause BSD license.
+
+"""
+Script to handle swapping of PAMs.
```
pkgw: This docstring needs updating.

hera_mc/cm_handling.py
david-deboer: I guess I'm not sure how to implement that in a way that is easier and more efficient...

scripts/add_part.py
david-deboer: Done.

scripts/add_part.py
david-deboer: Done

scripts/add_connection.py
david-deboer: Done

scripts/add_connection.py
david-deboer: Done

hera_mc/cm_handling.py
david-deboer: Done

hera_mc/cm_handling.py
pkgw: Well, actually, what I had in mind would change the semantics. Is the `elif` in the current code strictly necessary? If not, I think you could write:

```
input_ports = set()
output_ports = set()

for k, c in connection_dossier['connections'].iteritems():
    input_ports.add(c.downstream_input_port)
    output_ports.add(c.upstream_output_port)

input_ports = sorted(input_ports)
output_ports = sorted(output_ports)
```

HERA-project: I see - the elif is needed because there can be duplicates and I only want the unique ones.

pkgw: Err, I think the elif is doing more work than that. Because it's the `else` branch, the loop will only register the output ports of parts that it *hasn't* registered the input ports for. The logic is equivalent to:

```
if downstream_input not in input_ports:
   input_ports.append(downstream_input)
else:
   if upstream_output not in output_ports:
         output_ports.append(upstream_output)
```

Is that really intended, or should it be the following?

```
if downstream_input not in input_ports:
   input_ports.append(downstream_input)
if upstream_output not in output_ports:
   output_ports.append(upstream_output)
```

david-deboer:
Thanks - good catch on the elif thing.

pkgw:
OK. In that case, I think the approach using `set` objects is a bit nicer, but feel free to merge now if you don't feel like changing that.

PR #79, opened 2017-08-03
Cm fem

david-deboer:
This adds the script to swap out the frontends and has a few bug fixes.

### DIFF
--- a/hera_mc/cm_hookup.py
+++ b/hera_mc/cm_hookup.py
@@ -153,7 +153,7 @@ class Hookup:
 
         # Get all the appropriate parts
         parts = self.handling.get_part_dossier(hpn=hpn, rev=rev, at_date=self.at_date, exact_match=exact_match)
-        hookup_dict = {}
+        hookup_dict = {'hookup':{}}
         col_len_max = [0, '-']
 
         for hpnr in parts.keys():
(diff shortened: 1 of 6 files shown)


hera_mc/cm_hookup.py
```diff
@@ -153,7 +153,7 @@ def get_hookup(self, hpn, rev, port, at_date, state_args, exact_match=False):
 
         # Get all the appropriate parts
         parts = self.handling.get_part_dossier(hpn=hpn, rev=rev, at_date=self.at_date, exact_match=exact_match)
-        hookup_dict = {}
+        hookup_dict = {'hookup':{}}
```
bhazelton: Does this mean that `hookup_dict` is a nested dict that can contain several hookups? If so, I think the docstring needs to be updated to reflect this -- it looks like the keys of hookup_dict are not keyed on hpn:port anymore.

It looks like there are now two keys in hookup_dict: `hookup` and `columns`. Is that right? The docstring should explain what they are.

david-deboer:
Do you have to add a test file?  nosetests doesn't seem to be finding test_parts.py as far as I can tell

hera_mc/cm_hookup.py
```diff
@@ -136,7 +136,9 @@ def __follow_hookup_stream(self, part, rev, port, sn):
     def get_hookup(self, hpn, rev, port, at_date, state_args, exact_match=False):
         """
         Return the full hookup.
-        Returns hookup_dict, a dictionary keyed on derived key of hpn:port.
+        Returns hookup_dict, a dictionary with two entries:
+            'hookup': another dictionary keyed on part:rev:port:sn
+            'columns': names of column headers
```
bhazelton: What does 'column headers' mean? It seems to refer to something in a table, but I don't know what that table is.

PR #85, opened 2017-08-08
I had forgotten to actually add the script!

david-deboer:
Somewhat embarrassingly I had forgotten to add the actual script to the commit...
Everything is the same, except stop_connection.py script has been added.

### DIFF
--- a/scripts/stop_connection.py
+++ b/scripts/stop_connection.py
@@ -1,0 +1,79 @@
+#! /usr/bin/env python
+# -*- mode: python; coding: utf-8 -*-
+# Copyright 2017 the HERA Collaboration
+# Licensed under the 2-clause BSD license.
+
+"""
+Script to add a general connection to the database.
+"""
+
+from __future__ import absolute_import, division, print_function
+
+from hera_mc import mc, cm_utils, part_connect, cm_handling
+import sys
+import copy
+
+
+def query_args(args):
+    """
+    Gets information from user
+    """
+    if args.uppart is None:
+        args.uppart = raw_input("Upstream part number:  ")
+    if args.uprev is None:
+        args.uprev = raw_input("Upstream part revision:  ")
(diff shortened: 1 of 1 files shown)


scripts/stop_connection.py
```diff
@@ -0,0 +1,79 @@
...
+    if args.dnrev is None:
+        args.dnrev = raw_input("Downstream part revision:  ")
+    if args.dnport is None:
+        args.dnport = raw_input("Downstream input port:  ")
+    if args.date == 'now':
+        args.date = cm_utils._query_default('date', args)
+    return args
+
+
+if __name__ == '__main__':
```
pkgw: It can be convenient to group all of the `__main__` functionality in a single function so that the bottom of the script ends up looking like this:

```
if __name__ == '__main__':
    main()
```

scripts/stop_connection.py
```diff
@@ -0,0 +1,79 @@
...
+    chk = handling.get_specific_connection(c, at_date)
+    if len(chk) == 1 and chk[0].stop_gpstime is None:
+        connection_start_was = chk[0].start_gpstime
+        print('Stopping connection {}:{}:{} <-> {}:{}:{} : {}'
+              .format(args.uppart, args.uprev, args.upport, args.dnpart,
+                      args.dnrev, args.dnport, connection_start_was))
+        go_ahead = True
+    else:
+        print("Error:  Connection to stop is not valid.  Quitting.")
+        go_ahead = False
```
pkgw: ... for instance, if the `__main__` functionality were in a function, the flow of this code could be simplified as:

```
chk = ...

if len(chk) != 1 or chk[0].stop_gpstime is not None:
    print("error: ...")
    return

connection_start_was = ...
print(...)
npc = ...
part_connect.stop_connecitons(...)
```

PR #88, opened 2017-08-08
corrected feature that made running add_connection as a script difficult

david-deboer:
The add_connection script would always query for the date, which is annoying when you have a script running a couple of hundred times...

### DIFF
--- a/scripts/add_connection.py
+++ b/scripts/add_connection.py
@@ -30,7 +30,7 @@ def query_args(args):
         args.dnrev = raw_input('Downstream part revision:  ')
     if args.dnport is None:
         args.dnport = raw_input('Downstream input port:  ')
-    if args.date == 'now':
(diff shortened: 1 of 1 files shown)


scripts/add_connection.py
```diff
@@ -30,7 +30,8 @@ def query_args(args):
         args.dnrev = raw_input('Downstream part revision:  ')
     if args.dnport is None:
         args.dnport = raw_input('Downstream input port:  ')
-    args.date = cm_utils._query_default('date', args)
+    if args.date == 'now':
```
pkgw: I think it would be helpful to add a comment explaining that `"now"` is the default value of the *date* argument.

Even better, you could modify `cm_utils.add_date_time_args` to take a keyword argument that lets caller specify the default value for the *date* argument — my interpretation is that in this case, the standard default of "now" is not appropriate.

PR #90, opened 2017-08-08
added small gpstime finder script

david-deboer:
Added a trivial gpstime thing with hera_mc


pkgw:
Even though the script content isn't really specific to M&C, I think it would be good to name it `cm_gpstime.py` or something else with a prefix indicating its origin. All of these scripts get installed into the user's `bin` directory and without namespacing, it becomes pretty much impossible to figure out where a given script came from — witness the mess of names used by Aipy.

Alternatively, it might be time for us to just install one driver script named `heramc` or something, and set it up to provide the various scripted functions through sub-commands the way Git works. That way gives you naming flexibility without the problem of dozens of random programs cluttering up `bin`.

bhazelton:
I agree, although I'd suggest `mc_gpstime.py` or `heramc_gpstime.py` since this is really not tied to the config. management stuff.

I'm interested in that last suggestion, Peter, but I know very little about how to do it.

bhazelton:
I'm struggling to understand what's specific about this to configuration management. I'd really prefer a prefix that didn't suggest this was about configuration management.

scripts/mc_gps.py
```diff
@@ -0,0 +1,19 @@
+#! /usr/bin/env python
+# -*- mode: python; coding: utf-8 -*-
+# Copyright 2016 the HERA Collaboration
+# Licensed under the 2-clause BSD license.
+
+"""Returns info from Time for given date/time
```
bhazelton: It doesn't actually *return* anything, it just prints some info, right?

PR #91, opened 2017-08-09
Cm fix hookup

david-deboer:
This fixes hookup (even gets rid of the "messy" method!!) and adds full hookup start/time times, which will be used to fix how antennas are viewed as "active" (or "connected" -- we are going to specifically call the full thing "sky-ready").

### DIFF
--- a/hera_mc/cm_hookup.py
+++ b/hera_mc/cm_hookup.py
@@ -18,7 +18,7 @@ import copy
 
 class Hookup:
     """
-    Class to find and display the hookup.
+    Class to find and display the signal path hookup.
     """
 
     def __init__(self, session=None):
@@ -77,6 +77,7 @@ class Hookup:
                 if len(hookup_dict['hookup'][k][pol]) > col_len_max[0]:
                     col_len_max[0] = len(hookup_dict['hookup'][k][pol])
                     col_len_max[1] = k
+        hookup_dict = self.__add_hookup_timing(hookup_dict)
 
         hookup_dict['columns'] = self.__get_column_header(
             hookup_dict['hookup'][col_len_max[1]][pol])
@@ -172,6 +173,23 @@ class Hookup:
                     next_one = opc
         return next_one
 
+    def __add_hookup_timing(self, hookup_dict):
+        really_late = 9999999999
+        hookup_dict['timing'] = {}
+        for akey, hk in hookup_dict['hookup'].iteritems():
+            hookup_dict['timing'][akey] = {}
+            for pkey, pol in hk.iteritems():
+                latest_start = 0
+                earliest_stop = really_late
+                for c in pol:
+                    if c.start_gpstime > latest_start:
+                        latest_start = c.start_gpstime
+                    if c.stop_gpstime is not None and c.stop_gpstime < earliest_stop:
+                        earliest_stop = c.stop_gpstime
+                if earliest_stop == really_late:
+                    earliest_stop = None
+                hookup_dict['timing'][akey][pkey] = [latest_start, earliest_stop]
+        return hookup_dict
 
     def __get_column_header(self, hup0):
         parts_col = []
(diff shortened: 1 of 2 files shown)


hera_mc/cm_hookup.py
```diff
@@ -173,46 +64,128 @@ def get_hookup(self, hpn, rev, port, at_date, state_args, exact_match=False):
...
         hookup_dict['columns'] = self.__get_column_header(
-            hookup_dict['hookup'][col_len_max[1]])
+            hookup_dict['hookup'][col_len_max[1]][pol])
         if state_args['show_levels']:
             hookup_dict = self.__hookup_add_correlator_levels(
                 hookup_dict, state_args['levels_testing'])
+        return hookup_dict
+
+    def __get_pols_to_do(self, part, port, check_pol=False):
+        all_pols = ['e', 'n']
```
bhazelton: We are starting to run into a collision between 'E' and 'N' vs 'x' and 'y', partly because of what is supported by data file formats. I don't want to hold up this PR on this point, but we need a discussion to decide what to do about it. @adampbeardsley @dannyjacobs

hera_mc/cm_hookup.py
```diff
@@ -173,46 +64,128 @@ def get_hookup(self, hpn, rev, port, at_date, state_args, exact_match=False):
...
+        hookup_dict['timing'] = {}
+        for akey, hk in hookup_dict['hookup'].iteritems():
+            hookup_dict['timing'][akey] = {}
+            for pkey, pol in hk.iteritems():
+                latest_start = 0
+                earliest_stop = really_late
+                for c in pol:
+                    if c.start_gpstime > latest_start:
+                        latest_start = c.start_gpstime
+                    if c.stop_gpstime is not None and c.stop_gpstime < earliest_stop:
```
bhazelton: Why do you need this earliest_stop/really_late logic? If stop_gpstime is either a valid date or None, then the test for None should be adequate, right?

hera_mc/cm_hookup.py
```diff
@@ -231,9 +204,13 @@ def __get_column_header(self, hup0):
                                                        exact_match=True)
         pr_key = cm_utils._make_part_key(hu.downstream_part, hu.down_part_rev)
         parts_col.append(get_part_type[pr_key]['part'].hptype)
+        parts_col.append('Start')
+        parts_col.append('Stop')
         return parts_col
 
     def __hookup_add_correlator_levels(self, hookup_dict, testing):
+        print("CM_HOOKUP[208]:  Adding correlator levels doesn't work with new pol hookup scheme yet.")
```
bhazelton: I think this kind of thing should be a warning. In general, I'd like to avoid print statements that are warning about unexpected or adverse behavior -- they should really be warnings.

hera_mc/cm_hookup.py
david-deboer: This is a project decision from some time ago - that configuration management would use E and N.  It is now embedded in (almost) every connection.  'E' then maps to 'x' and 'N' to 'y'.

bhazelton: My concern is that I'm not at all sure it's getting handled the same in the data processing codes and I know that the hera_qm tables are using 'x' and 'y', in large part because that's what the files use (and that's partly because of the limitations of the file formats). I just really want to make sure everyone is on the same page here and to decide where we're using which nomenclature.

hera_mc/cm_hookup.py
bhazelton: Ok, I think I see -- the really_late value is just a starting point and it can go down from there. I think it works ok, it's just a little more confusing to read than if earliest_stop starts out as None and only gets an assignment if there is a stoptime.

hera_mc/cm_hookup.py
```diff
@@ -209,7 +209,7 @@ def __get_column_header(self, hup0):
         return parts_col
 
     def __hookup_add_correlator_levels(self, hookup_dict, testing):
-        print("CM_HOOKUP[208]:  Adding correlator levels doesn't work with new pol hookup scheme yet.")
+        print("Warning:  correlator levels don't work with new pol hookup scheme yet (CM_HOOKUP[212]).")
```
bhazelton: Sorry,  I should have been more clear. Issue a warning via the `warning` module, don't use the print command at all. You can see an example in `mc_session.py` on line 916.

PR #94, opened 2017-08-10
made generic stop_part function

david-deboer:
I need this script and method to fix up the database for some parts which should be shutdown

### DIFF
--- a/hera_mc/part_connect.py
+++ b/hera_mc/part_connect.py
@@ -70,15 +70,14 @@ class Parts(MCDeclarativeBase):
             setattr(self, key, value)
 
 
-def stop_existing_parts(session, p, hpnr_list, at_date, actually_do_it):
+def stop_existing_parts(session, hpnr_list, at_date, actually_do_it):
     """
     This adds stop times to the previous parts.
 
     Parameters:
     ------------
     session:  db session to use
-    p:  part object
-    hpnr_list:  list containing hpn and revision
+    hpnr_list:  list containing hpn
     at_date:  date to use for stopping
     actually_do_it:  boolean to allow the part to be stopped
     """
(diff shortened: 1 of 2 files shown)


scripts/stop_part.py
```diff
@@ -0,0 +1,51 @@
+#! /usr/bin/env python
+# -*- mode: python; coding: utf-8 -*-
+# Copyright 2017 the HERA Collaboration
+# Licensed under the 2-clause BSD license.
+
+"""
+Script to add a general connection to the database.
```
pkgw: This docstring does not look right.

scripts/stop_part.py
```diff
@@ -0,0 +1,51 @@
...
+
+def query_args(args):
+    """
+    Gets information from user
+    """
+    if args.part is None:
+        args.part = raw_input("Part number:  ")
+    if args.rev is None:
+        args.rev = raw_input("Revision:  ")
+    if args.date == 'now':
```
pkgw: Should have a comment indicating that `"now"` is the default arg value, and for this particular script we want to require the user to specify a date that isn't now.

scripts/stop_part.py
```diff
@@ -0,0 +1,51 @@
...
+    parser.add_argument('--actually_do_it', help="Flag to actually do it, as "
+                        "opposed to printing out what it would do.",
+                        action='store_true')
+    cm_utils.add_date_time_args(parser)
+    args = parser.parse_args()
+
+    args = query_args(args)
+
+    # Pre-process some args
+    if args.date is not None:
```
pkgw: Because of `query_args`, `args.date` will never be None, right? This code won't work otherwise since the `at_date` variable won't get set.

hera_mc/part_connect.py
```diff
@@ -70,15 +70,14 @@ def part(self, **kwargs):
...
+def stop_existing_parts(session, hpnr_list, at_date, actually_do_it):
     """
     This adds stop times to the previous parts.
 
     Parameters:
     ------------
     session:  db session to use
-    p:  part object
-    hpnr_list:  list containing hpn and revision
+    hpnr_list:  list containing hpn
```
pkgw: It seems as if the old docstring for `hpnr_list` was more precise. Is there a terminology change here?

scripts/stop_part.py
david-deboer: Thanks - I've addressed your comments.  The --date now thing is somewhat tricky since date is "universal", but I think I've described it in the header comment.

PR #95, opened 2017-08-11
Tweak conn

david-deboer:
Fixed up a few bugs in connections and hookups

### DIFF
--- a/hera_mc/cm_handling.py
+++ b/hera_mc/cm_handling.py
@@ -327,9 +327,9 @@ class Handling:
                         connection_dossier['connections'][ckey] = copy.copy(conn)
                         up_parts.append(ckey)
                 if len(up_parts) == 0:
-                    up_parts = [self.no_connection_designator]
+                    up_parts = [PC.no_connection_designator]
                 if len(down_parts) == 0:
-                    down_parts = [self.no_connection_designator]
+                    down_parts = [PC.no_connection_designator]
                 up_parts.sort()
                 down_parts.sort()
                 if len(up_parts) > len(down_parts):
@@ -363,21 +363,21 @@ class Handling:
                        'Part', '<Output:', ':Input>', 'Downstream', 'dStart', 'dStop']
         for ordered_pairs in connection_dossier['ordered-pairs']:
             for up, dn in zip(ordered_pairs[0], ordered_pairs[1]):
-                if (self.no_connection_designator in up and
-                        self.no_connection_designator in dn):
+                if (PC.no_connection_designator in up and
+                        PC.no_connection_designator in dn):
                     continue
                 already_shown.append(up)
                 already_shown.append(dn)
                 tdata = range(0, len(headers))
                 # Do upstream
-                if up==self.no_connection_designator:
+                if up == PC.no_connection_designator:
                     connup = PC.get_null_connection()
                 else:
                     connup = connection_dossier['connections'][up]
                 uup = connup.upstream_part + ':' + connup.up_part_rev
-                if self.no_connection_designator in uup:
-                    start_date = self.no_connection_designator
-                    stop_date = self.no_connection_designator
+                if PC.no_connection_designator in uup:
+                    start_date = PC.no_connection_designator
+                    stop_date = PC.no_connection_designator
                 else:
                     start_date = connup.start_date
                     stop_date = connup.stop_date
@@ -405,22 +405,22 @@ class Handling:
                     del tdata[pos['Part'][vb]]
                     tdata.insert(pos['Part'][vb], '[' + udn + ']')
                 # Do downstream
-                if dn == self.no_connection_designator:
+                if dn == PC.no_connection_designator:
                     conndn = PC.get_null_connection()
                 else:
                     conndn = connection_dossier['connections'][dn]
                 dup = conndn.upstream_part + ':' + conndn.up_part_rev
                 ddn = conndn.downstream_part + ':' + conndn.down_part_rev
-                if self.no_connection_designator in ddn:
-                    start_date = self.no_connection_designator
-                    stop_date = self.no_connection_designator
+                if PC.no_connection_designator in ddn:
+                    start_date = PC.no_connection_designator
+                    stop_date = PC.no_connection_designator
                 else:
                     start_date = conndn.start_date
                     stop_date = conndn.stop_date
                 pos = {'Part': {'h': 5, 'm': 3}, 'Output': {'h': 6, 'm': 4},
                        'Input': {'h': 7, 'm': 5}, 'Downstream': {'h': 8, 'm': 6},
                        'dStart': {'h': 9, 'm': -1}, 'dStop': {'h': 10, 'm': -1}}
-                if self.no_connection_designator in udn:
+                if PC.no_connection_designator in udn:
                     if pos['Part'][vb] > -1:
                         del tdata[pos['Part'][vb]]
                         tdata.insert(pos['Part'][vb], '[' + dup + ']')
@@ -504,9 +504,9 @@ class Handling:
                         output_ports = pd[pa]['output_ports']
                         found_connection = True
             if len(input_ports) == 0:
-                input_ports = [self.no_connection_designator]
+                input_ports = [PC.no_connection_designator]
             if len(output_ports) == 0:
-                output_ports = [self.no_connection_designator]
+                output_ports = [PC.no_connection_designator]
             self.part_type_dict[k]['input_ports'] = input_ports
             self.part_type_dict[k]['output_ports'] = output_ports
             self.part_type_dict[k]['revisions'] = sorted(found_revisions)
--- a/hera_mc/cm_handling.py
+++ b/hera_mc/cm_handling.py
@@ -24,7 +24,6 @@ class Handling:
     Class to allow various manipulations of parts and their properties etc.
     Things are manipulated/passed as dictionaries currently.
     """
-    no_connection_designator = PC.no_connection_designator
 
     def __init__(self, session=None):
         """
--- a/hera_mc/cm_handling.py
+++ b/hera_mc/cm_handling.py
@@ -24,7 +24,7 @@ class Handling:
     Class to allow various manipulations of parts and their properties etc.
     Things are manipulated/passed as dictionaries currently.
     """
-    no_connection_designator = '-X-'
+    no_connection_designator = PC.no_connection_designator
 
     def __init__(self, session=None):
         """
@@ -313,8 +313,7 @@ class Handling:
                                                              conn.start_gpstime)
                         connection_dossier['connections'][ckey] = copy.copy(conn)
                         down_parts.append(ckey)
-                # Find where the part is in the downward position, so identify its
-                # upward connection
+                # Find where the part is in the downward position, so identify its upward connection
                 for conn in self.session.query(PC.Connections).filter(
                         (PC.Connections.downstream_part == xhpn) &
                         (PC.Connections.down_part_rev == this_rev)):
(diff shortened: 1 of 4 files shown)


hera_mc/cm_hookup.py
```diff
@@ -62,28 +65,23 @@ def get_hookup(self, hpn, rev, port, at_date, state_args, exact_match=False):
...
-                    col_len_max[0] = len(hookup_dict['hookup'][k][pol])
-                    col_len_max[1] = k
-        hookup_dict = self.__add_hookup_timing(hookup_dict)
+        if len(hookup_dict['hookup'].keys()) > 0:
+            hookup_dict = self.__add_hookup_timing(hookup_dict)
+            hookup_dict['columns'] = self.__get_column_headers(hookup_dict['hookup'])
+            if state_args['show_levels']:
+                hookup_dict = self.__hookup_add_correlator_levels(
+                    hookup_dict, state_args['levels_testing'])
+        else:
```
pkgw: I think it's easier to reason about if/else statements if the short clause comes first, since when you're reading the code that keeps the `else` closer to the original `if` proposition. In this case that would mean:

```python
if not len(hookup_dict['hookup'].keys()):
    hookup_dict['columns'] = []
else:
    hookup_dict = self.__add_hookup_timing(hookup_dict)
    ...
```

hera_mc/cm_hookup.py
```diff
@@ -188,28 +186,49 @@ def __add_hookup_timing(self, hookup_dict):
...
+                hu_col[hk][pk].append('Start')
+                hu_col[hk][pk].append('Stop')
+                # Next two lines for "one column" solution
+                if len(hu_col[hk][pk]) > long_column[-1]:
+                    long_column = [hk, pk, len(hu_col[hk][pk])]
+        return hu_col[long_column[0]][long_column[1]]
 
     def __hookup_add_correlator_levels(self, hookup_dict, testing):
-        warnings.warn("Warning:  correlator levels don't work with new pol hookup scheme yet (CM_HOOKUP[212]).")
+        warnings.warn("Warning:  correlator levels don't work with new pol hookup scheme yet (CM_HOOKUP[210]).")
```
pkgw: It doesn't seem super helpful to include the line numbers in these messages; this is why editors have search features.

hera_mc/cm_hookup.py
```diff
@@ -188,28 +186,49 @@ def __add_hookup_timing(self, hookup_dict):
...
-                                                       exact_match=True)
-        pr_key = cm_utils._make_part_key(hu.downstream_part, hu.down_part_rev)
-        parts_col.append(get_part_type[pr_key]['part'].hptype)
-        parts_col.append('Start')
-        parts_col.append('Stop')
-        return parts_col
+    def __get_column_headers(self, huh):
+        """
+        This gets the full set of headers for a future show_hookup that doesn't require the same
+        hookup starting point.  Returns a single column for now (the one/long_column stuff)
```
pkgw: By reading this docstring I don't feel like I understand what the argument is or what the return value is. Fundamentally it's not clear to me what you mean by a "column".

hera_mc/cm_hookup.py
```diff
@@ -188,28 +186,49 @@ def __add_hookup_timing(self, hookup_dict):
...
+        """
+        hu_col = {}
+        return_one_column_for_now = True
+        if return_one_column_for_now:  # This is to just not find part_dossiers on everything for now
+            lc = [huh.keys()[0], huh[huh.keys()[0]].keys()[0], 0]  # ...and maybe we'll use the full
```
pkgw: I think this and the analogous logic below would be clearer if you split `lc` into two variables: `column_data` and `max_col_name_len`, or something like that. It's confusing to combine these two pieces of information into one variable using a list.

hera_mc/cm_hookup.py
```diff
@@ -188,28 +186,49 @@ def __add_hookup_timing(self, hookup_dict):
...
-        parts_col.append('Stop')
-        return parts_col
+    def __get_column_headers(self, huh):
+        """
+        This gets the full set of headers for a future show_hookup that doesn't require the same
+        hookup starting point.  Returns a single column for now (the one/long_column stuff)
+        """
+        hu_col = {}
+        return_one_column_for_now = True
+        if return_one_column_for_now:  # This is to just not find part_dossiers on everything for now
```
pkgw: It seems to me like you're trying to future-proof since you know you're going to change this code later. I think it is better to write the code to do what it does right now in the most direct way possible, then change it later when it needs to do something different. You're not going to forget about the need for the future change, unless it turns out that you don't actually need to do it! And it doesn't look like this is a case where there's tricky logic that you've worked out that deserves to preserved so you don't have to spend time recreating it later.

hera_mc/cm_hookup.py
```diff
@@ -188,28 +186,49 @@ def __add_hookup_timing(self, hookup_dict):
...
+            lc = [huh.keys()[0], huh[huh.keys()[0]].keys()[0], 0]  # ...and maybe we'll use the full
+            for hk, hu in huh.iteritems():                        # ...dictionary later
+                for pk, pv in hu.iteritems():
+                    if len(pv) > lc[-1]:
+                        lc = [hk, pk, len(pv)]
+            huh = {hk: {pk: huh[lc[0]][lc[1]]}}
+
+        long_column = [huh.keys()[0], huh[huh.keys()[0]].keys()[0], 0]  # Pick first one to start
```
pkgw: Some more comments would be helpful. I can't figure out what the following stanza is doing and why.

hera_mc/cm_hookup.py
```diff
@@ -62,28 +65,23 @@ def get_hookup(self, hpn, rev, port, at_date, state_args, exact_match=False):
...
             for pol in pols_to_do:
                 hookup_dict['hookup'][k][pol] = self.__follow_hookup_stream(part['part'].hpn, part['part'].hpn_rev, pol)
-                if len(hookup_dict['hookup'][k][pol]) > col_len_max[0]:
-                    col_len_max[0] = len(hookup_dict['hookup'][k][pol])
-                    col_len_max[1] = k
-        hookup_dict = self.__add_hookup_timing(hookup_dict)
+        if len(hookup_dict['hookup'].keys()) > 0:
+            hookup_dict = self.__add_hookup_timing(hookup_dict)
+            hookup_dict['columns'] = self.__get_column_headers(hookup_dict['hookup'])
+            if state_args['show_levels']:
```
david-deboer: I'm fixing this in a broader context in another branch.

hera_mc/cm_hookup.py
david-deboer: The difference between now and future in this module is not great, so I'd like to keep it as is.  The bigger difference is to the show_hookup method, which needs to be better anyway (but is a bigger job than I'd like to tackle right now).

hera_mc/cm_hookup.py
david-deboer: I made a little utility class for this.

hera_mc/cm_hookup.py
david-deboer: I like to know file/line numbers...  Harmless carrying of a few bytes that can save minutes/frustration

hera_mc/cm_handling.py
```diff
@@ -24,7 +24,7 @@ class Handling:
     Class to allow various manipulations of parts and their properties etc.
     Things are manipulated/passed as dictionaries currently.
     """
-    no_connection_designator = '-X-'
+    no_connection_designator = PC.no_connection_designator
```
pkgw: You should be able to delete this line now, I think.

hera_mc/cm_hookup.py
```diff
@@ -37,13 +38,20 @@ def __init__(self, session=None):
...
+            'timing':  valid times for the corresponding hookup in dictionary
+            'levels':  correlator levels, if desired
         This only gets the contemporary hookups (unlike parts and connections,
-            which get all.)
+            which get all.)  That is, only hookups valid at_date are returned.
+        Note that this also assumes all hookups returned are connected at the same
+            furthest upstream place.  This means that the columns heading won't match
+            if valid hookups for a list start at different upstream locations.  The plan
+            is to fix that.
+
```
pkgw: Thanks, this is more helpful now.

hera_mc/cm_hookup.py
```diff
@@ -188,28 +191,54 @@ def __add_hookup_timing(self, hookup_dict):
...
-        parts_col.append('Start')
-        parts_col.append('Stop')
-        return parts_col
+    def __get_column_headers(self, huh):
+        """
+        The columns in the hookup_dict contain parts in the hookup chain and the column headers are 
+        the part types contained in that column.  This returns the headers for the retrieved hookup.
+        This gets the full set of headers for a future show_hookup that doesn't require the same
+        hookup starting point.  Returns a single column for now (the one/long_column stuff).
+        The method searches all of the hookup chains to find the longest one, which occurs at
```
pkgw: Unfinished sentence?

hera_mc/cm_hookup.py
```diff
@@ -188,28 +191,54 @@ def __add_hookup_timing(self, hookup_dict):
...
+                hu_col[hk][pk].append('Start')
+                hu_col[hk][pk].append('Stop')
+                # Next two lines for "one column" solution
+                if len(hu_col[hk][pk]) >= lc.hlen:
+                    lc.v(part=hk, pol=pk, hlen=len(hu_col[hk][pk]))
+        return hu_col[lc.part][lc.pol]
 
     def __hookup_add_correlator_levels(self, hookup_dict, testing):
-        warnings.warn("Warning:  correlator levels don't work with new pol hookup scheme yet (CM_HOOKUP[212]).")
+        warnings.warn("Warning:  correlator levels don't work with new pol hookup scheme yet (CM_HOOKUP[210]).")
```
pkgw: For what it's worth, I'm not concerned about the space taken up by the line numbers in messages like these, it's that they get out of date easily — for instance, if GitHub is telling me the right thing, this message is now on line 241. And updating the line numbers as the files change leads to distracting elements in the diffs.

hera_mc/cm_utils.py
```diff
@@ -16,6 +16,17 @@
 PAST_DATE = '2000-01-01'
 
 
+class utility_method:
```
pkgw: I think what you're making here is what's generally referred to as a namespace object ([see eg this StackExchange question](https://stackoverflow.com/questions/20828277/what-is-a-namespace-object)). It would be good to re-use the existing [argparse.Namespace](https://docs.python.org/dev/library/argparse.html#argparse.Namespace) class using the standard nomenclature here.

hera_mc/cm_hookup.py
david-deboer: View the line number as the initial serial number, which need not change.  Having that tells you which file it is in, then you can unambiguously search for the cryptic string.

pkgw: But you just changed it! :-)

david-deboer: I promise I won't ever again.

hera_mc/cm_utils.py
david-deboer: Is there a risk or impact of the way I did it?  I like it because it is so simple and flexible.  You can add arguments so compactly etc.  With the namespace you add_argument for everything, you need the '--'(or not?), etc.  Unless there is a reason to, I'd like to leave it.  Thoughts?

pkgw: While the whole argparse module is tuned to arg parsing, its Namespace class is a generic helper that's *exactly* what you're creating. It can be initialized pretty much the way you use yours:

```
>>> from argparse import Namespace
>>> ns = Namespace(foo='bar', other=123)
>>> ns
  < Namespace(foo='bar', other=123)
>>> ns.foo
  < 'bar'
```

The advantage of the existing Namespace object is that it reuses existing infrastructure instead of reimplementing it, and uses standard terminology, and has built-in support for other standard operations like str() and `==`.

PR #98, opened 2017-08-12
Cm test cover

david-deboer:
Fixed my testing errors and implemented a few new tests.

### DIFF
--- a/hera_mc/cm_revisions.py
+++ b/hera_mc/cm_revisions.py
@@ -13,7 +13,7 @@ from tabulate import tabulate
 from hera_mc import cm_utils, part_connect
 
 
-def get_revisions_of_type(rev_type, hpn, session=None):
+def get_revisions_of_type(rev_type, hpn, at_date=None, session=None):
     """
     Returns list of revisions of type
 
@@ -28,7 +28,11 @@ def get_revisions_of_type(rev_type, hpn, session=None):
     if rq == 'LAST':
         rq = get_last_revision(hpn, session)
     elif rq == 'ACTIVE':
-        rq = get_contemporary_revision(hpn, session)
+        if at_date is None:
+            print("Need a date to find active revision.")
+            rq = None
+        else:
+            rq = get_contemporary_revision(hpn, at_date, session)
     elif rq == 'ALL':
         rq = get_all_revisions(hpn, session)
     else:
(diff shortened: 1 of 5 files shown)


hera_mc/tests/test_connections.py
```diff
@@ -15,82 +12,71 @@
...
-        self.part.start_gpstime = self.test_time
-        self.part.hpn = self.test_part1
-        self.test_session.add(self.part)
-        self.test_session.commit()
-        self.part.hpn = self.test_part2
-        self.test_session.add(self.part)
+        self.query_time = Time('2017-08-01 01:00:00', scale='utc')
+
+        for tp in self.test_hpn:
+            self.part = part_connect.Parts()
```
pkgw: Is there any need to assign to the class attribute `self.part`, instead of just a variable `part`? I see the same pattern not just at this particular location but in most of the functions in this file.

pkgw:
"Coverage increased (+3.8%) to 56.547%" nice!

hera_mc/tests/test_connections.py
david-deboer: That is a bit of a relic, given the "silent failure" error of putting stuff in SetUp.  I'll clean up a bit now.

david-deboer: ---the 'relic' being my brain...

hera_mc/cm_revisions.py
```diff
@@ -28,7 +28,11 @@ def get_revisions_of_type(rev_type, hpn, session=None):
     if rq == 'LAST':
         rq = get_last_revision(hpn, session)
     elif rq == 'ACTIVE':
-        rq = get_contemporary_revision(hpn, session)
+        if at_date is None:
+            print("Need a date to find active revision.")
```
pkgw: What's the outcome if `at_date is None`? I.e., what does a return value of `None` cause to happen? If the intention is that the program should abort, this should raise an exception instead of printing a message and returning None. If some backup behavior happens, the message should describe it; i.e., "Need a date to find the \"active\" revision; using the \"last\" revision instead", or something.

Stylistically, I like to structure these kinds of functions like this:

```python
def get_revisions_of_type(rev_type, hpn, at_date=None, session=None):
    rq = rev_type.upper()
    if rq == 'LAST':
        return get_last_revision(hpn, session)

    if rq == 'ACTIVE':
        if at_date is None:
            print("Need a date to find active revision.")
            return None
        else:
            return get_contemporary_revision(hpn, at_date, session)

    if rq == 'ALL':
        return get_all_revisions(hpn, session)

    return get_particular_revision(rq, hpn, session)
```

PR #99, opened 2017-08-12
Cm kwargs

david-deboer:
Fixed the show_args/state_args nonsense and added a couple of tests.


hera_mc/cm_hookup.py
```diff
@@ -57,11 +57,14 @@ def get_hookup(self, hpn, rev, port, at_date, state_args, exact_match=False):
...
-        port_:  a specifiable port name,  default is 'all'.  Unverified.
+        rev:  the revision number
+        port:  a specifiable port name,  default is 'all'.  Unverified.
         at_date:  date for hookup validity
-        state_args:  keyword dictionary specifying parameters for hookup
         exact_match:  boolean for either exact_match or partial
+        kwargs:  optional arguments to allow for inserting levels
+                    show_levels:  boolean to include levels
+                    levels_testing:  if present and not False will pull levels from
+                                     the file given in levels_testing
```
pkgw: Why are these options accessed through `**kwargs` and not simply two new keyword arguments defaulting to `False`? From reading the function I can't see a reason for this convoluted approach.

scripts/parts.py
```diff
@@ -51,11 +43,8 @@
         elif args.levels_testing == 'levels.tst':
             args.levels_testing = os.path.join(mc.test_data_path, 'levels.tst')
 
-    state_args = {'verbosity':args.verbosity.lower(),
-                  'mapr_cols':args.mapr_cols.lower(),
-                  'show_levels':args.show_levels,
-                  'show_state':args.show_state,
-                  'levels_testing':args.levels_testing}
+    state_args = {'verbosity':args.verbosity,
+                  'show_state':args.show_state}
```
pkgw: The `state_args` variable does not appear to be used at all in this file anymore, so you can just delete it altogether.

david-deboer: I made the two changes (I thought I had removed the state_args thing already...)

hera_mc/cm_hookup.py
```diff
@@ -84,9 +86,10 @@ def get_hookup(self, hpn, rev, port, at_date, state_args, exact_match=False):
         else:
             hookup_dict = self.__add_hookup_timing(hookup_dict)
             hookup_dict['columns'] = self.__get_column_headers(hookup_dict['hookup'])
-            if state_args['show_levels']:
-                hookup_dict = self.__hookup_add_correlator_levels(
-                    hookup_dict, state_args['levels_testing'])
+            if 'show_levels' in kwargs.keys() and kwargs['show_levels'] is True:
```
pkgw: This chunk of the code still references the `kwargs` variable, so it will crash if this code path is triggered.

david-deboer: That was careless.

PR #100, opened 2017-08-13
Update RTP table definition

plaplant:
This PR updates the RTP tables definitions. In addition to the git version and hash for RTP itself, a process record should also include the version and hash of other key software components, namely `hera_qm`, `hera_cal`, and `pyuvdata`. The column definitions, table interaction functions, and tests have been updated to reflect this change.


alembic/versions/8e7282ae4840_updated_rtp_tables.py
```diff
@@ -0,0 +1,55 @@
...
+
+# revision identifiers, used by Alembic.
+revision = '8e7282ae4840'
+down_revision = 'e66c069eb92b'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
```
pkgw: Should delete the "autogenerated" comments if the DB operations have been vetted. They look sensible enough (could imaging renaming the `git_hash` column to `rtp_git_hash` but we don't have any rows so whatever).

alembic/versions/8e7282ae4840_updated_rtp_tables.py
```diff
@@ -0,0 +1,55 @@
...
+    op.add_column('rtp_process_record', sa.Column('hera_cal_git_hash', sa.String(length=64), nullable=False))
+    op.add_column('rtp_process_record', sa.Column('hera_cal_git_version', sa.String(length=32), nullable=False))
+    op.add_column('rtp_process_record', sa.Column('hera_qm_git_hash', sa.String(length=64), nullable=False))
+    op.add_column('rtp_process_record', sa.Column('hera_qm_git_version', sa.String(length=32), nullable=False))
+    op.add_column('rtp_process_record', sa.Column('pyuvdata_git_hash', sa.String(length=64), nullable=False))
+    op.add_column('rtp_process_record', sa.Column('pyuvdata_git_version', sa.String(length=32), nullable=False))
+    op.add_column('rtp_process_record', sa.Column('rtp_git_hash', sa.String(length=64), nullable=False))
+    op.add_column('rtp_process_record', sa.Column('rtp_git_version', sa.String(length=32), nullable=False))
+    op.drop_column('rtp_process_record', 'git_hash')
+    op.drop_column('rtp_process_record', 'git_version')
```
pkgw: Does this table potentially have an unwanted autoincrement setup on its `time` column? It's got a multi-column primary key so it shouldn't, wright?

hera_mc/tests/test_rtp.py
```diff
@@ -284,7 +287,7 @@ def test_errors_rtp_process_record(self):
         self.assertTrue(len(obs_result), 1)
 
         self.assertRaises(ValueError, self.test_session.add_rtp_process_record, 'foo',
-                          *self.status_values[1:])
+                          *((self.status_values)*2))
```
pkgw: Shouldn't this be something like `*self.record_values`?

alembic/versions/8e7282ae4840_updated_rtp_tables.py
plaplant: Okay, I'll rename the columns instead of deleting and making new ones.

alembic/versions/8e7282ae4840_updated_rtp_tables.py
```diff
@@ -0,0 +1,55 @@
...
+# revision identifiers, used by Alembic.
+revision = '8e7282ae4840'
+down_revision = 'e66c069eb92b'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.drop_table('host_status')
```
plaplant: Now that I'm looking at this file a little more carefully, should I be dropping this table?

alembic/versions/8e7282ae4840_updated_rtp_tables.py
plaplant: It shouldn't have the auto-increment. How can I see if it does or not?

hera_mc/tests/test_rtp.py
plaplant: This is just an array of nonsense values passed in to cause an error. I made it longer given the greater number of arguments

alembic/versions/8e7282ae4840_updated_rtp_tables.py
pkgw: Most direct test is to look at the table schema onsite.

alembic/versions/8e7282ae4840_updated_rtp_tables.py
pkgw: Actually I suppose not. That table's been superseded, but I didn't drop it from the Karoo so that we could preserve its data.

If Alembic is going to want to drop the table every time we change the DB schema, though, we should come up with a solution. Most sensible might just be to back up the database contents with `pg_dump` then drop the table from the running DB.

hera_mc/tests/test_rtp.py
pkgw: I think it'd be clearer not to re-use `status_values` for this purpose. Also, the inner parentheses aren't needed.

alembic/versions/8e7282ae4840_updated_rtp_tables.py
plaplant: Looking at the schema onsite, the `time` column is a `bigint` type, and there is no `serial` attribute. So I think it's correct.

hera_mc/tests/test_rtp.py
plaplant: I made a new dummy array for this.

alembic/versions/8e7282ae4840_updated_rtp_tables.py
plaplant: That seems like a reasonable solution. In the meantime, I took the `drop_table` statement out of the `upgrade()` command, and the table addition out of the `downgrade()` command. Let me know if I should change it back.

PR #101, opened 2017-08-14
Fixrev

david-deboer:
This much needed update fixes the way that part revisions are handled.  It cleans up a few other small things and rewrites parts.py to be not so confusing.


hera_mc/cm_revisions.py
```diff
@@ -9,57 +9,106 @@
...
+            if revisions[first_one].ended is None:
+                revisions[first_one].ended = cm_utils._get_astropytime('>')
+            if revisions[second_one].started > revisions[first_one].ended:
+                overlap.append([r1, r2])
+    if len(overlap) > 0:
+        overlapping_revs_in_single_list = []
+        for ol in overlap:
+            overlapping_revs_in_single_list.append(ol[0])
+            overlapping_revs_in_single_list.append(ol[1])
+            print('\n\tWarning!!!  {} and {} are overlapping revisions of part {}\n\n'.format(
```
bhazelton: Should this be a warning or a print statement?

hera_mc/tests/test_connections.py
```diff
@@ -88,5 +88,19 @@ def test_return_is_in_connections(self):
         print(checking)
```
bhazelton: I don't think you need this print statement

PR #102, opened 2017-08-15
Update RTP Status table

plaplant:
This PR adds the hostname as part of the RTP Status table, and uses the hostname + time as the primary key. This avoids having multiple hosts check in simultaneously and create a collision. The alembic commands were made by hand, so comments on them are most welcome.


bhazelton:
I'm confused by this, we have an RTP server status table to check on the status of the RTP servers, I thought this table was meant to be the high-level RTP status that was updated by the head node, detailing what was going on with RTP as a whole.

plaplant:
Apologies, I didn't realize the RTP Status table was supposed to be RTP as a whole. I'll close this.

PR #104, opened 2017-08-16
Correlator info

bhazelton:
Before this branch is ready we need to fix the `write_anntenna_locations_file.py` script.

Also some testing should be added for full connections that are no longer active.


bhazelton:
This branch also needs a fix to aipy to be in place (https://github.com/HERA-Team/aipy/pull/22)

bhazelton:
Ok, I think this PR is now ready for review. @pkgw and @jack-h can I get you to look at it?

hera_mc/cm_hookup.py
```diff
@@ -277,7 +282,18 @@ def __header_entry_name_adjust(self, col):
             colhead = col
         return colhead
 
-    def show_hookup(self, hookup_dict, cols_to_show, show_levels):
+    def get_correlator_input_from_hookup(self, hookup_dict, corr_name=PC.roach_input_name):
+        """
+        Retrieve the correlator inputs from a hookup_dictionary
+        """
+        corr_input = {}
+        if corr_name in hookup_dict['columns']:
```
pkgw: So, if `corr_name` is not in `hookup_dict['columns']`, this function will return `{}`. Are those the best semantics? Would it be better to raise an exception in this case?

hera_mc/cm_utils.py
```diff
@@ -135,6 +150,29 @@ def _get_astropytime(_date, _time=0):
     return return_date
 
 
+def get_date_from_pair(d1, d2, ret='earliest'):
+    if d1 is None and d2 is None:
+        return None
+    retdat = None
```
pkgw: I feel that functions like these are much clearer if one uses `return` statements rather than having many different paths that set a `retval`-type variable.

hera_mc/cm_utils.py
```diff
@@ -135,6 +150,29 @@ def _get_astropytime(_date, _time=0):
...
+            retdat = d1 if d1 < d2 else d2
+    elif ret == 'latest':
+        if d1 is None:
+            retdat = d1
+        elif d2 is None:
+            retdat = d2
+        else:
+            retdat = d1 if d1 > d2 else d2
+    else:
+        warnings.warn("No return comparison value given.")
```
pkgw: Does it make sense for this condition to be a warning? It seems likely to me that it would be better to raise a ValueError if `ret` is not one of the known values.

hera_mc/geo_handling.py
```diff
@@ -253,7 +259,7 @@ def find_station_of_antenna(self, antenna, query_date):
             raise ValueError('More than one active connection between station and antenna')
         return antenna_connected.upstream_part
 
-    def get_location(self, to_find, query_date, show_location=False, verbosity='m'):
+    def get_location(self, to_find, query_date, station_types=None, show_location=False, verbosity='m'):
```
pkgw: `station_types` should be added to this function's list of parameters in the docstring.

hera_mc/geo_handling.py
```diff
@@ -313,33 +322,292 @@ def get_location(self, to_find, query_date, show_location=False, verbosity='m'):
...
-        Returns a list of all of the locations ever connected.
+    def _search_loop(self, stn, at_date, now, m, base_tdelta, full_req):
+        loop = True
+        while loop:
+            at_date += TimeDelta(m * base_tdelta, format='sec')
+            if at_date > now:
+                at_date = now + TimeDelta(1.0, format='sec')
+            fc = cm_revisions.get_full_revision(stn, at_date, full_req, self.session)
+            if len(fc) == 1:
+                loop = False
```
pkgw: This function should use `break` statements instead of a loop control variable.

hera_mc/geo_handling.py
```diff
@@ -313,33 +322,292 @@ def get_location(self, to_find, query_date, show_location=False, verbosity='m'):
                     print(L, ' not found.')
         return found_location
 
-    def get_all_everconnected_locations(self):
-        """
-        Returns a list of all of the locations ever connected.
+    def _search_loop(self, stn, at_date, now, m, base_tdelta, full_req):
```
pkgw: I really do not feel good about this function. What is its purpose? Where do the various hardcoded magic time intervals come from? Why are `m` and `base_tdelta` separate arguments if they're only ever used in conjunction? I would like someone to double-check its internal logic and add comments documenting how this function works.

hera_mc/geo_handling.py
```diff
@@ -313,33 +322,292 @@ def get_location(self, to_find, query_date, show_location=False, verbosity='m'):
...
+        from sqlalchemy import asc
+        hookup = cm_hookup.Hookup(self.session)
+
+        base_tdelta = 86400.0
+        m = 10.0
+        self.get_station_types()
+        station_conn = []
+        now = cm_utils._get_astropytime('now')
+        for k, stn_type in self.station_types.iteritems():
+            if station_types_to_check == 'all' or k in station_types_to_check:
```
pkgw: I think the logic here should be inverted to reduce the amount of rightward drift in the code:

```python
if station_types_to_check != 'all' and k not in station_types_to_check:
    continue

for stn in ...
```

hera_mc/geo_handling.py
```diff
@@ -313,33 +322,292 @@ def get_location(self, to_find, query_date, show_location=False, verbosity='m'):
...
+                        else:
+                            station_dict = self.get_fully_connected_location_at_date(stn=stn, at_date=at_date,
+                                                                                     hookup=hookup, fc=fc,
+                                                                                     full_req=full_req)
+                            station_conn.append(station_dict)
+                            k0 = fc[0].hookup['timing'].keys()[0]
+                            tp = fc[0].hookup['timing'][k0]
+                            fc_strt = Time(cm_utils.get_date_from_pair(tp['e'][0], tp['n'][0], 'earliest'), format='gps')
+                            fc_stop = cm_utils.get_date_from_pair(tp['e'][1], tp['n'][1], 'earliest')
+                            if fc_stop is None:
```
pkgw: Change to `if fc_stop is not None:` and eliminate the `pass` branch.

hera_mc/geo_handling.py
```diff
@@ -313,33 +322,292 @@ def get_location(self, to_find, query_date, show_location=False, verbosity='m'):
...
+                    fc = cm_revisions.get_full_revision(stn, at_date, full_req, self.session)
+            elif at_date > now:
+                loop = False
+        return fc, at_date
+
+    def get_all_fully_connected_ever(self, earliest_date=Time('2016-09-01'),
+                                     full_req=part_connect.full_connection_parts_paper,
+                                     station_types_to_check='all'):
         """
+        Returns a list of all of the locations fully connected on at_date that
```
pkgw: The summary description in this docstring seems inaccurate.

hera_mc/geo_handling.py
```diff
@@ -313,33 +322,292 @@ def get_location(self, to_find, query_date, show_location=False, verbosity='m'):
...
-                connections = self.session.query(part_connect.Connections).filter(
-                    part_connect.Connections.upstream_part == stn.station_name)
-                for conn in connections:
-                    ant_num = int(conn.downstream_part[1:])
+        Parameters
+        -----------
+        earliest_date:  earliest_date to check for connections.
+        full_req: list of parts required for a connection to be a 'full' connection
+            (default: part_connect.full_connection_parts_paper)
+        station_types_to_check:  list of station types to limit check, or all (default: all)
```
pkgw: Should be written as `"all"` to emphasize that the value is a string constant.

hera_mc/geo_handling.py
```diff
@@ -313,33 +322,292 @@ def get_location(self, to_find, query_date, show_location=False, verbosity='m'):
...
+                    fc_stop = cm_utils.get_date_from_pair(tp['e'][1], tp['n'][1], 'earliest')
+                    if fc_stop is None or fc_stop > now.gps:
+                        continue
+                    fc_stop = Time(fc_stop, format='gps')
+                    at_date = fc_stop + TimeDelta(base_tdelta, format='sec')
+                    while fc_stop is not None:
+                        fc, at_date = self._search_loop(stn, at_date, now, m, base_tdelta, full_req)
+                        if len(fc) == 0:
+                            fc_stop = None
+                        else:
```
pkgw: I believe this can be changed to:

```python
if not len(fc):
    fc_stop = None
    break

station_dict = ...
```

Once again this reduces the amount of rightward drift in the code.

hera_mc/geo_handling.py
```diff
@@ -313,33 +322,292 @@ def get_location(self, to_find, query_date, show_location=False, verbosity='m'):
...
+                    if at_date > now:
+                        at_date = now + TimeDelta(1.0, format='sec')
+                    fc = cm_revisions.get_full_revision(stn, at_date, full_req, self.session)
+            elif at_date > now:
+                loop = False
+        return fc, at_date
+
+    def get_all_fully_connected_ever(self, earliest_date=Time('2016-09-01'),
+                                     full_req=part_connect.full_connection_parts_paper,
+                                     station_types_to_check='all'):
```
pkgw: In this function too, I worry about the soundness of the search algorithm. Something this complicated should have inline comments explaining how the algorithm works.

hera_mc/geo_handling.py
```diff
@@ -313,33 +322,292 @@ def get_location(self, to_find, query_date, show_location=False, verbosity='m'):
...
+        station_types_to_check:  list of station types to limit check, or all
+        """
+        from hera_mc import cm_hookup
+        hookup = cm_hookup.Hookup(self.session)
+
+        at_date = cm_utils._get_astropytime(at_date)
+        self.get_station_types()
+        station_conn = []
+        for k, stn_type in self.station_types.iteritems():
+            if (station_types_to_check == 'all' or k in station_types_to_check):
```
pkgw: Parentheses unnecessary here.

hera_mc/geo_handling.py
```diff
@@ -313,33 +322,292 @@ def get_location(self, to_find, query_date, show_location=False, verbosity='m'):
...
+        stn:  is the station information to check
+        at_date:  date to check for connections, must be an astropy.Time
+        hookup:  is to provide a hookup instance to not have to do it everytime
+        fc:  is the full hookup/revision, in case you already have it don't call it again
+        full_req:  list contining needed parts to be considered complete
+        """
+        if fc is None:
+            fc = cm_revisions.get_full_revision(stn, at_date, full_req, self.session)
+        station_dict = {}
+        if len(fc) == 1:
```
pkgw: Another logic inversion can reduce rightward drift:

```python
if not len(fc):
    return {}

hu = ...
```

hera_mc/geo_handling.py
```diff
@@ -313,33 +322,292 @@ def get_location(self, to_find, query_date, show_location=False, verbosity='m'):
...
+            hu = fc[0].hookup
+            k0 = hu['hookup'].keys()[0]
+            ant_num = hu['hookup'][k0]['e'][0].downstream_part
+            ant_num = int(str(ant_num).split('A')[1])
```
pkgw: What's the format of the `ant_num` value? If it's just (e.g.) `A101`, this could just be `ant_num = int(ant_num[1:])`.

hera_mc/geo_handling.py
```diff
@@ -313,33 +322,292 @@ def get_location(self, to_find, query_date, show_location=False, verbosity='m'):
...
+            eastings.append(stn['easting'])
+            northings.append(stn['northing'])
+            longitudes.append(stn['longitude'])
+            latitudes.append(stn['latitude'])
+            elevations.append(stn['elevation'])
+
+        ecef_positions = uvutils.XYZ_from_LatLonAlt(latitudes, longitudes, elevations)
+        rotecef_positions = uvutils.rotECEF_from_ECEF(ecef_positions.T, cofa_loc.lon)
+
+        corr_dict = {'antenna_numbers': ant_nums,
```
pkgw: This could be just `return {'antenna_numbers': ...`

hera_mc/geo_handling.py
bhazelton: It comes out as unicode (`u'A101'`) I experimented around until I got an int out, but I'll try your suggestion.

hera_mc/geo_handling.py
bhazelton: sure. I'll change it.

hera_mc/geo_handling.py
david-deboer: I've tried to give a description.  The main problem is the time resolution and trading off speed.  Fortunately now we will not be using this operationally, but are keeping it as a backup.

hera_mc/geo_handling.py
```diff
@@ -115,17 +116,19 @@ def cofa(self, show_cofa=False):
...
         self.get_station_types(add_stations=True)
         current_cofa = self.station_types['COFA']['Stations']
-        located = self.get_location(current_cofa, 'now', show_cofa, 'm')
+        located = self.get_location(current_cofa, 'now', self.station_types)
         if len(located) == 0:
             located_cofa = None
         elif len(located) > 1:
-            print("Warning:  {} has returned multiple cofa values.  Returning None.")
+            s = "{} has multiple cofa values.".format(str(current_cofa))
+            warnings.warn(s)
```
pkgw: If the presence of multiple cofa values is just a warning, should the code return the first one instead of None?

david-deboer: I've gone back and forth on this, but I agree with you.  It can give a warning.warn but return the first one.
I'm about to push up a small PR for making the parts stuff case-insenstive and I'll include this change there.  Thanks.

jack-h:
Second @pkgw's notes about comments. But these additions look like they fulfill the needs from the correlator side.

jack-h:
Though clearly I have no idea how to officially approve.

pkgw:
PR has already been approved and merged. But the process is a bit obscure, so for future reference: visit the “Files changed” tab for the PR and click “Review changes”. For un-merged PRs, you’ll have options to specify the summary outcome of your review: neutral, approve, request changes.

PR #105, opened 2017-08-19
fixed fem_swap script

david-deboer:
Just fixed a script to add new fems based on new cm_hookup inputs

### DIFF
--- a/scripts/fem_swap.py
+++ b/scripts/fem_swap.py
@@ -67,7 +67,7 @@ if __name__ == '__main__':
                            exact_match=True)
     k = hd['hookup'].keys()[0]
     if len(hd['hookup'][k]['e']) == 0:
-        print("<<<<<<<ERROR>>>>>>>")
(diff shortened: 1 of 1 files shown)


scripts/fem_swap.py
```diff
@@ -60,39 +59,47 @@ def query_args(args):
...
-        print("Stopping this swap.")
+        print("<<<<<<<Error>>>>>>>  {} is already connected".format(fem_hpn))
+        sys.exit()
+    # ###ERROR EXIT POINT
+
+    hd = hookup.get_hookup(hpn=args.ant, rev='H', port='all', at_date=at_date,
+                           exact_match=True)
+    k = hd['hookup'].keys()[0]
+    if len(hd['hookup'][k]['e']) == 0:
+        print("<<<<<<<ERROR>>>>>>>")
```
bhazelton: would a more descriptive message be useful here? Up to you.

david-deboer:
You're right - better to provide at least a lttle detail, which I did.

PR #108, opened 2017-08-21
hacked away most of the test cm db

david-deboer:
This hacks away most of the test db to speed up tests.  Still tests everything it did.


hera_mc/tests/test_geo_location.py
```diff
@@ -93,8 +93,8 @@ def test_ever_fully_connected(self):
         now_list = self.h.get_all_fully_connected_at_date(at_date='now')
         ever_list = self.h.get_all_fully_connected_ever()
 
-        self.assertEqual(len(now_list), 12)
-        self.assertEqual(len(ever_list), 163)
+        self.assertEqual(len(now_list), 1)
+        self.assertEqual(len(ever_list), 1)
```
bhazelton: This worries me -- it looks like there is no longer an ever_connected one that isn't currently connected. I think that's why the test coverage dropped a little bit, because it may not be getting to the part of the code that handles past connections...

david-deboer:
I added a defunct path back in.

PR #111, opened 2017-08-22
chk_revisions work for overlapping

david-deboer:
Just added chk_revisions.py file to check for overlapping part revisions.  To my surprise, it found none.

### DIFF
--- a/scripts/chk_revisions.py
+++ b/scripts/chk_revisions.py
@@ -3,9 +3,12 @@
 # Copyright 2017 the HERA Collaboration
 # Licensed under the 2-clause BSD license.
 
-"""This is meant to hold scripts to check various parts/revisions
-
 """
+This checks either a supplied part list or all of the parts
+to find whether it has any overlapping revisions in time.  It
+only prints a warning and the info.
+"""
+
 from __future__ import absolute_import, division, print_function
 from hera_mc import mc, cm_revisions, cm_handling, cm_utils
 import sys
--- a/scripts/chk_revisions.py
+++ b/scripts/chk_revisions.py
@@ -34,17 +34,12 @@ if __name__ == '__main__':
     else:
         args.part = [args.part]
 
-    overlapping = {}
     found_some = False
     for p in args.part:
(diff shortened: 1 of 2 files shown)


scripts/chk_revisions.py
```diff
@@ -0,0 +1,50 @@
+#! /usr/bin/env python
+# -*- mode: python; coding: utf-8 -*-
+# Copyright 2016 the HERA Collaboration
```
pkgw: The year is 2017.

scripts/chk_revisions.py
```diff
@@ -0,0 +1,50 @@
...
+# Licensed under the 2-clause BSD license.
+
+"""This is meant to hold scripts to check various parts/revisions
```
pkgw: This docstring doesn't make sense to me.

scripts/chk_revisions.py
```diff
@@ -0,0 +1,50 @@
...
+    found_some = False
+    for p in args.part:
+        overlap = cm_revisions.check_part_for_overlapping_revisions(p, session)
+        if len(overlap) > 0:
+            overlapping[p] = overlap
+            found_some = True
+    if found_some:
+        print("Overlapping part revisions are:")
+        for k, p in overlapping.iteritems():
+            if len(p) > 0:
```
pkgw: Lengths are never negative, so this test can just be `if len(p):`

scripts/chk_revisions.py
```diff
@@ -0,0 +1,50 @@
...
+                print('\t' + p)
```
pkgw: Here, `p` is whatever value will be returned by `check_part_for_overlapping_revisions`. Is that value something that can be safely `+`'ed with a string? Is it correct that there's no need to print out the `k` variable for context?

pkgw: Reading the code ... it looks like `check_part_for_overlapping_revsisions` will print out any overlap information all by itself, so it's not clear to me that this script actually needs to do any printing at all. If it does, though, the `p` object will be a list of tuples of revision Namespace obejcts, which needs more care to properly stringify.

PR #112, opened 2017-08-23
Cm dataview

david-deboer:


hera_mc/cm_dataview.py
```diff
@@ -0,0 +1,101 @@
...
+
+from __future__ import absolute_import, print_function
+from hera_mc import mc, cm_utils, cm_revisions, cm_hookup
+from astropy.time import Time, TimeDelta
+import numpy as np
+import matplotlib.pyplot as plt
+
+
+def write_file(filename, parts_list, fc_map, output='flag'):
+    from hera_mc import cm_hookup
```
bhazelton: This looks like a duplicate import -- cm_hookup is already imported on line 12

hera_mc/cm_dataview.py
```diff
@@ -0,0 +1,101 @@
...
+    hu = cm_hookup.Hookup(session)
```
bhazelton: It doesn't look like 'session' will be defined here, and it's not passed in...
edit: I see, it's a global variable you define in a later function. I'm not sure I understand the value of the global variable here, rather than just passing it in, I think generally avoiding globals is a good thing. But if there's a really good reason to use a global, it really should be defined earlier in the file than it's used for readability.

In general, I think that functions should occur in a file in an order such that you can always look earlier in the file to find something that is being called by a later function.

hera_mc/cm_dataview.py
```diff
@@ -0,0 +1,101 @@
...
+        while at_date < stop_date:
+            fc = cm_revisions.get_full_revision(p, at_date, full_req, session)
+            fc_map[p]['datetime'].append(at_date.datetime)
+            fc_map[p]['flag'].append(len(fc))
+            fc_map[p]['fc'].append(fc)
+            at_date += TimeDelta(dt * 24 * 3600, format='sec')
+    return fc_map
+
+
+def plot_data(parts, fc_map, separate=True):
```
bhazelton: It looks like there's no test coverage of this function.

hera_mc/cm_dataview.py
```diff
@@ -0,0 +1,101 @@
...
+    fp.close()
+
+
+def read_files(filenames):
+    fc_map = {}
+    parts = []
+    for fn in filenames:
+        try:
+            fp = open(fn, 'r')
+        except IOError:
```
bhazelton: This except statement doesn't have test coverage

hera_mc/cm_utils.py
```diff
@@ -80,6 +80,14 @@ def _split_connection_key(key):
     return ks[0], ks[1], ks[2], ks[3]
 
 
+def listify(X):
```
bhazelton: It looks like there's no test coverage of this function.

hera_mc/cm_dataview.py
david-deboer: Thanks - fixed it.

hera_mc/cm_dataview.py
david-deboer: I fixed the goofy global

hera_mc/cm_dataview.py
david-deboer: Added

hera_mc/cm_dataview.py
david-deboer: How do you test a graphing function?

hera_mc/cm_utils.py
david-deboer: Added

PR #117, opened 2017-08-24
Script update

plaplant:
This changes the way that cm_version information is retrieved. Script works on-site


bhazelton:
This should not be merged before PR #115

PR #121, opened 2017-08-28
Dataview

david-deboer:
This is the first "phase" of harmonizing the cm data handling/viewing scripts etc

### DIFF
--- a/hera_mc/cm_dataview.py
+++ b/hera_mc/cm_dataview.py
@@ -159,6 +159,9 @@ class Dataview:
         ------------
         separate:  T/F to plot the rows separated.
         """
+        if self.fc_map is None or self.parts_list is None:
+            print("You first need to generate fc_map and parts_list")
+            return False
 
         dy = 1.0 / len(self.parts_list)
         for i, p in enumerate(self.parts_list):
@@ -169,3 +172,4 @@ class Dataview:
         plt.legend(bbox_to_anchor=(1.04, 1), loc='upper left')
         plt.grid()
         plt.show()
+        return True
--- a/hera_mc/cm_dataview.py
+++ b/hera_mc/cm_dataview.py
@@ -85,8 +85,7 @@ class Dataview:
         """
 
         if self.fc_map is None or self.parts_list is None:
-            print("You first need to generate fc_map and parts_list")
-            return None
+            raise RuntimeWarning("You first need to generate fc_map and parts_list")
         from hera_mc import cm_hookup
         hu = cm_hookup.Hookup(self.session)
         p0 = self.parts_list[0]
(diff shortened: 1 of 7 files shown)


hera_mc/cm_dataview.py
```diff
@@ -15,85 +15,161 @@
...
+        (i.e. 0 or 1) for the plotter, or the 'corr'elator hookups
+
+        Parameters:
+        ------------
+        filename:  name of output file
+        output:  type of file 'flag' or 'corr'
+        """
+
+        if self.fc_map is None or self.parts_list is None:
+            print("You first need to generate fc_map and parts_list")
```
pkgw: This should raise an exception.

hera_mc/cm_dataview.py
```diff
@@ -15,85 +15,161 @@
...
+    def plot_fc_map(self):
+        """
+        Plots the fc_map flgs.
+
+        Parameters:
+        ------------
+        separate:  T/F to plot the rows separated.
+        """
+        if self.fc_map is None or self.parts_list is None:
+            print("You first need to generate fc_map and parts_list")
```
pkgw: This should raise an exception.

hera_mc/cm_dataview.py
```diff
@@ -15,85 +15,161 @@
...
+        self.fc_map = fc_map
+        return parts, fc_map
+
+    def plot_fc_map(self):
+        """
+        Plots the fc_map flgs.
+
+        Parameters:
+        ------------
+        separate:  T/F to plot the rows separated.
```
pkgw: There is no `separate` parameter anymore.

scripts/geo.py
```diff
@@ -57,31 +56,60 @@
...
-                geo.py corr --loc(-l) --graph(-g)  will provide/plot correlator info for --loc
-
             since: Provides a list of all antennas installed --since DATE
                  geo.py since --date DATE             will provide a list of all antennas built --since DATE
                  geo.py since --date DATE --graph(-g)  will provide a list/plot those antennas
 
             info:  Print this information and exit.
                 geo.py info
+
+        usage: geo.py [-h]
```
pkgw: Is this just a giant copy-paste of the help text producted by the `argparse` module? Surely there's a better way to emit that information.

david-deboer: Well, it is a slightly edited one for clarity -- I wanted `info` to be one-shop stop.

hera_mc/cm_dataview.py
```diff
@@ -85,8 +85,7 @@ def write_fc_map_file(self, filename, output='flag'):
         """
 
         if self.fc_map is None or self.parts_list is None:
-            print("You first need to generate fc_map and parts_list")
-            return None
+            raise RuntimeWarning("You first need to generate fc_map and parts_list")
```
pkgw: If it's being raised, it should be a `RuntimeError`.

hera_mc/cm_dataview.py
```diff
@@ -154,14 +153,9 @@ def read_fc_map_files(self, filenames):
...
         Plots the fc_map flgs.
-
-        Parameters:
-        ------------
-        separate:  T/F to plot the rows separated.
         """
         if self.fc_map is None or self.parts_list is None:
-            print("You first need to generate fc_map and parts_list")
-            return False
+            raise RuntimeWarning("You first need to generate fc_map and parts_list")
```
pkgw: Ditto.

scripts/geo.py
pkgw: It would be nice if this help text could leverage the argparser's knowledge more, since I predict that they will get out of sync in hard-to-track ways. But that's a bridge we can cross later.

PR #126, opened 2017-08-31
Fix parts

david-deboer:
This was to do some fixes on hookup display which ended up being a fairly major rewrite of sections, which also impact the database.  They will need it working on-site.


pkgw:
Looks like this needs rebasing on `master` ... maybe that will fix the test failures?

PR #129, opened 2017-09-03
a script to add observations. works on dcj laptop

dannyjacobs:


scripts/mc_add_observation.py
```diff
@@ -0,0 +1,28 @@
+import argparse,pyuvdata,aipy
```
pkgw: Can you add a hashbang line? `#! /usr/bin/env python`

Also copyright and license notices.

PR #132, opened 2017-09-04
fixed add-part

david-deboer:
Need it to update cm db

### DIFF
--- a/scripts/add_part.py
+++ b/scripts/add_part.py
@@ -54,11 +54,11 @@ if __name__ == '__main__':
     connect = part_connect.Connections()
     part = part_connect.Parts()
     handling = cm_handling.Handling(session)
-    part_check = handling.get_part_dossier(hpn=args.hpn, rev=args.rev,
+    part_check = handling.get_part_dossier(hpn_list=[args.hpn], rev=args.rev,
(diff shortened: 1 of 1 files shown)


scripts/add_part.py
```diff
@@ -54,11 +54,11 @@ def query_args(args):
     connect = part_connect.Connections()
     part = part_connect.Parts()
     handling = cm_handling.Handling(session)
-    part_check = handling.get_part_dossier(hpn=args.hpn, rev=args.rev,
+    part_check = handling.get_part_dossier(hpn_list=[args.hpn], rev=args.rev,
                                            at_date=at_date, exact_match=True)
 
     # Check for part
-    if len(part_check.keys() > 0):
+    if len(part_check.keys()) > 0:
```
pkgw: Stylistically, lengths can never be negative so I like writing `if len(part_check.keys()):` for this sort of thing.

pkgw:
BTW, there has been a whole series of piecemeal updates to function calls like `get_part_dossier` to change argument types. When making interface changes it's very powerful to use `git grep` to make sure that you update everything all at once. For instance, `git grep` very quickly shows me that it looks like `scripts/add_part.py` has not been update for the new `get_part_dossier` call signature.

PR #133, opened 2017-09-06
fixed (most) of the hookup squirreliness

david-deboer:
removes the double entries (for most, but at least the remaining aren't confusing)

### DIFF
--- a/hera_mc/cm_hookup.py
+++ b/hera_mc/cm_hookup.py
@@ -39,7 +39,7 @@ class Hookup:
         self.handling = cm_handling.Handling(session)
         self.part_type_cache = {}
 
-    def get_hookup(self, hpn_list, rev, port, at_date, exact_match=False,
+    def get_hookup(self, hpn_list, rev, port_query, at_date, exact_match=False,
                    show_levels=False, levels_testing=False):
         """
         Return the full hookup to the supplied part/rev/port in the form of a dictionary
@@ -74,13 +74,12 @@ class Hookup:
                                                exact_match=exact_match,
                                                full_version=False)
         hookup_dict = {'hookup': {}, 'fully_connected': {}, 'parts_epoch': []}
-        pols_to_do = self.__get_pols_to_do(hpn_list[0], port, check_pol=False)
-
         part_types_found = []
         for k, part in parts.iteritems():
             if not cm_utils._is_active(self.at_date, part['part'].start_date, part['part'].stop_date):
                 continue
             hookup_dict['hookup'][k] = {}
+            pols_to_do = self.__get_pols_to_do(part, port_query)
             for pol in pols_to_do:
                 hookup_dict['hookup'][k][pol] = self.__follow_hookup_stream(part['part'].hpn, part['part'].hpn_rev, pol)
                 part_types_found = self.__get_part_types_found(hookup_dict['hookup'][k][pol], part_types_found)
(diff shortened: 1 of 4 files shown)


hera_mc/cm_hookup.py
```diff
@@ -106,21 +105,48 @@ def __get_part_types_found(self, huco, part_types_found):
...
+        this figures out which pols to do.  Basically, given 'all' and part it
+        figures out whether to return ['e'], ['n'], ['e', 'n']
+
+        Parameter:
+        -----------
+        part:  current part dossier
+        port_query:  the ports that were requested.
+        """
+        single_pol_parts_paper = ['RI', 'RO', 'CR']
+        pq = port_query.lower()
```
pkgw: Stylistically, I find it neater to modify the variable in question rather than adding another variable with a similar name: `port_query = port_query.lower()`.

hera_mc/cm_hookup.py
```diff
@@ -161,31 +187,35 @@ def __get_next_connection(self, direction, part, rev, pol):
...
+                    cpo = conn.upstream_output_port
+                    npa = conn.downstream_part
+                    npo = conn.downstream_input_port
         next_one = None
         if len(options) == 0:
             next_one = None
         elif len(options) == 1:
-            next_one = options[0]
+            opc = options[0]
+            if self._check_next_port(opc.upstream_part, opc.upstream_output_port, pol, len(options)):
```
pkgw: Does this branch need to check the value of the `direction` variable, like some other spots in this function?

hera_mc/cm_hookup.py
```diff
@@ -106,21 +105,48 @@ def __get_part_types_found(self, huco, part_types_found):
...
             else:
-                pols = False
+                pols = PC.both_pols
+        elif pq in PC.both_pols:
+            pols = [ports]
+        else:
+            pols = PC.both_pols
         return pols
 
+    def _check_next_port(self, next_part, option_port, pol, lenopt):
```
pkgw: Can you add a docstring explaining what is being "checked" here? It's a bit hard for me to figure out.

hera_mc/cm_hookup.py
```diff
@@ -39,7 +39,7 @@ def __init__(self, session=None):
         self.handling = cm_handling.Handling(session)
         self.part_type_cache = {}
 
-    def get_hookup(self, hpn_list, rev, port, at_date, exact_match=False,
+    def get_hookup(self, hpn_list, rev, port_query, at_date, exact_match=False,
                    show_levels=False, levels_testing=False):
```
pkgw: The docstring should be updated to track the changed argument name.

hera_mc/cm_hookup.py
```diff
@@ -106,21 +105,48 @@ def __get_part_types_found(self, huco, part_types_found):
...
+        single_pol_parts_paper = ['RI', 'RO', 'CR']
+        pq = port_query.lower()
+        if pq == 'all':  # Need to figure out if return 'e', 'n' or both
+            if part['part'].hpn[:2].upper() in single_pol_parts_paper:
+                pols = [part['part'].hpn[-1].lower()]
             else:
-                pols = False
+                pols = PC.both_pols
+        elif pq in PC.both_pols:
+            pols = [ports]
```
pkgw: The variable `ports` is not defined.

hera_mc/cm_hookup.py
```diff
@@ -106,21 +105,48 @@ def __get_part_types_found(self, huco, part_types_found):
...
+        else:
+            pols = PC.both_pols
```
pkgw: I'm not sure if this is the best default behavior. The `port_query` passed to this function is either a hardcoded `all` value, or the value of the `--port` argument to `parts.py`. So if I run `parts.py --port=blooga` the program will basically ignore my `--port` argument altogether. I think it's better to be a bit stricter and raise an error if the user provides a value that we don't recognize.

hera_mc/cm_hookup.py
david-deboer: Thanks!

PR #134, opened 2017-09-06
added function and script to add part comments to part_info

david-deboer:
Now that we have failing parts I need to add to the part_info table.

### DIFF
--- a/hera_mc/cm_handling.py
+++ b/hera_mc/cm_handling.py
@@ -155,7 +155,7 @@ class Handling:
                     pr_key = cm_utils._make_part_key(part.hpn, part.hpn_rev)
                     part_dossier[pr_key] = {'Time': at_date, 'part': part}
                     if full_version:
-                        part_dossier[pr_key]['part_info'] = None
+                        part_dossier[pr_key]['part_info'] = []
                         part_dossier[pr_key]['connections'] = None
                         part_dossier[pr_key]['geo'] = None
                         part_dossier[pr_key]['input_ports'] = []
@@ -164,7 +164,7 @@ class Handling:
                                 (func.upper(PC.PartInfo.hpn) == part.hpn.upper()) &
                                 (func.upper(PC.PartInfo.hpn_rev) == part.hpn_rev.upper())):
                             # part_info.gps2Time()
-                            part_dossier[pr_key]['part_info'] = part_info
+                            part_dossier[pr_key]['part_info'].append(part_info)
                         connections = self.get_connection_dossier(
                             hpn_list=[part.hpn], rev=part.hpn_rev, port='all',
                             at_date=at_date, exact_match=True)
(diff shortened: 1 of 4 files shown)


hera_mc/cm_handling.py
```diff
@@ -251,8 +251,13 @@ def show_parts(self, part_dossier, verbosity='h'):
...
                     tdata.append(ptsout.strip().strip(','))
-                    if (part_dossier[hpnr]['part_info'] is not None):
-                        comment = part_dossier[hpnr]['part_info'].comment
+                    if len(part_dossier[hpnr]['part_info']):
+                        comment = ''
+                        for pi in part_dossier[hpnr]['part_info']:
+                            if not len(comment):
+                                comment = pi.comment
+                            else:
+                                comment = comment + ', ' + pi.comment
```
pkgw: This can all be simplified to `comment = ', '.join(part_dossier[hpnr]['part_info'])`.

scripts/part_info.py
```diff
@@ -0,0 +1,51 @@
...
+    parser = mc.get_mc_argument_parser()
+    parser.add_argument('-p', '--hpn', help="HERA part number", default=None)
+    parser.add_argument('-r', '--rev', help="Revision of part", default=None)
+    parser.add_argument('-c', '--comment', help="Comment on part", default=None)
+    parser.add_argument('-l', '--library_file', help="Library filename", default=None)
+    cm_utils.add_date_time_args(parser)
+    args = parser.parse_args()
+
+    if args.hpn is None or args.rev is None or args.comment is None:
+        args = query_args(args)
```
pkgw: Given how `query_args` is structured, it can just be called unconditionally.

scripts/part_info.py
```diff
@@ -0,0 +1,51 @@
...
+
+    # Pre-process some args
+    at_date = cm_utils._get_astropytime(args.date, args.time)
+
+    db = mc.connect_to_mc_db(args)
+    session = db.sessionmaker()
+
+    # Check for part
+    print("Adding comment for part {}:{}".format(args.hpn, args.rev))
```
pkgw: This should be phrased more generically as "Adding info for part...", no? One might want to use this command to register a `library_file`.

hera_mc/cm_handling.py
david-deboer: Actually, I only want the .comment field from it so I think I have to loop.

hera_mc/cm_handling.py
```diff
@@ -252,12 +252,10 @@ def show_parts(self, part_dossier, verbosity='h'):
...
                     if len(part_dossier[hpnr]['part_info']):
-                        comment = ''
+                        comment_list = []
                         for pi in part_dossier[hpnr]['part_info']:
-                            if not len(comment):
-                                comment = pi.comment
-                            else:
-                                comment = comment + ', ' + pi.comment
+                                comment_list.append(pi.comment)
+                        comment = ', '.join(comment_list)
```
pkgw: Ah, right. Could be further tightened up with a list comprehension:

```python
comment = ', '.join(pi.comment for pi in part_dossier[hpnr]['part_info'])
```

PR #136, opened 2017-09-07
Fix levels

david-deboer:
this fixes the hookup output showing correlator levels.


hera_mc/correlator_levels.py
```diff
@@ -65,15 +67,16 @@ def __read_levels_file(name):
...
-def __get_current_levels_from_url(name):
+def __get_current_levels_from_url(name, timeout=5):
     try:
-        if in_container:
-            url = urllib2.urlopen('http://paper1.paper.pvt:3000/instruments/psa256/levels.txt')
+        if paper_network:
+            url = urllib2.urlopen('http://10.0.1.1:3000/instruments/psa256/levels.txt', timeout=timeout)
         else:
-            url = urllib2.urlopen('http://paper1.karoo.kat.ac.za:3000/')
+            url = urllib2.urlopen('http://10.0.1.1:3000/instruments/psa256/levels.txt', timeout=timeout)
```
pkgw: After this PR the two branches of the `if` statement do exactly the same thing! Is this an accident? If not should the statement just be deleted?

pkgw:
Travis is happy again. Good to merge once the if-statement is clarified.

hera_mc/correlator_levels.py
david-deboer: That was left because when/if we get an external path it is there. But u r right, it should just give a warning. Also right now the value is hard-coded in the file...  I'll fix tonight (I'm off today).

PR #138, opened 2017-09-12
added the show-ports and show-revs option for hookup

david-deboer:
default is to not show

### DIFF
--- a/hera_mc/cm_hookup.py
+++ b/hera_mc/cm_hookup.py
@@ -363,7 +363,7 @@ class Hookup:
         else:
             return num_fully_connected > 0
 
-    def show_hookup(self, hookup_dict, cols_to_show='all', show_levels=False):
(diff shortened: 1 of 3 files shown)


hera_mc/cm_hookup.py
```diff
@@ -398,24 +399,41 @@ def __make_header_row(self, header_col_list, cols_to_show):
...
+            elif part_type in headers:
+                if show_port:
+                    td[headers.index(part_type, j)] = dip
+                else:
+                    td[headers.index(part_type, j)] = ''
+                td[headers.index(part_type, j)] += d.upstream_part
+                if show_rev:
+                    td[headers.index(part_type, j)] += ':' + d.up_part_rev
+                if show_port:
+                    td[headers.index(part_type, j)] += ' <' + d.upstream_output_port
```
pkgw: It would be a bit more readable, I think, to construct this string as a local variable, then assign it to `td[headers.index(part_type, j)]` after all of the conditionals have been worked through.

PR #139, opened 2017-09-14
had to fix up part_info to make it scriptable

david-deboer:


scripts/part_info.py
```diff
@@ -34,17 +35,29 @@ def query_args(args):
     parser.add_argument('-r', '--rev', help="Revision of part", default=None)
     parser.add_argument('-c', '--comment', help="Comment on part", default=None)
     parser.add_argument('-l', '--library_file', help="Library filename", default=None)
+    parser.add_argument('--swap-underscores', dest='swap_underscores', help="Replaces underscores in comment with spaces.", action='store_true')
```
pkgw: Is this argument really necessary? One should be able to use shell quoting to get spaces into the argument value. If there's a place where that's presenting a challenge, I think the solution is to fix the shell stuff rather than add bells and whistles to the program.

david-deboer: Within a script file for some reason it does not recognize quotes.   If I type it at the command line it works fine.  So, I took the stupid way out just to move on.

pkgw: Where's the script?

david-deboer: it is hera_cm_db_updates/scripts/170913_fem_swap_HH67

david-deboer: I did take out the quotation marks in that one and put the underscores in.

pkgw: How are you running the scripts? It should definitely work to just write

```
part_info.py [...] -c "Was on HH67 but failed."
```

But the scripts don't have a hashbang line (`#! /bin/bash`) so they can't be run by the shell in the standard way. If you're running them in some unusual way that might cause problems.

david-deboer: Still doesn't work.  Just try
#! /bin/bash
part_info.py --date 2017/09/22 --time 8.0 -p FEM75026 -r A -c “Testing 123 456” -l none

Actually, right now it doesn't work on the command line for me either...

david-deboer: Actually, it is because you need to use single quotes '...

david-deboer: Although it still doesn't work within a script file.

david-deboer: I've tried \', \", etc and none will work if run from within a script file...

PR #152, opened 2017-10-08
Fix maindb

david-deboer:
This changes how I handle maindb in a much simpler way.


david-deboer:
I did a pull test to check tests online (see my Slack note on nose tests)

PR #161, opened 2017-10-19
added html stuff for sys_handling

david-deboer:
but also a few other goodies

### DIFF
--- a/hera_mc/cm_hookup.py
+++ b/hera_mc/cm_hookup.py
@@ -1,6 +1,6 @@
 #! /usr/bin/env python
 # -*- mode: python; coding: utf-8 -*-
-# Copyright 2016 the HERA Collaboration
+# Copyright 2017 the HERA Collaboration
 # Licensed under the 2-clause BSD license.
 
 """
@@ -285,24 +285,23 @@ class Hookup:
         hookup_dict['columns'].append('level')
         hookup_dict['levels'] = {}
         pf_input = []
-        for k in sorted(hookup_dict['hookup'].keys()):
-            for p in hookup_dict['hookup'][k].keys():
-                try:
+        sorted_hkeys = sorted(hookup_dict['hookup'].keys())
+        # *** These nested loops must be in the same order as below ***
+        for k in sorted_hkeys:
+            for p in sorted(hookup_dict['hookup'][k].keys()):
+                if len(hookup_dict['hookup'][k][p]):
                     f_engine = hookup_dict['hookup'][k][p][-1].downstream_part
-                except IndexError:
-                    continue
+                else:
+                    f_engine = None
                 pf_input.append(f_engine)
         levels = correlator_levels.get_levels(pf_input, dummy_file)
+        # *** These nested loops must be in the same order as above ***
         level_ctr = 0
-        for k in sorted(hookup_dict['hookup'].keys()):
+        for k in sorted_hkeys:
             hookup_dict['levels'][k] = {}
-            for p in hookup_dict['hookup'][k].keys():
-                try:
-                    lstr = '%s' % (levels[level_ctr])
-                except IndexError:
-                    lstr = 'x'
+            for p in sorted(hookup_dict['hookup'][k].keys()):
+                hookup_dict['levels'][k][p] = str(levels[level_ctr])
                 level_ctr += 1
-                hookup_dict['levels'][k][p] = lstr
         return hookup_dict
 
     def get_correlator_input_from_hookup(self, hookup_dict):
@@ -370,7 +369,7 @@ class Hookup:
             return num_fully_connected > 0
 
     def show_hookup(self, hookup_dict, cols_to_show='all', show_levels=False, show_ports=True, show_revs=True,
-                    write_to_file=False):
+                    file=None, output_format='ascii'):
         """
         Print out the hookup table -- uses tabulate package.
 
@@ -381,7 +380,7 @@ class Hookup:
         show_levels:  boolean to either show the correlator levels or not
         show_ports:  boolean to include ports or not
         show_revs:  boolean to include revisions letter or not
-        wrie_to_file:  write to the supplied filename instead of screen
+        file:  file to use, None goes to stdout
         """
         headers = self.__make_header_row(hookup_dict['columns'], cols_to_show)
         table_data = []
@@ -398,15 +397,13 @@ class Hookup:
                                                headers, timing, level,
                                                show_ports, show_revs)
                     table_data.append(td)
-        if type(write_to_file) == str:
-            fp = open(write_to_file, 'w')
-            fp.write('<html>\n<body>\n<pre>\n')
-            fp.write(tabulate(table_data, headers=headers, tablefmt='orgtbl'))
-            fp.close()
-        else:
-            print('\n')
-            print(tabulate(table_data, headers=headers, tablefmt='orgtbl'))
-            print('\n')
+        table = tabulate(table_data, headers=headers, tablefmt='orgtbl') + '\n'
+        if file is None:
+            import sys
+            file = sys.stdout
+        if output_format == 'html':
+            table = '<html>\n\t<body>\n\t\t<pre>\n' + table + '\t\t</pre>\n\t</body>\n</html>\n'
+        print(table, file=file)
 
     def __make_header_row(self, header_col_list, cols_to_show):
         headers = []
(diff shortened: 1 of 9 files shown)


hera_mc/cm_hookup.py
```diff
@@ -287,14 +287,20 @@ def __hookup_add_correlator_levels(self, hookup_dict):
         pf_input = []
         for k in sorted(hookup_dict['hookup'].keys()):
             for p in hookup_dict['hookup'][k].keys():
-                f_engine = hookup_dict['hookup'][k][p][-1].downstream_part
+                try:
+                    f_engine = hookup_dict['hookup'][k][p][-1].downstream_part
+                except IndexError:
+                    continue
```
pkgw: Some people like it, but I personally am not a big fan of this approach to bounds-checking. In this case in particular, it is not obvious which indexing operation is the one that can fail. Can you rewrite using an `if` statement with a `break` if a usable record is found?

hera_mc/cm_hookup.py
```diff
@@ -287,14 +287,20 @@ def __hookup_add_correlator_levels(self, hookup_dict):
...
         levels = correlator_levels.get_levels(pf_input, dummy_file)
         level_ctr = 0
         for k in sorted(hookup_dict['hookup'].keys()):
             hookup_dict['levels'][k] = {}
             for p in hookup_dict['hookup'][k].keys():
-                lstr = '%s' % (levels[level_ctr])
+                try:
+                    lstr = '%s' % (levels[level_ctr])
+                except IndexError:
+                    lstr = 'x'
```
pkgw: Same goes here. Also, `'%s' % x` can be expressed more simply as `str(x)`.

hera_mc/cm_hookup.py
```diff
@@ -363,7 +369,8 @@ def is_fully_connected(self, hookup_dict, any_or_all='all'):
         else:
             return num_fully_connected > 0
 
-    def show_hookup(self, hookup_dict, cols_to_show='all', show_levels=False, show_ports=True, show_revs=True):
+    def show_hookup(self, hookup_dict, cols_to_show='all', show_levels=False, show_ports=True, show_revs=True,
+                    write_to_file=False):
```
pkgw: Instead of a filename flag, the new argument should be called `file` and optionally receive a file-like object (e.g. the return value from the `open()` function) as an argument. Its default value should be `None`, in which case this function should override the argument's value with `sys.stdout`. Then, the `print` calls should use the `file` keyword argument to specify where the printed text should go:

```
print('foo', file=file)
```

Finally, there should be a separate `output_format` keyword that defaults to `"ascii"` but can take on the value `"html"`, in which case the outputted text should have the wrapping HTML tags added.

hera_mc/cm_hookup.py
```diff
@@ -372,10 +379,14 @@ def show_hookup(self, hookup_dict, cols_to_show='all', show_levels=False, show_p
         hookup_dict:  generated in self.get_hookup
         cols_to_show:  list of columns to include in hookup listing
         show_levels:  boolean to either show the correlator levels or not
+        show_ports:  boolean to include ports or not
+        show_revs:  boolean to include revisions letter or not
+        wrie_to_file:  write to the supplied filename instead of screen
```
pkgw: Typo here (although this should be superseded).

hera_mc/cm_hookup.py
```diff
@@ -387,9 +398,15 @@ def show_hookup(self, hookup_dict, cols_to_show='all', show_levels=False, show_p
...
                                                show_ports, show_revs)
                     table_data.append(td)
-        print('\n')
-        print(tabulate(table_data, headers=headers, tablefmt='orgtbl'))
-        print('\n')
+        if type(write_to_file) == str:
+            fp = open(write_to_file, 'w')
+            fp.write('<html>\n<body>\n<pre>\n')
+            fp.write(tabulate(table_data, headers=headers, tablefmt='orgtbl'))
+            fp.close()
```
pkgw: When the output format is HTML, it should be valid HTML. There should be an empty `<head></head>` section and closing tags at the end of the file (`</pre></body></html>`).

hera_mc/cm_utils.py
```diff
@@ -157,6 +159,36 @@ def _get_astropytime(_date, _time=0):
...
+        for i in range(len(k)):
+            try:
+                n = int(k[i:colon])
+                break
+            except ValueError:
+                continue
+        if n in keylib.keys():
+            dup_key = keylib[n][0] + str(n) + keylib[n][1]
+            print('Duplicated number found while trying to put keys in numerical order ({}, {})'.format(k, dup_key))
+            print('Returning original order')
```
pkgw: I would suppress these messages. It is very hard to write programs that have compact, easily-understood output if random bits of code deep in the software stack occasionally print miscellaneous diagnostic messages.

hera_mc/cm_utils.py
```diff
@@ -157,6 +159,36 @@ def _get_astropytime(_date, _time=0):
...
+            return keys
+        keylib[n] = [k[:i], k[colon:]]
+    if not len(keylib.keys()):
+        return keys
+    sortkn = sorted(keylib.keys())
+    keyordered = []
+    for n in sortkn:
```
pkgw: You could avoid the temporary variable here and just write `for n in sorted(keylib.keys()):`.

Even more compact would be replacing this whole final stanza with `return [keylib[n][0] + str(n) + keylib[n][1] for n in sorted(keylib.keys())` but it would be reasonable to find that a bit too terse.

hera_mc/sys_handling.py
```diff
@@ -369,3 +365,20 @@ def get_pam_info(self, stn, at_date, fc=None):
         H = cm_hookup.Hookup(self.session)
         pams = H.get_pam_from_hookup(hu)
         return pams
+
+    def publish_summary(self, hlist=['HH'], rev='A', exact_match=False,
+                        hookup_cols=['station', 'front-end', 'cable-post-amp(in)', 'post-amp', 'cable-container', 'f-engine', 'level']):
+        file2write = 'sys_conn.html'
```
pkgw: Personal pet peeve: `2` standing for `to`. Please use `file_to_write`.

hera_mc/sys_handling.py
```diff
@@ -369,3 +365,20 @@ def get_pam_info(self, stn, at_date, fc=None):
...
+        location_on_paper1 = 'paper1:/home/davidm/local/src/rails-paper/public'
+        hookup_dict = self.hookup.get_hookup(hpn_list=hlist, rev=rev, port_query='all',
+                                             at_date='now', exact_match=exact_match, show_levels=True)
+        self.hookup.show_hookup(hookup_dict, hookup_cols, True, False, False, write_to_file=file2write)
```
pkgw: Once the arguments to `show_hookup` are fixed up, this chunk of code can become something like:

```
with open(file_to_write, 'w') as f:
  self.hookup.show_hookup(..., file=f, format='html')
```

The `with` statement makes sure that the file is closed, and gives precise control over when exactly that happens.

scripts/online_roach_powers.py
```diff
@@ -0,0 +1,35 @@
+#! /usr/bin/env python
+# -*- mode: python; coding: utf-8 -*-
+# Copyright 2016 the HERA Collaboration
```
pkgw: The year is 2017.

scripts/online_roach_powers.py
```diff
@@ -0,0 +1,35 @@
...
+# Licensed under the 2-clause BSD license.
+
+"""This is meant to hold utility scripts for handling parts and connections
```
pkgw: This docstring does not seem correct.

scripts/online_roach_powers.py
```diff
@@ -0,0 +1,35 @@
+#! /usr/bin/env python
```
pkgw: I'm not sure I get the name of this script. Something with a verb would be better, I think. How about "mc_publish_summary.py"? I also think it's important to namespace the scripts (e.g. use the `mc_` prefix) to maintain some level of organization; our `bin/` directories are rapidly accumulating a whole mess of scripts from a whole bunch of subsystems, which makes it hard to remember names and meanings.

hera_mc/cm_hookup.py
david-deboer: OK - so you are going to make me understand what's going on again eh? ;-)  I've fixed this and below (and found a potential silent error).

scripts/online_roach_powers.py
david-deboer: renamed to mc_publish_sys_levels.py

PR #164, opened 2017-10-23
Change hookup to use a "local" file to speed it up.

david-deboer:
This changes hookup to use a local file if it is time-stamped after the latest cm_version table entry.
You can force it to use a new hookup and you can also force it to act as it did previously.

### DIFF
--- a/hera_mc/cm_hookup.py
+++ b/hera_mc/cm_hookup.py
@@ -69,7 +69,7 @@ class Hookup:
         force_new:  boolean to force a full database read as opposed to checking file
         force_specific:  boolean to force this to read/write the file to use the supplied values
         """
-        if force_specfic:
+        if force_specific:
             force_new = True
         if force_new or not self.check_if_hookup_file_is_current():
             if force_specific:
@@ -78,7 +78,7 @@ class Hookup:
                 p = port_query
             else:
                 h = self.hookup_list_to_cache
-                r = 'LAST'
+                r = 'ACTIVE'
                 p = 'all'
             hookup_dict = self.__get_hookup(hpn_list=h, rev=r, port_query=p,
                                             at_date=at_date, exact_match=exact_match, show_levels=show_levels)
@@ -87,20 +87,20 @@ class Hookup:
             hookup_dict = self.read_hookup_from_file()
             if show_levels:
                 hookup_dict = self.__hookup_add_correlator_levels(hookup_dict)
-
         # Now we need to delete the ones we don't use since the file (probably) has all of them.
-        for k in hookup_dict['hookup'].keys():
-            if exact_match:
-                if k not in hpn_list:
-                    del hookup_dict['hookup'][k]
-            else:
-                del_this_one = True
-                for p in hpn_list:
-                    if k.lower()[:len(p)] == p.lower():
-                        del_this_one = False
-                        break
-                if del_this_one:
-                    del hookup_dict['hookup'][k]
+        if not force_specific:
+            for k in hookup_dict['hookup'].keys():
+                if exact_match:
+                    if k not in hpn_list:
+                        del hookup_dict['hookup'][k]
+                else:
+                    del_this_one = True
+                    for p in hpn_list:
+                        if k.lower()[:len(p)] == p.lower():
+                            del_this_one = False
+                            break
+                    if del_this_one:
+                        del hookup_dict['hookup'][k]
         return hookup_dict
 
     def __get_hookup(self, hpn_list, rev, port_query, at_date,
@@ -118,7 +118,7 @@ class Hookup:
                                                full_version=False)
         print(hpn_list)
         print(parts)
-        hookup_dict = {'hookup': {}, 'fully_connected': {}, 'parts_epoch': []}
+        hookup_dict = {'hookup': {}, 'fully_connected': {}, 'parts_epoch': {}}
         part_types_found = []
         for k, part in parts.iteritems():
             if not cm_utils._is_active(self.at_date, part['part'].start_date, part['part'].stop_date):
--- a/hera_mc/cm_hookup.py
+++ b/hera_mc/cm_hookup.py
@@ -37,7 +37,7 @@ class Hookup:
             self.session = session
         self.handling = cm_handling.Handling(session)
         self.part_type_cache = {}
-        self.hookup_local_file = os.path.join(os.path.expanduser('~'), 'hookup_tmp.npy')
+        self.hookup_local_file = os.path.expanduser('~/.hera_mc_hookup_cache.npy')
         self.hookup_list_to_cache = hookup_list_to_cache
 
     def get_hookup(self, hpn_list, rev, port_query='all', at_date='now',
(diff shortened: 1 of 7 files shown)


hera_mc/cm_hookup.py
```diff
@@ -7,15 +7,14 @@
 This is computes and displays part hookups.
 
 """
-from __future__ import absolute_import, division, print_function
+from __future__ import print_function
 from tabulate import tabulate
```
pkgw: Why were these removed? They enable behavior that is both more intuitive *and* consistent with Python 3.

hera_mc/cm_hookup.py
```diff
@@ -71,7 +116,9 @@ def get_hookup(self, hpn_list, rev, port_query, at_date,
                                                at_date=self.at_date,
                                                exact_match=exact_match,
                                                full_version=False)
-        hookup_dict = {'hookup': {}, 'fully_connected': {}, 'parts_epoch': []}
+        print(hpn_list)
+        print(parts)
```
pkgw: These look like debugging print statements that should not be kept in the final code.

hera_mc/cm_hookup.py
```diff
@@ -368,8 +416,31 @@ def is_fully_connected(self, hookup_dict, any_or_all='all'):
...
+        from astropy.time import Time
+        from hera_mc import cm_transfer
+        if os.path.isfile(self.hookup_local_file):
+            result = self.session.query(cm_transfer.CMVersion).order_by(cm_transfer.CMVersion.update_time).all()
+            cm_hash_time = Time(result[-1].update_time, format='gps')
+            stats = os.stat(self.hookup_local_file)
+            file_mod_time = Time(stats.st_mtime, format='unix')
+            return file_mod_time > cm_hash_time
+        else:
+            return False
```
pkgw: Two things could be improved here. First, the `os.stat` call effectively tests whether the file exists anyway, so the `os.path.isfile` call can be avoided. In fact, doing both calls introduces the possibility of a race condition between them, and it is good to develop a habit of spotting and avoiding race conditions since they can lead to subtle but significant bugs.

Second, the logic can be reordered to reduce the amount of rightward drift in the code.

A refined version would look something like:

```python
try:
  stats = os.stat(self.hookup_local_file)
except OSError as e:
  if e.errno == 2:
    return False # ENOENT: file does not exist
  raise # some other error stat'ing the file that we can't handle

result = self.session.query(...)
...
```

hera_mc/cm_hookup.py
```diff
@@ -412,11 +483,13 @@ def show_hookup(self, hookup_dict, cols_to_show='all', show_levels=False, show_p
             table = '<html>\n\t<body>\n\t\t<pre>\n' + table + '\t\t</pre>\n\t</body>\n</html>\n'
         print(table, file=file)
 
-    def __make_header_row(self, header_col_list, cols_to_show):
+    def __make_header_row(self, header_col_list, cols_to_show, show_levels):
```
pkgw: I am a bit confused by this approach. Why not just add `level` to `cols_to_show` when one wants to achieve the effect of `show_levels = True`?

hera_mc/correlator_levels.py
```diff
@@ -18,7 +18,7 @@
 default_levels_filename = 'levels.tmp'
 
 
-def get_levels(pf_input, testing, network='local', timeout=5):
+def get_levels(pf_input, testing, network='local', timeout=2):
```
pkgw: The change of the timeout here is an example of an unrelated change that should really go in its own commit, so that the commit message can explain the change. Unless it's not unrelated?

(If Future Me wanted to know how this timeout was chosen, I'd use `git blame` and then `git log` to find this commit, and then I'd be stuck since there is presently no explanation for why the change was made, and it is not at all obvious to me what the change has to do with the general thrust of this pull request.)

scripts/mc_publish_sys_levels.py
```diff
@@ -19,6 +19,8 @@
     parser.add_argument('-r', '--revision', help="Specify revision or last/active/full/all for hpn.  [A]", default='A')
     parser.add_argument('-e', '--exact-match', help="Force exact matches on part numbers, not beginning N char. [False]",
                         dest='exact_match', action='store_true')
+    parser.add_argument('-f', '--force-new', dest='force_new', help="Force it to write a new hookup file.", action='store_true')
+    parser.add_argument('--force-specific', dest='force_specific', help="Force db and local cache file to use these values", action='store_true')
```
pkgw: Is one ever going to want to pass these arguments to this particular script? It doesn't seem like the right venue for fooling with what goes in the cache file.

hera_mc/cm_hookup.py
```diff
@@ -368,8 +416,31 @@ def is_fully_connected(self, hookup_dict, any_or_all='all'):
         else:
             return num_fully_connected > 0
 
+    def write_hookup_to_file(self, hookup_dict):
+        np.save(self.hookup_local_file, hookup_dict)
+        np.save(self.part_type_cache_file, self.part_type_cache)
```
pkgw: The two arrays can be written to a single file thusly:

```python
with open(self.hookup_local_file, 'wb') as f:
    np.save(f, hookup_dict)
    np.save(f, self.part_type_cache)
```

And analogous code (changing the "wb" to "rb" to open the file in read mode) will read them back out.

hera_mc/cm_hookup.py
```diff
@@ -38,11 +36,17 @@ def __init__(self, session=None):
             self.session = session
         self.handling = cm_handling.Handling(session)
         self.part_type_cache = {}
+        self.hookup_local_file = 'hookup_tmp.npy'
+        self.part_type_cache_file = 'part_type_cache_tmp.npy'
```
pkgw: I do not like these file names since they are relative, so the cache files will land in whatever directory you happen to be running the code in. This means the module will end up dropping little turds all over your working directories over time. You can use the function `os.path.expanduser` to create absolute paths that point to the user's home directory, which will give one cache file per user. I believe that it is safe to assume that the generic cache directory `~/.cache/` exists, so that's probably a good place to store the files.

See also farther down for a way to just use one file instead of two.

hera_mc/cm_hookup.py
```diff
@@ -62,6 +66,47 @@ def get_hookup(self, hpn_list, rev, port_query, at_date,
         at_date:  date for hookup validity
         exact_match:  boolean for either exact_match or partial
         show_levels:  boolean to include correlator levels
+        force_new:  boolean to force a full database read as opposed to checking file
+        force_specific:  boolean to force this to read/write the file to use the supplied values
```
pkgw: I'm having a bit trouble understanding the purpose of the `force_specific` mode. It writes out a cache file based on the input arguments if I understand things correctly. I imagine that users are unlikely to type in very many arguments manually, so this means that the output file will only contain a few pieces of information, right? Can you add some text to the docstring explaining the intended use case?

david-deboer: I want to keep the earlier functionality, so force_specific makes it act just like the earlier version.

hera_mc/cm_hookup.py
david-deboer: You can always include "levels" in the list, but if you show-levels you don't necessarily need to or want to remember to do that.  Also, we now have a file to deal with that may or may not have the flag set.  It just sort of made sense with no harm.

hera_mc/correlator_levels.py
david-deboer: I was just getting impatient with waiting the 5 seconds...

scripts/mc_publish_sys_levels.py
david-deboer: Maybe no force-specific, but possibly force-new.  This is going into the crontab and I liked the flexibility.

hera_mc/cm_hookup.py
pkgw: So the problem is that, if I'm reading the code correctly, if a person runs the program once with `--force-specific`, then reruns it later without that flag, the second invocation will pick up the cache file made by the first one. This file will contain some arbitrary subset of the full hookup rather than the default, full `HH`hookup. It seems like this could be very confusing if one forgot about the caching or if the two invocations were separated widely in time. Should the cache file only be written out if `force_specific` is false?

hera_mc/cm_hookup.py
```diff
@@ -38,11 +37,16 @@ def __init__(self, session=None):
             self.session = session
         self.handling = cm_handling.Handling(session)
         self.part_type_cache = {}
+        self.hookup_local_file = os.path.join(os.path.expanduser('~'), 'hookup_tmp.npy')
```
pkgw: Do Macs have a `~/.cache/` directory by default? I'd like to store this file as `~/.cache/hera_mc_hookup_cache.npy` or `~/.hera_mc_hookup_cache.npy` for clarity and to keep directory listings clean.

Also, you can jut write `os.path.expanduser('~/.hera_mc_hookup_cache.npy')` and it will work, rather than using the `join`.

PR #168, opened 2017-10-25
added the hookup_dict_library thing

david-deboer:
This just adds a "library" file so that repeated calls don't need to read in the file every time.  Two comments:

(1) arguably the "intelligence" should be in the show, not the dictionary (but that is not how it is...)
(2) do I need the copy.copy s

### DIFF
--- a/hera_mc/cm_hookup.py
+++ b/hera_mc/cm_hookup.py
@@ -39,6 +39,7 @@ class Hookup:
         self.part_type_cache = {}
         self.hookup_local_file = os.path.expanduser('~/.hera_mc_hookup_cache.npy')
         self.hookup_list_to_cache = hookup_list_to_cache
+        self.hookup_dict_library = None
 
     def get_hookup(self, hpn_list, rev, port_query='all', at_date='now',
                    exact_match=False, show_levels=False,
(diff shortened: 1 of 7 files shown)


hera_mc/cm_hookup.py
```diff
@@ -79,10 +80,15 @@ def get_hookup(self, hpn_list, rev, port_query='all', at_date='now',
             hookup_dict = self.__get_hookup(hpn_list=self.hookup_list_to_cache, rev='ACTIVE', port_query='all',
                                             at_date=at_date, exact_match=exact_match, show_levels=show_levels)
             self.write_hookup_to_file(hookup_dict)
+            self.hookup_dict_library = copy.copy(hookup_dict)
```
pkgw: I don't think this `copy` is necessary.

hera_mc/cm_hookup.py
```diff
@@ -79,10 +80,15 @@ def get_hookup(self, hpn_list, rev, port_query='all', at_date='now',
...
-            hookup_dict = self.read_hookup_from_file()
-            if show_levels:
-                hookup_dict = self.__hookup_add_correlator_levels(hookup_dict)
+            if self.hookup_dict_library is None:
+                hookup_dict = self.read_hookup_from_file()
+                if show_levels:
+                    hookup_dict = self.__hookup_add_correlator_levels(hookup_dict)
+                self.hookup_dict_library = copy.copy(hookup_dict)
+            else:
+                hookup_dict = copy.copy(self.hookup_dict_library)
```
pkgw: If the caller of this function is likely to modify the returned `hookup_dict`, then a copy should be used. If it's going to be more or less immutable, then it can be skipped.

hera_mc/cm_hookup.py
```diff
@@ -39,6 +39,7 @@ def __init__(self, session=None, hookup_list_to_cache=['HH']):
         self.part_type_cache = {}
         self.hookup_local_file = os.path.expanduser('~/.hera_mc_hookup_cache.npy')
         self.hookup_list_to_cache = hookup_list_to_cache
+        self.hookup_dict_library = None
```
pkgw: I think it would be better to name this `saved_hookup_dict` or `cached_hookup_dict`.

hera_mc/cm_hookup.py
david-deboer: In fact, the code just below modifies the hookup_dict before it sends it, but we want to keep the full for the next time this gets called.  I would think that all of the copy.copy(ies) would be needed, if any are needed.

david-deboer:
How did this fail?!?  Grrrrrrrrrrrrrrrrrr

PR #169, opened 2017-11-02
Fix hookup

david-deboer:
This is a rather substantive set of changes that change quite a bit of how the hookup stuff is handled.


hera_mc/tests/test_dataview.py
```diff
@@ -19,34 +19,34 @@
...
+    #             os.remove(file)
+    #
+    # def test_read_files(self):
+    #     filename = [os.path.join(mc.test_data_path, 'HH0_15_flag.txt'), 'not_a_real_file']
+    #     parts, fc_map = self.dv.read_fc_map_files(filename)
+    #     self.assertTrue(parts[0] == 'HH0')
+
+
+# if __name__ == '__main__':
+#     unittest.main()
```
pkgw: This could should be simply deleted instead of commented out in bulk. We can recover it from the Git history if we ever want it back.

hera_mc/cm_hookup.py
```diff
@@ -4,52 +4,108 @@
...
+            if pind < 0:
+                pind = 0
+            parts[k][pol] = (p[pind].upstream_part, p[pind].downstream_part)
+    return parts
+
+
 class Hookup:
     """
-    Class to find and display the signal path hookup.
+    Class to find and display the signal path hookup.  It only has three public methods:
```
pkgw: One of the follow-up commits should have updated the word "three" to "four". Going forward, probably best just to write "Its public methods are:".

david-deboer:
Thanks Peter.  I certainly agree that, given where it is currently at, hookup_dict could/should be its own class.  I'll make that a future rewrite.  I think/hope that it is all OK -- it is a bit leaner/cleaner at least.  I've deleted a number of de facto deprecated things (and just did the test_dataview until/if it gets revamped).

PR #170, opened 2017-11-03
Publev

david-deboer:
Makes the small changes to implement the mc_publish_sys_levels thing.


hera_mc/sys_handling.py
```diff
@@ -267,9 +268,10 @@ def publish_summary(self, hlist=['HH'], rev='A', exact_match=False,
         import subprocess
         from hera_mc import cm_transfer
         if cm_transfer.check_if_main():
+            import time
+            time.sleep(2)
```
pkgw: Why is this needed? If it's truly necessary there should at least be a comment explaining why.

david-deboer: Good point - I don't know if it is needed.  It was an attempt to fix the earlier unknown error.  That's how it currently works and I'm uncertain whether I should remove it.

pkgw: I have trouble seeing how it could possibly be necessary, and things like this should be taken out unless they really have a specific motivation.

PR #173, opened 2017-11-15
Weather table

bhazelton:
I need to figure out how to test the part of this that talks to katportalclient.


hera_mc/weather.py
```diff
@@ -0,0 +1,128 @@
...
+from math import floor
+from sqlalchemy import Column, BigInteger, Float, String
+
+import tornado.gen
+
+from . import MCDeclarativeBase
+
+weather_sensor_dict = {'wind_speed': {'sensor_name': 'anc_wind_wind_speed', 'units': 'm/s'},
+                       'wind_direction': {'sensor_name': 'anc_wind_wind_direction', 'units': 'degrees'},
+                       'temperature': {'sensor_name': 'anc_weather_temperature', 'units': 'deg. Celcius'}}
```
pkgw: Spelling mistake (perhaps theirs, though?)

hera_mc/weather.py
```diff
@@ -0,0 +1,128 @@
...
+        variable: string
+            must be a key in weather_sensor_dict
+        value: float
+            value from the sensor associated with the variable
+        """
+        if not isinstance(time, Time):
+            raise ValueError('time must be an astropy Time object')
+        if variable not in weather_sensor_dict.keys():
+            raise ValueError('variable must be a key in weather_sensor_dict.')
+        if not isinstance(value, (float)):
```
pkgw: Superfluous parentheses.

hera_mc/weather.py
```diff
@@ -0,0 +1,128 @@
...
+    for var in variables:
+        if var not in weather_sensor_dict.keys():
+            raise ValueError('variable must be a key in weather_sensor_dict')
+        sensor_names.append(weather_sensor_dict[var]['sensor_name'])
+        sensor_var_dict[weather_sensor_dict[var]['sensor_name']] = var
+
+    portal_client = KATPortalClient('http://portal.mkat.karoo.kat.ac.za/api/client',
+                                    on_update_callback=None)
+
+    sensor_names = yield portal_client.sensor_names(sensor_names)
```
pkgw: So here it seems that we're filtering down our list of sensors to use; the docs to the `sensor_names` method say that the inputs are regular expressions. But we're not using any regexes, so I think at best all this call does is trigger an error if a sensor is unknown. But that will happen later if that's the case. Given how our code works, I think it would be better just to delete this line.

hera_mc/weather.py
```diff
@@ -0,0 +1,128 @@
...
+
+    histories = yield portal_client.sensors_histories(sensor_names, starttime.unix,
+                                                      stoptime.unix, timeout_sec=60)
+
+    weather_obj_list = []
+    for sensor_name, history in histories.items():
+        if len(history) > 0:
```
pkgw: Superfluous if statement. Without it, if `len(history)` is 0, the right thing still happens.

hera_mc/weather.py
```diff
@@ -0,0 +1,128 @@
...
+            for count in range(len(history)):
```
pkgw: The only usage of `count` and `history` inside the loop is the expression `history[count]`, so it would be simply just to write `for item in history:` and use `item` instead of `history[count]`.

hera_mc/weather.py
```diff
@@ -0,0 +1,128 @@
...
+                if 'value_timestamp' in history[count]._fields:
+                    timestamp = history[count].value_timestamp
+                else:
+                    timestamp = history[count].timestamp
```
pkgw: What are the semantics of the `value_timestamp` and `timestamp` field? This is the kind of thing not obvious to casual reader (me) that would be helpful to briefly document in a comment.

hera_mc/weather.py
```diff
@@ -0,0 +1,128 @@
...
+                time_use = Time(timestamp, format='unix')
+                value = float(history[count].value)
+                status = history[count].status
+                if status == 'nominal':
```
pkgw: If we're going to be filtering on this, it seems clearer to do so at the top of the loop:

```
for item in history:
  if item.status != 'nominal':
    continue
  ...
```

pkgw:
Mostly looks good. I added some comments about small-bore items.

One big one is that it doesn't seem great to have a `units` column for every single measurement in the database. Hopefully for a given variable, the units are always the same. If you're being pedantically databasey about things I think the solution would be to add another table of measurement types (columns: id number, name, units) and then have the measurement values table foreign-key into it rather than have the name and units columns. But I'd be just as happy to track this information in a hashtable in the Python file.

The other thing is that I have a mild preference for keeping all of the Tornado stuff isolated in `weather.py`; right now some of it leaks into `mc_session.py`. You'd do that by building up a big list of new weather sensor measurements rather than adding them one-by-one inside the Tornado coroutine. But the current approach should be more efficient.

Also should `Docs/mc_definition.tex` be updated in this PR?

pkgw:
Oh, and I forgot to mention about testing: you can *almost* do this over an SSH tunnel, but inside the guts of `katportalclient` it asks the primary server for a list of other URLs to fetch, so it doesn't work in the end. I think the best testing approach is to create a temporary Conda environment and make sure that one is not connecting to the live M&C database (relatively easy here if one makes sure to only add the new weather table to the testing database).

hera_mc/mc_session.py
```diff
@@ -748,6 +749,77 @@ def get_paper_temps(self, starttime, stoptime=None):
...
+        ------------
+        starttime: astropy time object
+            time to look for records after
+
+        stoptime: astropy time object
+            last time to get records for. If none, only the first record after
+            starttime will be returned.
+
+        variable: string
+            Name of variablen to get records for, must be a key in weather.weather_sensor_dict.
```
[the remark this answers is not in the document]
bhazelton: thanks

hera_mc/weather.py
bhazelton: no, it was mine, embarrassing as it is

hera_mc/weather.py
bhazelton: ok

hera_mc/weather.py
bhazelton: thanks

hera_mc/weather.py
bhazelton: thanks.

hera_mc/weather.py
bhazelton: the value_timestamp is the sensor timestamp, while the other is when the recording system got it. The value_timestamp isn't always present, so I test for it. I'll add a comment.

hera_mc/weather.py
bhazelton: sure, thanks.

bhazelton:
Yeah, I was trying to figure out how to keep the tornado stuff in weather.py, I'll make a stab at changing that. I was having trouble figuring it out last night, but I think I can make it work.

You're right, I should update the docs, I'll do that.

Thanks for looking into the testing situation. I think I can do the testing I need with just the test database and nosetests. The thing that confuses me about the environments is that if I checkout a different branch, that branch is present in all the environments, right? It's just a matter of what branch you're on when you run setup.py? So I guess I need to install and then checkout master so no one else gets the wrong thing installed into the HERA environment by accident...

bhazelton:
I went back and forth about the units and I think I agree with you. What do you mean by "track this information in a hashtable in the Python file"?

bhazelton:
This has been tested on site in the HERA-test environment and seems to work.

The open question is whether we should really be putting the data in at the full rate (~1Hz) into the M&C database or whether we should reduce the data rate by doing some decimation or averaging or other statistics.

bhazelton:
Ok, I think this is ready for review. I've added support for reducing the data volume and added more sensors.

hera_mc/weather.py
```diff
@@ -0,0 +1,226 @@
...
+import numpy as np
+from astropy.time import Time
+from math import floor
+from sqlalchemy import Column, BigInteger, Float, String
+
+import tornado.gen
+
+from . import MCDeclarativeBase
+
+weather_sensor_dict = {'wind_speed': {'sensor_name': 'anc_mean_wind_speed',
```
dannyjacobs: Is this where we add if we want to get more sensors?

hera_mc/mc_session.py
```diff
@@ -748,6 +748,81 @@ def get_paper_temps(self, starttime, stoptime=None):
...
+        value: float
+            value from the sensor associated with the variable
+        """
+        from .weather import WeatherData
+
+        self.add(WeatherData.create(time, variable, value))
+
+    def add_weather_data_from_sensors(self, starttime, stoptime, variables=None):
+        """
+        Add weather data for a given variable and timespan from KAT sensors.
```
dannyjacobs: This function connects to the meerkat db and grabs the latest data using the "create_from_sensors" function.

hera_mc/weather.py
```diff
@@ -0,0 +1,226 @@
...
+
+    sensor_names = []
+    sensor_var_dict = {}
+    for var in variables:
+        if var not in weather_sensor_dict.keys():
+            raise ValueError('variable must be a key in weather_sensor_dict')
+        sensor_names.append(weather_sensor_dict[var]['sensor_name'])
+        sensor_var_dict[weather_sensor_dict[var]['sensor_name']] = var
+
+    portal_client = KATPortalClient('http://portal.mkat.karoo.kat.ac.za/api/client',
```
dannyjacobs: This is where the KAT URL is hard coded.  If we have to change it in the future, is this the only place it is?

hera_mc/mc_session.py
bhazelton: ok, I'll add that

hera_mc/weather.py
```diff
@@ -0,0 +1,226 @@
+# -*- mode: python; coding: utf-8 -*-
+# Copyright 2017 the HERA Collaboration
+# Licensed under the 2-clause BSD license.
+
+"""Handling quality metrics of HERA data.
```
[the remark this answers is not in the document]
bhazelton: thanks!

hera_mc/weather.py
bhazelton: Yes. You should also add their details to docs/weather_sensor_details.txt (you can get that using their example scripts).

hera_mc/weather.py
bhazelton: yes. Do you want me to put it at the top of the file?

hera_mc/tests/test_weather.py
```diff
@@ -0,0 +1,171 @@
...
+from hera_mc import mc, weather
+from hera_mc.tests import TestHERAMC
+
+
+def is_at_katcp_enabled_site():
+    hostname = socket.gethostname()
+    if hostname == 'qmaster':
+        return True
+    else:
+        return False
```
pkgw: This could just be:

```python
return (socket.gethostname() == 'qmaster')
```

The parentheses are of course not syntactically necessary, but I find that they help clarify that we're evaluating a boolean expression.

hera_mc/weather.py
```diff
@@ -0,0 +1,233 @@
...
+                       'rain': {'sensor_name': 'anc_weather_rain', 'units': 'mm',
+                                'reduction': 'sum', 'period': 60,
+                                'description': "Rainfall (report period: 10s)"}}
+
+
+def _reduce_time_vals(times, vals, period, strategy='decimate'):
+    if not isinstance(period, (int, np.int)):
+        raise ValueError('period must be an integer')
+
+    times_keep, inds = np.unique(np.floor(times / period) * period, return_index=True)
```
pkgw: Not a big deal but `times // period` is a slightly more idiomatic way to get the floor division. (The `//` operator in Python is even called `__floordiv__`.)

hera_mc/weather.py
```diff
@@ -0,0 +1,233 @@
...
+        inds = inds[1:]
+
+    if strategy == 'decimate':
+        vals_keep = vals[inds]
+    elif strategy == 'max':
+        vals_keep = []
+        times_keep = times_keep[:-1]
+        for count in range(len(times_keep)):
+            vals_keep.append(np.max(vals[inds[count]:inds[count + 1]]))
+        vals_keep = np.array(vals_keep)
```
pkgw: I feel like these stanzas could be made a bit more compact using a list comprehension or something, but it might hurt readability for no real gain.

I would however prefer to write `range(len(inds) - 1)` rather than the current expression. I think that's clearer than lopping off the last element of `times_keep`, which AFAICT is just done to reduce the length count. And that way it is clearer that the `range` call is tuned to index into the `inds` variable.

hera_mc/weather.py
```diff
@@ -0,0 +1,233 @@
...
+
+            # the value_timestamp is the sensor timestamp, while the other is
+            # when the recording system got it. The value_timestamp isn't always
+            # present, so test for it
+            if 'value_timestamp' in item._fields:
+                timestamp = item.value_timestamp
+            else:
+                timestamp = item.timestamp
+            # sometimes there are duplicates. Protect against that
+            if timestamp not in times:
```
pkgw: If this is what we're doing, `times` should be a `dict` or a `set`, for which the `x in y` test will be much more efficient. Seeing as the sensor data comes in at a high cadence this is one of the few times when the performance implications of that kind of test might have non-negligible implications.

hera_mc/weather.py
```diff
@@ -0,0 +1,233 @@
...
+                values.append(float(item.value))
+
+        reduction = weather_sensor_dict[variable]['reduction']
+        if reduction is None:
+            # decimate to 1Hz to prevent duplicate keys
+            reduction = 'decimate'
+            period = 1
+        else:
+            period = weather_sensor_dict[variable]['period']
+
```
pkgw: Would it be wiser to instead require that `weather_sensor_dict` always have `reduction` and `period` keys? Genuine question; I'm not sure what I think.

Docs/weather_sensor_details.txt
```diff
@@ -0,0 +1,142 @@
+Detail for sensor anc_mean_wind_speed:
```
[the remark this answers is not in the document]
bhazelton: I'll just get rid of it. Saul started it and I continued it, but I agree with your points.

hera_mc/weather.py
bhazelton: ok, I changed it and added a comment so I'll know what it means later :)

hera_mc/weather.py
bhazelton: sure, I'll change it.

hera_mc/weather.py
bhazelton: ok, I turned the times and values into a dict.

hera_mc/weather.py
bhazelton: Maybe that's clearer. I'll change it.

PR #175, opened 2017-11-29
Script cleanup

david-deboer:
This just fixes a few errors in scripts and adds a new hookup default


hera_mc/cm_hookup.py
```diff
@@ -130,6 +130,8 @@ def get_hookup(self, hpn_list, rev='ACTIVE', port_query='all', exact_match=False
                          Setting this makes get_hookup provide specific hookups (mimicking the
                          action before the file option was instituted)
         """
+        if True in [x[:2] != 'HH' for x in hpn_list]:
+            force_specific = True
```
pkgw: Umm ... if I've parsed the logic correctly, I think a more intuitive way to write this is:

```python
if not all(x[:2] == 'HH' for x in hpn_list):
    force_specific = True
```

PR #177, opened 2017-12-01
Wx script

david-deboer:
Did a bit more error checking in data.  Changes using keys (which aren't guaranteed in order).  First pass at script.  I thought it would write to teh actual database, but didn't...

### DIFF
--- a/hera_mc/weather.py
+++ b/hera_mc/weather.py
@@ -9,7 +9,7 @@ from __future__ import absolute_import, division, print_function
 
 import numpy as np
 from astropy.time import Time
-from math import floor
+from math import floor, isnan
 from sqlalchemy import Column, BigInteger, Float, String
 
 import tornado.gen
@@ -179,6 +179,9 @@ def _helper_create_from_sensors(starttime, stoptime, variables=None):
             # Since we can't do anything about those and the data might be bad, ignore them
             if item.status != 'nominal':
                 continue
+            # skip it if nan is supplied
+            if math.isnan(float(item.value)):
+                continue
 
             # the value_timestamp is the sensor timestamp, while the other is
             # when the recording system got it. The value_timestamp isn't always
@@ -191,17 +194,18 @@ def _helper_create_from_sensors(starttime, stoptime, variables=None):
             if timestamp not in time_vals.keys():
                 time_vals[timestamp] = float(item.value)
 
-        reduction = weather_sensor_dict[variable]['reduction']
-        period = weather_sensor_dict[variable]['period']
+        if len(time_vals.keys()):
+            reduction = weather_sensor_dict[variable]['reduction']
+            period = weather_sensor_dict[variable]['period']
 
-        times_use, values_use = _reduce_time_vals(np.array(time_vals.keys()),
-                                                  np.array(time_vals.values()),
-                                                  period, strategy=reduction)
+            times_use, values_use = _reduce_time_vals(np.array(time_vals.keys()),
+                                                      np.array(time_vals.values()),
+                                                      period, strategy=reduction)
 
-        for count, timestamp in enumerate(times_use.tolist()):
-            time_obj = Time(timestamp, format='unix')
-            weather_obj_list.append(WeatherData.create(time_obj, variable,
-                                                       values_use[count]))
+            for count, timestamp in enumerate(times_use.tolist()):
+                time_obj = Time(timestamp, format='unix')
+                weather_obj_list.append(WeatherData.create(time_obj, variable,
+                                                           values_use[count]))
 
     raise tornado.gen.Return(weather_obj_list)
 
--- a/hera_mc/weather.py
+++ b/hera_mc/weather.py
@@ -175,8 +175,7 @@ def _helper_create_from_sensors(starttime, stoptime, variables=None):
     weather_obj_list = []
     for sensor_name, history in histories.items():
         variable = sensor_var_dict[sensor_name]
-        duplicate_check = {}
-        sensor_times = []
+        sensor_times = [0.0]
         sensor_data = []
         for item in history:
             # status is usually nominal, but can indicate sensor errors.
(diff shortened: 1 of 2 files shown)


hera_mc/weather.py
```diff
@@ -189,19 +196,22 @@ def _helper_create_from_sensors(starttime, stoptime, variables=None):
...
-                                                  np.array(time_vals.values()),
-                                                  period, strategy=reduction)
-
-        for count, timestamp in enumerate(times_use.tolist()):
-            time_obj = Time(timestamp, format='unix')
-            weather_obj_list.append(WeatherData.create(time_obj, variable,
-                                                       values_use[count]))
+                time_vals[timestamp] = item.value
+                value_times.append(timestamp)
+                sensor_values.append(float(item.value))
```
bhazelton: Did you switch to lists because you actually saw a problem? I had started with lists but switched to dicts at Peter's suggestion. If you're worried that the key and value won't match because of the order not being guaranteed, see this: https://stackoverflow.com/questions/835092/python-dictionary-are-keys-and-values-always-the-same-order

david-deboer: As far as I could tell, the resampling schemes need monotonically (essentially) increasing values.  If the dict isn't doing anything else I can remove it.  In the random-ish key order, the resampling fails (or can -- it worked for me sometimes, but sometimes not).

bhazelton: Ah, I think when I wrote it in the first place I was just doing the decimate and that doesn't require that they're ordered, but the others might. If there's an example when it fails, we should add it to the tests to make sure it works. If ordering is an issue, I'm fine with either using lists in the first place or sorting them afterwards.

hera_mc/weather.py
```diff
@@ -168,17 +170,21 @@ def _helper_create_from_sensors(starttime, stoptime, variables=None):
...
     histories = yield portal_client.sensors_histories(sensor_names, starttime.unix,
-                                                      stoptime.unix, timeout_sec=60)
+                                                      stoptime.unix, timeout_sec=120)
 
     weather_obj_list = []
     for sensor_name, history in histories.items():
-        time_vals = {}
         variable = sensor_var_dict[sensor_name]
+        value_times = []
+        sensor_values = []
```
pkgw: It shouldn't be necessary to change these to lists: the first thing that happens in `_reduce_time_vals` is a call to `np.unique`, whose output gives a sorted version of the inputs.

Can you see if you can find a way to make it work with the dict as before? As I commented in the original PR, doing an `in` test with a list is much more expensive than with a dict, and the weather data come in at such a high cadence that the performance difference might actually matter.

scripts/mc_wx.py
```diff
@@ -0,0 +1,52 @@
...
+# Licensed under the 2-clause BSD license.
+
+"""This is meant to hold utility scripts for weather
+
+"""
+from __future__ import absolute_import, division, print_function
+
+from hera_mc import cm_utils, mc
+
+wx_options = ['wind_speed', 'wind_gust', 'wind_direction', 'humidity', 'pressure', 'temperature', 'rain']
```
pkgw: I don't think these should be hardcoded here because it duplicates information that's already stored in `hera_mc.weather`. Passing `None` for the `variables` arguments to the weather functions means "use the default list of sensors" so we should use that.

`wx_options` is also used for the `choices` keyword to limit the available value choices for the `--variables` argument but I think it might actually be better not to apply that constraint. You could imagine extending this script to be able to show values from any of the (many thousands) of sensor points that we can fetch from KATCP.

scripts/mc_wx.py
```diff
@@ -0,0 +1,52 @@
...
+            session = db.sessionmaker()
+            session.add_weather_data_from_sensors(start_time, stop_time, variables)
+        else:
+            from hera_mc import weather
+            wx = weather.create_from_sensors(start_time, stop_time, variables)
+            for w in wx:
+                displayTime = cm_utils.get_displayTime(Time(w.time, format='gps'))
+                print("{}: {} ({}) -- {}".format(w.variable, displayTime, w.time, w.value))
+    else:
+        print("Need valid start/stop times - or can specify last-period.")
```
pkgw: I think the script should exit with an error code if this situation arises. `sys.exit(1)`. You could also re-arrange the logic to read

```python
if not isinstance(start_time, Time) or not isinstance(stop_time, Time):
    print("Need valid start/stop times - or can specify last-period.", file=sys.stderr)
    sys.exit(1)
```

And then you can skip the `else`, so that the rest of the main program logic needs less identation.

david-deboer:
I made the suggested changes.  Note that for the dict I have to sort the return from np.unique to work.  So the speed comparison is `a_value not in list_b' vs `a_value not in b.keys()` and two `np.sorts`.

david-deboer:
The other problem though is that it doesn't actually write it to the database though!  It goes to the correct mc_session -- do I need to close it somehow in order to commit?

hera_mc/weather.py
```diff
@@ -57,6 +57,8 @@ def _reduce_time_vals(times, vals, period, strategy='decimate'):
 
     # the // operator is a floored divide.
     times_keep, inds = np.unique((times // period) * period, return_index=True)
+    times_keep = sorted(times_keep)
+    inds = sorted(inds)
```
pkgw: I am quite sure this is not what we want; since each array gets sorted on its own terms, the matching of times and indices will be broken unless both have the same rank-ordering in their respective arrays, i.e. that the times are already sorted. To sort both arrays using the same key, you'd use something like the return value from `np.argsort(times)`. But again, the way that `np.unique` works, everything should Just Work without having to apply any extra sorts before or after. Reading the code, I have trouble seeing how this might break down — the key requirement is the `vals` always be indexed through `inds`, which appears to be the case. Can you come up with a minimal example set of inputs that yields bad outputs?

hera_mc/weather.py
```diff
@@ -194,16 +196,16 @@ def _helper_create_from_sensors(starttime, stoptime, variables=None):
             else:
                 timestamp = item.timestamp
             # sometimes there are duplicates. Protect against that
-            if timestamp not in value_times:
-                value_times.append(timestamp)
-                sensor_values.append(float(item.value))
+            if timestamp not in sensor_data.keys():
```
pkgw: This should just be `if timestamp not in sensor_data:`.

pkgw:
Oh, right. For the rows not being added, I think you just need to call `session.commit()` at the end of that code path.

pkgw:
Maybe it will be helpful to start the decimate function off like this:

```python
times_keep, inds = np.unique((times // period) * period, return_index=True)
vals = vals[inds]
```

This will give two arrays, `times_keep` and `vals`, that are of the same size, are sorted in time, and contain no duplicate timestamps. It might be easier to reason about the rest of the function body given those as foundation.

david-deboer:
On-site is where I've been testing it (which explains all of the pushes...).

PR #178, opened 2017-12-04
Badants

david-deboer:
added the unproven ("dubitable") antenna cm structure for rtp


david-deboer:
Note that I did increase the weather logging in here, since I think we are keeping too much now...

hera_mc/cm_utils.py
```diff
@@ -126,9 +120,42 @@ def add_date_time_args(parser):
...
+
+def future_date():
+    """
+    Future is defined here, since defining a far FUTURE_DATE typically gives a
+    warning about UTC vs UT1 etc
+    """
+    return Time.now() + TimeDelta(300, format='jd')
+
+
+def get_stopdate(stop_date):
```
pkgw: This function is only ever called from `is_active`, so it should just be merged into that function.

hera_mc/cm_utils.py
```diff
@@ -126,9 +120,42 @@ def add_date_time_args(parser):
...
+    if isinstance(stop_date, Time):
+        return stop_date
+    else:
+        return future_date()
```
pkgw: This chain of logic should be pickier, in case someone accidentally passes a wrong argument to the function (or to `is_active` when they are merged). Assuming that the "default" case corresponds to an argument of None:

```python
if isinstance(stop_date, Time):
  return stop_date
if stop_date is None:
  return future_date()
raise ValueError('unexpected value for "stop_date" argument: %r' % (stop_date,))
```

(The parenthesization in the last line makes sure that the string formatting works reliably even if `stop_date` happens to have been passed in as a tuple.)

hera_mc/cm_utils.py
```diff
@@ -126,9 +120,42 @@ def add_date_time_args(parser):
...
+
+
+def get_displayTime(display):
```
pkgw: The naming of this function is odd. `get_display_time` instead? Even that I don't think is super transparent, though. `format_time_for_printing`?

hera_mc/cm_utils.py
```diff
@@ -126,9 +120,42 @@ def add_date_time_args(parser):
...
+    if isinstance(display, str) and display.lower() != 'now':
+        d = display
+    else:
+        d = get_astropytime(display)
+
+    if d is None:
+        d = 'None'
```
pkgw: If the intended functionality is that there are cases when it is desirable for the input value to be `None`, there should be a docstring documenting that fact. If not, the `None` case should raise a `ValueError`.

hera_mc/cm_utils.py
```diff
@@ -126,9 +120,42 @@ def add_date_time_args(parser):
...
+    elif isinstance(d, Time):
+        d = "{:%Y-%m-%d %H:%M:%S}".format(d.datetime)
+    return d
+
+
 def get_astropytime(_date, _time=0):
     """
-    Take in various incarnations of _date/_time and return an astropy.Time object
+    Take in various incarnations of _date/_time and return an astropy.Time object or None
```
pkgw: I worry about these kinds of swiss-army-knife functions since they make it easy to get sloppy about variable types (namely, the types of whatever variables the caller is passing to this function), which is pretty reliable route to buggy code.

One thing that this function clearly does well is handle textual command line arguments expressing times. That's a nicely focused functionality with a consistent input type. In the other cases that this function handles —  `astropy.Time` objects, integers, None, False — the function is not doing much. Can the non-text input cases be split into a function separate from the one that parses text? At the call sites, we should strive to avoid situations where the types of the date/time variables are ambiguous.

hera_mc/cm_utils.py
```diff
@@ -137,7 +164,10 @@ def get_astropytime(_date, _time=0):
         return_date = _date
     elif _date is None or _date is False:
         return_date = _date
-    else:
+    elif isinstance(_date, (int, long, float)):
+        if int(_date) > 1000000000:
+            return Time(_date, format='gps')
```
pkgw: If the input is 56646.6, this logic will lead to a return value of None, and not an exception. Is that the best behavior?

hera_mc/cm_utils.py
pkgw: Also, the overall clarity of this function could be improved greatly by peeling off cases with the general form

```
if (input is of type X):
   return (appropriate output)
```

Rather than chains of if-else statements and a `return_date` variable. Returning as early as possible prevents rightward drift and helps the reader understand which cases have been completely dealt with and therefore need not be worried about for the rest of the function.

hera_mc/cm_utils.py
```diff
@@ -146,7 +176,7 @@ def get_astropytime(_date, _time=0):
             return_date = None
         elif _date.lower() == 'now':
             return_date = Time.now()
-        elif _date is not None:
+        elif isinstance(_date, str):
```
pkgw: This `if` statement is redundant with the one that occurs at the higher level.

hera_mc/cm_utils.py
```diff
@@ -164,6 +194,9 @@ def get_astropytime(_date, _time=0):
                 except ValueError:
                     raise ValueError('Invalid format:  time should be H[:M[:S]] '
                                      '(HMS can be float or int)')
+        else:
+            raise TypeError("Not supported:  type {}".format(type(_date)))
+
```
pkgw: Due to the above redundancy, this `else` branch can never be executed.

hera_mc/cm_utils.py
```diff
@@ -222,35 +255,6 @@ def get_date_from_pair(d1, d2, ret='earliest'):
...
-        return stop_date
-    else:
-        return future_date()
-
-
-def is_active(at_date, start_date, stop_date):
-    stop_date = get_stopdate(stop_date)
-    return at_date >= start_date and at_date <= stop_date
-
-
```
pkgw: Note that these functions appear in the diffs even though their code hasn't changed because their position in the file has been reordered. It's better to avoid that when possible to reduce the noise in the diffs.

hera_mc/part_connect.py
```diff
@@ -292,6 +292,68 @@ def get_part_revisions(hpn, session=None):
...
+
+    __tablename__ = 'dubitable'
+
+    start_gpstime = Column(BigInteger, primary_key=True)
+    stop_gpstime = Column(BigInteger)
+    ant_list = Column(Text, nullable=False)
+
+    def __repr__(self):
+        return ('<{self.start_gpstime}>'.format(self=self))
+    #    return ('<{} dubitable entries>'.format(len(self.ant_list)))
```
pkgw: Please don't commit commented-out code.

Object's reprs should ideally convey their detailed contents, but this code only reports the start time. If there's a compelling reason to not include the other information, it should be reported in a comment.

hera_mc/part_connect.py
```diff
@@ -292,6 +292,68 @@ def get_part_revisions(hpn, session=None):
...
+
+    new_dubi = Dubitable()
+    new_dubi.start_gpstime = transition_gpstime + 1
+    new_dubi.ant_list = ''
+    for a in data:
+        new_dubi.ant_list += (str(a) + ',')
+    new_dubi.ant_list = new_dubi.ant_list.strip(',')
+    session.add(new_dubi)
+    session.commit()
+    data_dict = {'start_gpstime': [transition_gpstime], 'ant_list:len': [len(data)]}
```
pkgw: Is the thinking that the list of antennas might be so long that we won't want to include it in the various logfiles and whatnot? That doesn't seem very likely to occur, to me.

hera_mc/part_connect.py
```diff
@@ -292,6 +292,68 @@ def get_part_revisions(hpn, session=None):
...
+        session.add(old_dubi)
+    elif last_one.count() > 1:
+        raise ValueError('Too many open dubitable lists.')
+
+    new_dubi = Dubitable()
+    new_dubi.start_gpstime = transition_gpstime + 1
+    new_dubi.ant_list = ''
+    for a in data:
+        new_dubi.ant_list += (str(a) + ',')
+    new_dubi.ant_list = new_dubi.ant_list.strip(',')
```
pkgw: This can all just be `new_dubi.ant_list = ','.join(str(a) for a in ant_data)`.

hera_mc/part_connect.py
```diff
@@ -292,6 +292,68 @@ def get_part_revisions(hpn, session=None):
     return revisions
 
 
+class Dubitable(MCDeclarativeBase):
+    """
+    A table to track the dubitable antennas.
+
+    start_gpstime:  start time of list
+    stop_gpstime:  stop time of list
+    ant_list:  list of "unapproved" antennas
```
pkgw: This needs to be much more explicit about the contents of `ant_list`. Below I see that it is comma-separated, but that's not specified here. How are we expressing antennas? Prefixed part numbers? Simple integers? Etc. Can you add a cross-reference to the CM docs about what precise antenna identifier will be used here?

hera_mc/part_connect.py
```diff
@@ -292,6 +292,68 @@ def get_part_revisions(hpn, session=None):
...
+    """
+    update the database given a new list of dubitable antennas.
+    It will stop the previous list at transition_gpstime and start the new one
+    at transition_gpstime + 1sec
+
+    Parameters:
+    ------------
+    session:  db session to use
+    transition_gpstime:  gps time to make the change
+    data:  list of antennas
```
pkgw: This also needs to be more explicit about what `data` contains. Based on the code below, it looks like the list elements are expected to be integers? Can you add a cross-reference to the CM docs about what precise antenna identifier will be used here?

hera_mc/sys_handling.py
```diff
@@ -45,6 +45,39 @@ def cofa(self):
...
+
+        Parameters:
+        ------------
+        date:  something understandable by cm_utils.get_astropytime
+        return_full:  boolean to return the full object as opposed to the csv string
+        """
+
+        at_date = cm_utils.get_astropytime(date)
+        fnd = []
+        for dubi in self.session.query(part_connect.Dubitable):
```
pkgw: This is a little concerning because the sort of search that's happening here should be happening on the DB server side, rather than on our side. The dubitable table is probably not going to get big enough where the performance implications will be a big deal, but it's kind of ugly.

hera_mc/sys_handling.py
```diff
@@ -45,6 +45,39 @@ def cofa(self):
...
 
+    def get_dubitable_list(self, date='now', return_full=False):
+        """
+        Returns start_time and a list with the dubitable antennas.
+        If date is supplied, returns for that date, otherwise now.
+
+        Parameters:
+        ------------
+        date:  something understandable by cm_utils.get_astropytime
+        return_full:  boolean to return the full object as opposed to the csv string
```
pkgw: The return value needs to be documented more clearly, breaking out the relevant cases (I count three: None, str, tuple) and the semantics of each (including the individual elements when a tuple is returned).

It also seems like it might just be simpler to obey the "return_full" semantics always. The caller can ignore the boundary times and re-join the array if strings if that's what they want to do.

hera_mc/weather.py
```diff
@@ -25,19 +25,19 @@
 #   https://github.com/ska-sa/katportalclient/tree/master/examples
 weather_sensor_dict = {'wind_speed': {'sensor_name': 'anc_mean_wind_speed',
                                       'units': 'm/s',
-                                      'reduction': 'mean', 'period': 60,
+                                      'reduction': 'mean', 'period': 300,
```
pkgw: Really, please, separate unrelated changes into different pull requests. The git/GitHub history is a valuable asset for understanding the codebase and part of that value comes from keeping the intellectual content of the code well-matched to the concepts used by our change management tools.

hera_mc/sys_handling.py
david-deboer: This was a specific request to provide back exactly what rtp is looking for.  I wanted the ability to provide more for flexibility and didn't want another function so added return_full.  I've rewritten to try and be clearer.

hera_mc/sys_handling.py
david-deboer: The problem is how to handle None, which is what the current stop_gpstime is.  I also wanted to allow dates in the past that have a stop_gpstime.  Given how small the db is for this, it seems clearer to do it this way.

bhazelton:
Peter makes a good point about whitelisting vs blacklisting. Did this get discussed somewhere?

pkgw:
Yeah it did. Plan is for the initial blacklist to contain all ~310 not-known-good ants at first, so basically the same semantics as a whitelist.

hera_mc/cm_utils.py
```diff
@@ -126,47 +121,82 @@ def add_date_time_args(parser):
...
+        d = 'None'
+    elif isinstance(d, Time):
+        d = "{:%Y-%m-%d %H:%M:%S}".format(d.datetime)
+    return d
+
+
 def get_astropytime(_date, _time=0):
     """
-    Take in various incarnations of _date/_time and return an astropy.Time object
+    Take in various incarnations of _date/_time and return an astropy.Time object or None.
```
pkgw: This docstring doesn't specify what values of `_date` are actually acceptable.

hera_mc/cm_utils.py
```diff
@@ -126,47 +121,82 @@ def add_date_time_args(parser):
...
+        if _date.lower().replace('/', '') == 'na' or _date.lower() == 'none':
+            return None
+        _date = _date.replace('/', '-')
+        try:
+            return_date = Time(_date, scale='utc')
+        except ValueError:
+            raise ValueError('Invalid format:  date should be YYYY/M/D or YYYY-M-D')
+        if isinstance(_time, (float, int)):
+            return return_date + TimeDelta(_time * 3600.0, format='sec')
+        if isinstance(_time, str) and ':' in _time:
```
pkgw: So, if this function is called to parse a command line date/time argument, the `_time` argument will be passed to this function as a `str` object. If the user gave the time as just an integer (eg "11"), this function will raise a ValueError since this branch only gets triggered if the `_time` string contains a ":" as well. It seems like it would be helpful for it too handle the no-colon case too, especially since the case numeric inputs is handled OK.

Also note that if the time is given as "11:00:59:ABC", the code will silently handle it as if the user gave "11:00". I generally think parsing should be conservative about these kinds of cases since a mistake in this sort of input could easily signify that something bigger has gone wrong in whatever code is driving this routine.

hera_mc/sys_handling.py
pkgw: OK.

hera_mc/sys_handling.py
pkgw: One can map the Python values into SQL values by building up the query in pieces, using Python logic to decide whether to add extra terms like the stop time check on top of the start time check. The Librarian does this a lot. But as we've both said, this is unlikely to be a big deal in practice.

PR #181, opened 2017-12-09
Fixcache

david-deboer:
This (starts) to address the issue of the (unexpected) frequent rewriting of the cache file.

It corrects an undesirable item where the system requests-to-hookup addressed the database instead of the cache file.  This _seems_ related to the cache rewriting issue, but I'm not sure (I can't point to the guilty code).  I've added a cm_log entry when it gets rewritten to help diagnose for the future.

### DIFF
--- a/hera_mc/cm_hookup.py
+++ b/hera_mc/cm_hookup.py
@@ -128,6 +128,7 @@ class Hookup:
         exact_match:  boolean for either exact_match or partial
         show_levels:  boolean to include correlator levels
         force_new:  boolean to force a full database read as opposed to checking file
+                    this will also rewrite the cache-file
         force_specific:  boolean to force this to read/write the file to use the supplied values
                          Setting this makes get_hookup provide specific hookups (mimicking the
                          action before the cache file option was instituted)
--- a/hera_mc/cm_hookup.py
+++ b/hera_mc/cm_hookup.py
@@ -194,7 +194,7 @@ class Hookup:
         for k, part in parts.iteritems():
             if not cm_utils.is_active(self.at_date, part['part'].start_date, part['part'].stop_date):
                 continue
-            huk = k.upper()
+            huk = k
             hookup_dict['hookup'][huk] = {}
             pols_to_do = self.__get_pols_to_do(part, port_query)
             for pol in pols_to_do:
--- a/hera_mc/cm_hookup.py
+++ b/hera_mc/cm_hookup.py
@@ -95,7 +95,7 @@ class Hookup:
             self.cached_hookup_dict = self.__get_hookup(hpn_list=self.hookup_list_to_cache, rev='ACTIVE',
                                                         port_query='all', at_date=self.at_date,
                                                         exact_match=False, show_levels=False)
-            self.__write_hookup_cache_to_file()
+            self.__write_hookup_cache_to_file(force_new)
         elif self.cached_hookup_dict is None:
             self.__read_hookup_cache_from_file()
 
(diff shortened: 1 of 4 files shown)


hera_mc/cm_hookup.py
```diff
@@ -71,30 +74,28 @@ def __init__(self, at_date='now', session=None, hookup_list_to_cache=['HH']):
...
+        If new one is needed, it re-writes the cache file.
 
         Parameters
         -----------
-        force_new:  boolean to force a new read as opposed to using file, if valid.
-        provided_hookup:  will set the cached_hookup_dict to this, if present
-        reset_cache:  will reset cached_hookup_dict to None, if True
+        force_new:  boolean to force a new read as opposed to using (valid) file.
+        provided_hookup:  will set the cached_hookup_dict to this, if present (only does this)
+        reset_cache:  will reset cached_hookup_dict to None, if True (only does this)
```
pkgw: I have a bit of trouble understanding what these parentheticals mean ...

hera_mc/cm_hookup.py
```diff
@@ -180,6 +183,7 @@ def __get_hookup(self, hpn_list, rev, port_query, at_date,
                      exact_match=False, show_levels=False):
         """
         This gets called by the get_hookup wrapper if the database is to be read.
+        It is the full original method.
```
pkgw: This new doc text also confuses me.

hera_mc/cm_hookup.py
```diff
@@ -191,11 +195,12 @@ def __get_hookup(self, hpn_list, rev, port_query, at_date,
         for k, part in parts.iteritems():
             if not cm_utils.is_active(self.at_date, part['part'].start_date, part['part'].stop_date):
                 continue
-            hookup_dict['hookup'][k] = {}
+            huk = k
```
pkgw: Is there a reason to (essentially) rename this variable in this way? From my read of the code it seems as if one could write `for huk, par in parts.iteritems()`. Or just leave it as `k`.

If there's a reason for this that's escaping me, it should be documented in a comment.

hera_mc/cm_hookup.py
```diff
@@ -419,14 +424,18 @@ def add_correlator_levels(self, hookup_dict):
...
+    def __write_hookup_cache_to_file(self, force_new):
         with open(self.hookup_cache_file, 'wb') as f:
             np.save(f, self.at_date)
             np.save(f, cm_utils.stringify(self.hookup_list_to_cache))
             np.save(f, self.cached_hookup_dict)
             np.save(f, self.part_type_cache)
+        log_dict = {'hu-list': cm_utils.stringify(self.hookup_list_to_cache),
+                    'check_hookup': str(self.__check_hookup_cache_file_date()),
+                    'force_new': str(force_new)}
+        cm_utils.log('update_cache', log_dict=log_dict)
```
pkgw: If I'm reading this correctly, the only thing this function does with the `force_new` argument is log it. The function shouldn't require arguments that it doesn't really use. With the right logging info we can probably infer the value of `force_new` anyway.

hera_mc/cm_hookup.py
david-deboer: As soon as this PR is done, I'm going to start a new one where I resolve a worry I have about the hookup key.  I can reset it for now...

hera_mc/cm_hookup.py
david-deboer: Yes - and as soon as we figure out what is going on I'll remove this.  This was just an expedient way to do it for now.

hera_mc/cm_hookup.py
david-deboer: I've tried to clarify both doc strings

PR #183, opened 2017-12-09
Add Roach temperature code and table. Closes #180

### LINKED ISSUE #180: Start monitoring ROACH temperatures
As per HERA-Team/HERA_Commissioning#54. In-container temperatures are oscillating in a way that affects the data, and @jack-h has added monitoring points in the correlator redis database that record the temps in a better way than we have right now. We should start recording these data into M&C ASAP — this could be pretty important for H1C analysis going forward.

bhazelton:
This has been tested at site and it seems to work fine.


hera_mc/mc_session.py
```diff
@@ -825,6 +825,72 @@ def get_weather_data(self, starttime, stoptime=None, variable=None):
                                  stoptime=stoptime, filter_column='variable',
                                  filter_value=variable)
 
+    def add_roach_temperature(self, time, roach, ambient_temp, inlet_temp,
+                              outlet_temp, fpga_temp, ppc_temp):
+        """
+        Add new roach temperature to the M&C database.
```
pkgw: For posterity, it might help to define what a "roach" is.

hera_mc/mc_session.py
```diff
@@ -825,6 +825,72 @@ def get_weather_data(self, starttime, stoptime=None, variable=None):
...
+
+        Parameters:
+        ------------
+        time: astropy time object
+            astropy time object based on a timestamp from the katportal sensor.
+        roach: int
+            roach number
```
pkgw: Perhaps these should be strings? One could imagine a future where there's not a clear and easy numbering of the ROACHes.

hera_mc/roach.py
```diff
@@ -0,0 +1,127 @@
...
+    redis_dict: A dict spoofing the return dict from _get_redis_dict for testing
+        purposes. Default: None
+
+    Returns:
+    -----------
+    A list of RoachTemperature objects
+    """
+
+    roach_obj_list = []
+    for i in range(1, 9):
```
pkgw: This feels like something that shouldn't be hardcoded here.

hera_mc/mc_session.py
bhazelton: sure, I'll change it.

hera_mc/mc_session.py
bhazelton: I'll add a little more info.

hera_mc/roach.py
bhazelton: yeah, I agree, I just used some code from Jack. @jack-h can you suggest something better?

bhazelton: It's not much better, but I moved the hardcoded list of roach names to the top of roach.py. I'm happy for other suggestions or we can just go with this.

PR #184, opened 2017-12-11
Fixcache2

david-deboer:
This is another attempt to see about fixing the cache issue, which includes an addition to the hookup script to examine cache file info.  It tries to clean it up more and hopefully clarify it.


david-deboer:
Note - I still want to do this PR regardless of the HERA/RTP environment discussion...

hera_mc/cm_hookup.py
```diff
@@ -168,6 +182,18 @@ def get_hookup(self, hpn_list, rev='ACTIVE', port_query='all', exact_match=False
             hookup_dict = self.add_correlator_levels(hookup_dict)
         return hookup_dict
 
+    def __double_check_request(self, hpn_list):
```
pkgw: The name of this function is too vague.

hera_mc/cm_hookup.py
```diff
@@ -446,17 +483,22 @@ def __check_hookup_cache_file_date(self, contemporaneous_minutes=5.0):
             raise
         result = self.session.query(cm_transfer.CMVersion).order_by(cm_transfer.CMVersion.update_time).all()
         cm_hash_time = Time(result[-1].update_time, format='gps')
-        file_mod_time = Time(stats.st_mtime, format='unix')
+        file_mod_time = Time(stats.st_ctime, format='unix')
```
pkgw: We're changing from looking at file modification time to creation time here. Is that really what we want? Modification time does seem like the more relevant quantity. And all of the variable names and docstrings still refer to this quantity as the "modification" time.

hera_mc/cm_hookup.py
```diff
@@ -466,6 +508,21 @@ def __read_hookup_cache_from_file(self):
...
             self.part_type_cache = np.load(f).item()
 
+    def show_hookup_cache_file_info(self):
+        if not os.path.exists(self.hookup_cache_file):
+            return "{} does not exist.".format(self.hookup_cache_file)
+        self.__read_hookup_cache_from_file()
+        print('Cache file:  {}'.format(self.hookup_cache_file))
+        print('Cache time:  {}'.format(cm_utils.get_time_for_display(self.cached_at_date)))
+        stats = os.stat(self.hookup_cache_file)
+        file_mod_time = Time(stats.st_ctime, format='unix')
```
pkgw: Same mtime/ctime query.

hera_mc/cm_hookup.py
```diff
@@ -466,6 +508,21 @@ def __read_hookup_cache_from_file(self):
             self.cached_hookup_dict = np.load(f).item()
             self.part_type_cache = np.load(f).item()
 
+    def show_hookup_cache_file_info(self):
+        if not os.path.exists(self.hookup_cache_file):
+            return "{} does not exist.".format(self.hookup_cache_file)
```
pkgw: The main line of this function prints information, but this case returns a string. For consistency I think this case should print the message and just return.

hera_mc/tests/test_sys_handling.py
```diff
@@ -33,6 +33,12 @@ def test_publish_summary(self):
         msg = self.h.publish_summary()
         self.assertEqual(msg, 'Not on "main"')
 
+    def test_other_hookup(self):
+        H = cm_hookup.Hookup(at_date='now', session=self.test_session)
+        H.reset_cache(None)
+        self.assertEqual(H.cached_hookup_dict, None)
+        H.show_hookup_cache_file_info()
```
pkgw: Can we devise a higher-level test here? The assertion right now is basically testing that we can successfully set a variable to None, which is not super informative.

Also, the call to `show_hookup_cache_file_info` seems a bit out of place here. I think that should be its own test case.

scripts/hookup.py
```diff
@@ -27,6 +27,7 @@
     parser.add_argument('--show-levels', help="Show power levels if enabled (and able) [False]", dest='show_levels', action='store_true')
     parser.add_argument('--show-ports', help="Show ports on hookup.", dest='show_ports', action='store_true')
     parser.add_argument('--show-revs', help="Show revs on hookup.", dest='show_revs', action='store_true')
+    parser.add_argument('--show-cache-file-info', help="Shows information about the hookup cache file.", dest='cache_info', action='store_true')
```
pkgw: This argument should be named `show_cache_info` for consistency and clarity.

david-deboer:
I addressed the comments, and "helpfully" rebased on the new master -- hopefully that doesn't confuse it.

pkgw:
I think the rebase/merge should be fine but GitHub does seem a bit confused. I'm going to fool around a bit to try to straighten it out.

david-deboer:
Thanks and sorry.  I thought that rebasing on the most recent master on my machine would be safe, but then as soon as I did it I worried...  It _should_ have just been like I did my work on top of the most recent master, correct?

pkgw:
Well, somehow there's a recent merge commit showing up in the PR (4807bef4b38d0f83192c7124fe12bc6f0286c4dd) which shouldn't show up if the rebase is done in the cleanest way. But your branch seems to merge on to master in "fast-forward" mode without conflicts so I think it's all basically fine, just GitHub's UI is confused.

scripts/hookup.py
```diff
@@ -27,7 +27,7 @@
     parser.add_argument('--show-levels', help="Show power levels if enabled (and able) [False]", dest='show_levels', action='store_true')
     parser.add_argument('--show-ports', help="Show ports on hookup.", dest='show_ports', action='store_true')
     parser.add_argument('--show-revs', help="Show revs on hookup.", dest='show_revs', action='store_true')
-    parser.add_argument('--show-cache-file-info', help="Shows information about the hookup cache file.", dest='cache_info', action='store_true')
+    parser.add_argument('--show-cache-info', help="Shows information about the hookup cache file.", dest='cache_info', action='store_true')
```
pkgw: I meant the `dest` setting here, not the command-line argument name (although I think the shorter argument name is probably an improvement too).

PR #185, opened 2017-12-12
Viewwx

david-deboer:
Adds a weather Handling class, which currently gets reads the table and writes files (and then also can read those files).  A flag has been added to mc_wx.py (-W) which dumps the weather data to files.

### DIFF
--- a/hera_mc/tests/test_weather.py
+++ b/hera_mc/tests/test_weather.py
@@ -157,6 +157,36 @@ class TestWeather(TestHERAMC):
 
             self.assertRaises(ValueError, weather.create_from_sensors, t1, t2, variables='foo')
 
+    def test_read_weather_table(self):
+        t1 = Time('2016-01-10 01:15:23', scale='utc')
+        t2 = t1 + TimeDelta(120.0, format='sec')
+
+        wind_speeds = [2.5487029167, 2.5470608333]
+        wind_directions = [239.382, 223.11]
+        temperatures = [11.505, 12.29]
+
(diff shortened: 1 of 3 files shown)


pkgw:
What's the motivation for this PR? If we're basically interested in having a way to dump the weather data to a file for diagnostics, a bunch of this code could be eliminated. I.e., nothing actually uses the `read_weather_files` function right now.

I think this is particularly important to think about since we can expect the weather tables to get quite sizeable. As written, any operations involving the `weather.Handling` class are eventually going to involve gigantic dictionaries that will be extremely slow to work with. If we don't actually need a general-purpose "all weather data from all time" data structure, we can almost surely design it to be much more efficient for whatever our actual use case is.

david-deboer:
We want to look at weather data and this is a start.  I could write my own code to do it on my machine and Danny could write his own code to do it on his machine and ...  so I thought I'd give some basic starter functions/structure.  A logical extension is to allow frequency ranges etc etc, which I'll likely do but there was a call in HERA_Commissioning to look at weather data and this was the underlying code.

bhazelton:
I think we need to understand the Commissioning issue a bit more. It seems very inefficient to write it to a file when it already exists in the database and there are methods to get it all out for a given time range. Are you planning to copy the files somewhere else to plot them? That seems a bit hacky and as Peter points out, the files are likely to get very big. Is it a bad idea to plot them on qmaster?

pkgw:
Are you referring to HERA-Team/HERA_Commissioning#80? (GitHub cross-references the mentions so it's helpful to drop them in the discussions explicitly.)

PR #187, opened 2017-12-12
Alternate version of "viewwx"

pkgw:
This one avoids a class that mirrors the whole database in-memory.

It also leverages the existing `mc_wx.py` script to allow the user to filter on start and stop dates and which variables should be printed.

### DIFF
--- a/hera_mc/__init__.py
+++ b/hera_mc/__init__.py
@@ -70,6 +70,7 @@ MCDeclarativeBase.__repr__ = MethodType(MCDeclarativeBase_repr, None, MCDeclarat
 MCDeclarativeBase.isclose = MethodType(MCDeclarativeBase_close, None, MCDeclarativeBase)
 
 import logging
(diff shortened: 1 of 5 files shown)


david-deboer:
Without the full branch I can't tell a few things:  
 - what happens if start_time is None?
 - in the script, does it leave after writing the files (ie. not do anything else)?
Can you push the branch?

pkgw:
I've pushed the branch to the HERA-Team repo. Alternatively, in the diff view, you can click the little expand-looking icon to show the surrounding code:

screenshot from 2017-12-12 20-39-54

Or with a bit of command-line futzing you can pull down the branch from my GitHub user account, but there is some futzing required.

As to the questions:

- None for `start_time` is symmetric with how it works for `stop_time`. If you just specify a `stop_time`, you get all entries before the specified time. If you specify netiher, you get absoluteley everything, as per your version.
- Yeah, when the `-W` mode is invoked, nothing else happens.

PR #190, opened 2017-12-16
Make the obsids optional on Librarian files.

pkgw:
The Librarian is going to start having some files without obsids, for tracking maintenance files (e.g. weekly database backups) that are not tied to data-taking. M&C should still know about such files, though, so it needs to learn to deal with them.


bhazelton:
Disregard my previous question, there is already a time column. I just didn't read carefully enough.

PR #191, opened 2017-12-18
scripts/mc_monitor_roach_temps.py: add ROACH monitor

### LINKED ISSUE #189: make a script to put roach temps in database

pkgw:
This is being tested on-site and so far seems to work.

Closes #189.

### DIFF
--- a/hera_mc/mc_session.py
+++ b/hera_mc/mc_session.py
@@ -904,17 +904,43 @@ class MCSession(Session):
                                          outlet_temp, fpga_temp, ppc_temp))
 
     def add_roach_temperature_from_redis(self):
-        """
-        Read and add roach (fpga correlator board) temperatures from the Redis database
-        This function connects to the Redis database and grabs the latest data
-        using the "create_from_redis" function.
+        """Read and add ROACH (FPGA correlator board) temperatures from the Redis
+        database. This function connects to the Redis database and grabs the
+        latest data using the `create_from_redis` function.
+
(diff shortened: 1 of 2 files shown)


pkgw:
It does produce a fairly annoying amount of log chatter, although since the system is set up as a daemon it's a bit hidden. So, the temperatures change only once a minute? Should I just change the polling interval to be exactly 60 seconds?

pkgw:
Looking into this a bit, I think we can get fairly ideal behavior by using some postgres-specific constructs that allow the DB server to decide to abandon the INSERTs if the relevant rows already exist: [cf these SQLAlchemy docs](http://docs.sqlalchemy.org/en/latest/dialects/postgresql.html#insert-on-conflict-upsert). That way we can be sure that we're not missing samples, but not have to do clunky things to detect duplicate inserts.

The downside would be that it looks like one would have to manually construct the INSERT commands. Given that this table is quite simple and we don't expect to make complex use of the `RoachTemperature` objects, I think this is OK. I envision a new keyword on `MCSession.add_roach_temperature_from_redis` such that by setting `postgres=True` one can activate the specialized code. What do you think?

bhazelton:
That sounds great to me! I was worried about dropping the polling to exactly 60 seconds because you'd likely miss one every so often.

pkgw:
OK, I think the PostgreSQL magic is working. New rows are appearing in M&C and the script is not longer emitting a bunch of exceptions.

PR #194, opened 2018-01-18
Sqlite

david-deboer:
Only changed the documentation in INSTALL.md to reflect what is needed for sqlite to work.


hera_mc/cm_transfer.py
```diff
@@ -259,3 +259,28 @@ def _initialization(session=None, cm_csv_path=None, tables='all', base=False,
                         setattr(table_inst, field_name[i], r)
                     session.add(table_inst)
                     session.commit()
+
+
+def convert_all_csv_to_sqlite3():
```
pkgw: Does this function ever get called? Your description of the PR makes me wonder if you intended to include this change.

david-deboer:
It doesn't get called - I thought I had deleted it...  I've now removed the convert_all_cm.... function

PR #197, opened 2018-01-20
Add rtp_task_resource_record table

plaplant:
This PR adds the "RTP task resource record", which is a way to keep track of the performance of individual tasks of RTP. For every given obsid and task name, a start and stop time is recorded. Also if available, the maximum memory used and average CPU load are recorded. This table will help us monitor how the computation requirements of individual tasks change as we continue to add more antennas.


hera_mc/mc_session.py
```diff
@@ -714,6 +714,53 @@ def get_rtp_process_record(self, starttime, stoptime=None, obsid=None):
...
+        -----------
+        list of RTPTaskResourceRecord objects
+
+        """
+        from .rtp import RTPTaskResourceRecord
+
+        return self.query(RTPTaskResourceRecord).filter(
+            getattr(RTPTaskResourceRecord, 'obsid') == obsid,
+            getattr(RTPTaskResourceRecord, 'task_name') == task_name).limit(1).all()
+
```
bhazelton: You might take a look at the interface other getters in this object use. There are nice things about keeping them consistent, but if you strongly prefer this one I can live with it.

pkgw: Also ... is there a reason for `getattr(RTPTaskResourceRecord, 'obsid')` as opposed to just `RTPTaskResourceRecord.obsid`?

plaplant:
@bhazelton okay, I'll change the getter to be similar to the others.

@pkgw I'll update the docs as well. I think the body of the getter function is going to change, but that syntax is good to know for the future.

PR #199, opened 2018-01-30
Ren parts paper

david-deboer:
This just renames the parts-paper table to parts


david-deboer:
I assume it is failing because I need to somehow alembic upgrade the online version?  It passes locally.

david-deboer:
Strike that last comment -- I had errantly pushed master (unsuccessfully) rather than the needed fix to this branch...

alembic/versions/e33c1d5684cf_changed_parts_paper_table_to_parts.py
```diff
@@ -0,0 +1,28 @@
...
+
+# revision identifiers, used by Alembic.
+revision = 'e33c1d5684cf'
+down_revision = '3d3c72ecbc0d'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
```
bhazelton: fix this line -- it's not true

bhazelton: you also need to fix the one on line 26.

PR #200, opened 2018-01-31
Use aipy instead of pyuvdata to query file metadata

plaplant:
This PR uses `aipy` instead of `pyuvdata` to query the metadata necessary for uploading files to the librarian. This is necessary on paper1.

### DIFF
--- a/scripts/mc_add_observation.py
+++ b/scripts/mc_add_observation.py
@@ -3,31 +3,35 @@
 # Copyright 2017 the HERA Collaboration
 # Licensed under the 2-clause BSD license.
 
-import argparse,pyuvdata,aipy
+import argparse, aipy
 from hera_mc.observations import Observation
 from hera_mc import mc
-from astropy.time import Time,TimeDelta
+from astropy.time import Time, TimeDelta
+
 a = mc.get_mc_argument_parser()
 a.description = """Read the obsid from a file and create a record in M&C."""
-a.add_argument('files',metavar='file',type=str,nargs='*',default=[],
-        help='*.uv files to extract')
+a.add_argument('files', metavar='file', type=str, nargs='*', default=[],
+               help='*.uv files to extract')
 args = a.parse_args()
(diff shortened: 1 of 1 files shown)


scripts/mc_add_observation.py
```diff
@@ -3,31 +3,35 @@
...
+a.add_argument('files', metavar='file', type=str, nargs='*', default=[],
+               help='*.uv files to extract')
 args = a.parse_args()
 db = mc.connect_to_mc_db(args)
 
 
 for uvfile in args.files:
-    #we need aipy to read the stopt and duration because everything is aweful
-    #but first tell AIPY we're allowed to read these variables
+    # we need aipy to read the stopt and duration because everything is awful
```
pkgw: It is fun to complain but at some point we should make this comment a bit clearer.

scripts/mc_add_observation.py
```diff
@@ -3,31 +3,35 @@
...
+    # but first tell AIPY we're allowed to read these variables
```
pkgw: The `itemtable` modifications aren't needed anymore since we patched AIPY. That being said, it is strange to me that `startt` doesn't need to be defined. And `duration` isn't used here.

scripts/mc_add_observation.py
```diff
@@ -3,31 +3,35 @@
...
-    UVData = pyuvdata.UVData()
-    UVData.read_miriad(uvfile)
-    obsid = UVData.extra_keywords['obsid']
-    starttime = Time(UVData.extra_keywords['startt'],scale='utc',format='gps')
+    obsid = UV['obsid']
 
     with db.sessionmaker() as session:
+        obs = session.get_obs(obsid)
+        if len(obs) > 0:
+            print("observation {obs} already in M&C, skipping".format(obs=obsid))
```
pkgw: Because there is an `obs` variable in this loop, would be clearer to replace `obs` with `obsid` in this print statement.

PR #201, opened 2018-02-01
Issue182

david-deboer:
What started as a branch to fix issue 182 became a general clean-up (mostly stemming from the fixing of the issue...).


david-deboer:
Travis seems to be stuck...?

PR #204, opened 2018-02-05
Syscomment

david-deboer:
This has a system logging feature requested by Jack.


hera_mc/sys_handling.py
```diff
@@ -314,3 +314,17 @@ def publish_summary(self, hlist=['HH'], rev='A', exact_match=False,
             return 'OK'
         else:
             return 'Not on "main"'
+
+    def system_comments(self, system_kw='System', kword='all', tabkw=7):
+        longest_comment = 0
+        s = ''
+        for P in self.session.query(part_connect.PartInfo).filter((func.upper(part_connect.PartInfo.hpn) == system_kw.upper())):
```
pkgw: One-letter variable names should be lower case.

hera_mc/sys_handling.py
```diff
@@ -314,3 +314,17 @@ def publish_summary(self, hlist=['HH'], rev='A', exact_match=False,
             return 'OK'
         else:
             return 'Not on "main"'
+
+    def system_comments(self, system_kw='System', kword='all', tabkw=7):
```
pkgw: I see what `tabkw` does here but the variable name is pretty opaque to me. Also, the one place that calls this function never passes a value for that keyword. It seems like it would be better to dynamically adjust the column size in the same way that the comment column is adjusted.

hera_mc/sys_handling.py
```diff
@@ -314,3 +314,17 @@ def publish_summary(self, hlist=['HH'], rev='A', exact_match=False,
...
+        s = ''
+        for P in self.session.query(part_connect.PartInfo).filter((func.upper(part_connect.PartInfo.hpn) == system_kw.upper())):
+            if kword.lower() == 'all' or P.hpn_rev.lower() == kword.lower():
+                commlib = '{}  <{}>'.format(P.comment, P.library_file)
+                display_time = cm_utils.get_time_for_display(P.posting_gpstime)
+                s += "{:{tkw}s} | {:{pt}s} | {}\n".format(P.hpn_rev, display_time, commlib, tkw=tabkw, pt=len(display_time))
+                if len(commlib) > longest_comment:
+                    longest_comment = len(commlib)
+        header = "\n{:{tkw}s} | {:{pt}s} | {}  <{}>\n".format('Keyword', 'Posting', 'Comment', 'file/url', tkw=tabkw, pt=len(display_time))
+        header += "{}+{}+{}\n".format((tabkw + 1) * '-', (len(display_time) + 2) * '-', (longest_comment + 1) * '-')
```
pkgw: This doesn't *need* to be changed, but it's more efficient to build up a list of strings and then convert them into a multi-line string all in one go with `'\n'.join(line_list)`. Repeatedly extending long strings, as is done here, is an operation that can get quite slow.

scripts/cm_syslog.py
```diff
@@ -0,0 +1,44 @@
+#! /usr/bin/env python
+# -*- mode: python; coding: utf-8 -*-
+# Copyright 2017 the HERA Collaboration
```
pkgw: It's 2018 now.

scripts/cm_syslog.py
```diff
@@ -0,0 +1,44 @@
...
+from __future__ import print_function
+
+from hera_mc import mc, cm_utils, part_connect, sys_handling
+
+SYSTEM = 'System'  # word used in part_info as hpn for these comments
+
+
+if __name__ == '__main__':
+    parser = mc.get_mc_argument_parser()
+    parser.add_argument('-k', '--keyword', help="Keyword for comments [general]", default='geneRAL')
```
pkgw: What is with the weird capitalization of `geneRAL`??

PR #208, opened 2018-02-07
trying to repair cache file issue PLEASE MERGE

david-deboer:
Testing for now


david-deboer:
Found the bug, which in normal mode still worked but made it slow.

PR #209, opened 2018-02-13
Health

david-deboer:
This is a branch that starts putting in "health" functions so that I can better spot issues of e.g. multiple connections as just happened.

### DIFF
--- a/hera_mc/cm_health.py
+++ b/hera_mc/cm_health.py
@@ -20,11 +20,11 @@ class RevisionError(Exception):
         super(RevisionError, self).__init__(message)
 
 
-def check_for_overlap(start_i, stop_i, start_j, stop_j):
-    if start_j <= start_i:
-        if stop_j > start_i:
+def check_for_overlap(interval_i, interval_j):
+    if interval_j[0] <= interval_i[0]:
+        if interval_j[1] > interval_i[0]:
             return True
-    elif start_j <= stop_i:
+    elif interval_j[0] <= interval_i[1]:
             return True
     return False
 
@@ -87,20 +87,14 @@ class Connections:
         for k in self.multiples:
             for i in range(len(self.conndict[k])):
                 for j in range(i):
-                    start_i = self.conndict[k][i][0]
-                    stop_i = self.conndict[k][i][1]
-                    start_j = self.conndict[k][j][0]
-                    stop_j = self.conndict[k][j][1]
-                    if check_for_overlap(start_i, stop_i, start_j, stop_j):
+                    if check_for_overlap(self.conndict[k][i], self.conndict[k][j]):
                         self.duplicates.append([k, i, j])
         if len(self.duplicates):
             print('{} duplications found.'.format(len(self.duplicates)))
             for d in self.duplicates:
-                start_i = self.conndict[d[0]][d[1]][0]
-                stop_i = self.conndict[d[0]][d[1]][1]
-                start_j = self.conndict[d[0]][d[2]][0]
-                stop_j = self.conndict[d[0]][d[2]][1]
-                print('\t{} <1>{}-{}  <2>{}-{}'.format(k, start_i, stop_i, start_j, stop_j))
+                i = self.conndict[d[0]][d[1]]
+                j = self.conndict[d[0]][d[2]]
+                print('\t{} <1>{}-{}  <2>{}-{}'.format(k, i[0], i[1], j[0], j[1]))
         else:
             print('No duplications found.')
         print('Out of {} connections checked.'.format(self.num_connections))
@@ -145,11 +139,9 @@ def check_part_for_overlapping_revisions(hpn, session=None):
     revisions = cm_revisions.get_all_revisions(hpn, session)
     for i in range(len(revisions)):
         for j in range(i):
-            start_i = revisions[i].started
-            stop_i = cm_utils.get_stopdate(revisions[i].ended)
-            start_j = revisions[j].started
-            stop_j = cm_utils.get_stopdate(revisions[j].ended)
-            if check_for_overlap(start_i, stop_i, start_j, stop_j):
+            interval_i = [revisions[i].started, cm_utils.get_stopdate(revisions[i].ended)]
+            interval_j = [revisions[j].started, cm_utils.get_stopdate(revisions[j].ended)]
+            if check_for_overlap(interval_i, interval_j):
                 self.overlap.append([revisions[i], revisions[j]])
 
     if len(overlap) > 0:
--- a/hera_mc/cm_health.py
+++ b/hera_mc/cm_health.py
@@ -2,7 +2,7 @@
 # Copyright 2018 the HERA Collaboration
 # Licensed under the 2-clause BSD license.
 
-"""Functions to handle various cm table health checks.
+"""Functions to handle various cm table "health" checks.
 
 """
 
@@ -20,6 +20,15 @@ class RevisionError(Exception):
         super(RevisionError, self).__init__(message)
 
 
+def check_for_overlap(start_i, stop_i, start_j, stop_j):
+    if start_j <= start_i:
+        if stop_j > start_i:
+            return True
+    elif start_j <= stop_i:
+            return True
+    return False
+
+
 class Connections:
     def __init__(self, session=None):
         if session is None:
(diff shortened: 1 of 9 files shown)


hera_mc/cm_health.py
```diff
@@ -0,0 +1,133 @@
...
+        if session is None:
+            db = mc.connect_to_mc_db(None)
+            self.session = db.sessionmaker()
+        else:
+            self.session = session
+        self.conndict = None
+
+    def get_conndict(self, session=None):
+        self.conndict = {}
+        self.multiples = []
```
pkgw: This should be a `set`, rather than a list, to make the `not in` check below more efficient and more precisely capture what this data structure is doing.

hera_mc/cm_health.py
```diff
@@ -0,0 +1,133 @@
...
+            connection = [conn.upstream_part, conn.up_part_rev, conn.upstream_output_port,
+                          conn.downstream_part, conn.down_part_rev, conn.downstream_input_port]
+            connection = [x.lower() for x in connection]
+            k = ':'.join(connection)
+            if k in self.conndict.keys():
+                self.conndict[k].append([conn.start_gpstime, conn.stop_gpstime])
+                if k not in self.multiples:
+                    self.multiples.append(k)
+            else:
+                self.conndict[k] = [[conn.start_gpstime, conn.stop_gpstime]]
```
pkgw: If `multiples` is turned into a set, lines 42-47 can be condensed down to:

```
self.multiples.add(k)
self.conndict.setdefault(k, []).append([conn.start_gpstime, conn.stop_gpstime])
```

hera_mc/cm_health.py
```diff
@@ -0,0 +1,133 @@
...
+
+    def check_for_duplicate_connections(self):
+        if self.conndict is None:
+            self.get_conndict()
```
pkgw: It's a better pattern to have `get_conndict` contain this sort of state-checking logic:

```
def get_conndict(self):
    if self.conndict is not None:
        return

    self.conndict = {}
    ....
```

That way callers don't have to remember the optimization/validation logic. They can just call `get_conndict()` (I like to name these sorts of functions `ensure_conndict()`) and the right thing will happen if needed.

hera_mc/cm_health.py
```diff
@@ -0,0 +1,133 @@
...
+                    start_i = cm_utils.get_astropytime(self.conndict[k][i][0])
+                    stop_i = cm_utils.get_stopdate(self.conndict[k][i][1])
+                    start_j = cm_utils.get_astropytime(self.conndict[k][j][0])
+                    stop_j = cm_utils.get_stopdate(self.conndict[k][j][1])
+                    print("CM_H:  ", k)
+                    if start_j <= start_i:
+                        if stop_j > start_i:
+                            duplicates.append([k, i, j])
+                    else:
+                        if start_j <= stop_i:
```
pkgw: These two lines can be condensed into `elif start_j <= stop_i:`.

hera_mc/cm_health.py
```diff
@@ -0,0 +1,133 @@
...
+                self.conndict[k] = [[conn.start_gpstime, conn.stop_gpstime]]
+
+    def check_for_duplicate_connections(self):
+        if self.conndict is None:
+            self.get_conndict()
+        if len(self.multiples) == 0:
+            print("No duplications found.")
+            return 0
+        duplicates = []
+        for k in self.multiples:
```
pkgw: You could write `for k, timeranges in self.multiples.items():` and then replace all references to `self.conndict[k]` below with `timeranges`. It's more efficient and a bit cleaner in my opinion.

hera_mc/cm_health.py
```diff
@@ -0,0 +1,133 @@
...
+                    print("CM_H:  ", k)
+                    if start_j <= start_i:
+                        if stop_j > start_i:
+                            duplicates.append([k, i, j])
+                    else:
+                        if start_j <= stop_i:
+                            duplicates.append([k, i, j])
+        if len(duplicates):
+            print('{} duplications found.'.format(len(duplicates)))
+            for d in duplicates:
```
pkgw: This loop is incorrect because it is reusing the `i` and `j` variables as they existed at the end of the previous loop. You want this to be `for k, i, j in duplicates:` and then delete the `k = d[0]` line below.

hera_mc/cm_health.py
```diff
@@ -0,0 +1,133 @@
...
+        print('Out of {} connections checked.'.format(self.num_connections))
+        return len(duplicates)
+
+    def check_for_existing_connection(self, connection, at_date='now'):
+        at_date = cm_utils.get_astropytime(at_date)
+        connection = [x.lower() for x in connection]
+        k = ':'.join(connection)
+        if self.conndict is None:
+            self.get_conndict()
+        if k not in self.conndict.keys():
```
pkgw: This can just be `if k not in self.conndict:`. Or, alternately,

```
timeranges = self.conndict.get(k)
if timeranges is None:
    return False

for t in timeranges:
    ....
```

hera_mc/cm_health.py
```diff
@@ -0,0 +1,133 @@
...
+        self.multiples = []
+        self.num_connections = 0
+        for conn in self.session.query(PC.Connections).all():
+            self.num_connections += 1
+            connection = [conn.upstream_part, conn.up_part_rev, conn.upstream_output_port,
+                          conn.downstream_part, conn.down_part_rev, conn.downstream_input_port]
+            connection = [x.lower() for x in connection]
+            k = ':'.join(connection)
+            if k in self.conndict.keys():
+                self.conndict[k].append([conn.start_gpstime, conn.stop_gpstime])
```
pkgw: Should these GPS times be converted to Astropy times at this point? It looks like most if not all of the code below converts the times to Astropy times whenever they're used, so it might be better to do the conversions up-front and not have to worry about them later.

hera_mc/cm_health.py
```diff
@@ -0,0 +1,133 @@
...
+        k = ':'.join(connection)
+        if self.conndict is None:
+            self.get_conndict()
+        if k not in self.conndict.keys():
+            return False
+        for t in self.conndict[k]:
+            start_t = cm_utils.get_astropytime(t[0])
+            stop_t = cm_utils.get_astropytime(t[1])
+            if cm_utils.is_active(at_date, start_t, stop_t):
+                print('Duplicate found for connection {} at {}:  {} - {}'.format(k, at_date, start_t, stop_t))
```
pkgw: If this function returns True, it doesn't necessarily indicate a *duplicate* connection, right? I.e., should this printed message be changed?

hera_mc/cm_health.py
```diff
@@ -0,0 +1,133 @@
...
+    ------------
+    hpn:  hera part name
+    """
+
+    overlap = []
+    revisions = cm_revisions.get_all_revisions(hpn, session)
+    for i, r1 in enumerate(revisions):
+        for j, r2 in enumerate(revisions):
+            if i >= j:
+                continue
```
pkgw: This can be simplified by writing `for j, r2 in enumerate(revisions[:i]):`, I think. But also I don't think the inner index is needed at all; see below.

hera_mc/cm_health.py
```diff
@@ -0,0 +1,133 @@
...
+            first_one = i if r1.started <= r2.started else j
+            second_one = i if r1.started > r2.started else j
```
pkgw: This could be tightened up:

```python
first_one, second_one = (r1, r2) if r1.started <= r2.started else (r2, r1)

if first_one.ended is None:
   first_one.ended = cm_utils.get_astropytime('>')

...
```

hera_mc/cm_health.py
```diff
@@ -0,0 +1,133 @@
...
+            if revisions[first_one].ended is None:
+                revisions[first_one].ended = cm_utils.get_astropytime('>')
+            if revisions[second_one].started > revisions[first_one].ended:
```
pkgw: Is this test correct? I'd think that the normal expectation would be that the later revision starts after the first revision ends.

hera_mc/cm_health.py
```diff
@@ -0,0 +1,133 @@
+# -*- mode: python; coding: utf-8 -*-
+# Copyright 2018 the HERA Collaboration
+# Licensed under the 2-clause BSD license.
+
+"""Functions to handle various cm table health checks.
```
pkgw: For terminology, I guess the idea of a "health check" implies to me that this code is going to query the various bits of hardware in the running system and check that they're all operating correctly. What the code actually does I might describe more as a "consistency check" or something. But it would be annoying to switch the terminology everywhere and I don't think it's worth doing unless you were already dissatisfied with it.

david-deboer:
I made most as recommended, however pulled the overlapping thing out as a separate function to clean up where it is used (in various guises).  The set multiple needs one more check than suggested for the `add` since it requires 2, not 1 time range in conndict. 

Thanks for catching the error in the `duplicate printout!

david-deboer:
It did go up 0.4%!  (Mainly because it made me...).  It will keep going up (slowly), but I'd like this to go and not have PRs sitting around.

bhazelton:
Ok. I don't want to stop progress, but I'd really like for testing coverage to sometime rise up in the priority list. I see writing new code (and a whole new module) as an opportunity to at least completely cover the new code.

david-deboer:
I did do that!

hera_mc/cm_health.py
```diff
@@ -0,0 +1,156 @@
...
+    """
+
+    overlap = []
+    revisions = cm_revisions.get_all_revisions(hpn, session)
+    for i in range(len(revisions)):
+        for j in range(i):
+            interval_i = [revisions[i].started, cm_utils.get_stopdate(revisions[i].ended)]
+            interval_j = [revisions[j].started, cm_utils.get_stopdate(revisions[j].ended)]
+            if check_for_overlap(interval_i, interval_j):
+                self.overlap.append([revisions[i], revisions[j]])
```
pkgw: This reference to `self` is wrong since this is a freestanding function.

PR #212, opened 2018-02-15
Hookupinfo

david-deboer:
This includes antenna counts as a hookup flag, to see how many antennas are there etc.  It also fixes a bad fail when a part number isn't supplied for parts.py

### DIFF
--- a/hera_mc/cm_hookup.py
+++ b/hera_mc/cm_hookup.py
@@ -558,9 +558,14 @@ class Hookup:
             s += 'Cache file_mod_time:  {}\n'.format(cm_utils.get_time_for_display(file_mod_time))
             s += 'Cached hookup list:  {}\n'.format(self.cached_hookup_list)
             s += 'Cached hookup has {} keys.\n'.format(len(self.cached_hookup_dict['hookup'].keys()))
+            hooked_up = 0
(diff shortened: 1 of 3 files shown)


hera_mc/cm_hookup.py
```diff
@@ -558,9 +558,14 @@ def hookup_cache_file_info(self):
             s += 'Cache file_mod_time:  {}\n'.format(cm_utils.get_time_for_display(file_mod_time))
             s += 'Cached hookup list:  {}\n'.format(self.cached_hookup_list)
             s += 'Cached hookup has {} keys.\n'.format(len(self.cached_hookup_dict['hookup'].keys()))
+            hooked_up = 0
+            for k, p in self.cached_hookup_dict['fully_connected'].iteritems():
+                if p['e'] is True or p['n'] is True:
```
pkgw: Should this just be `if p['e] or p['n']:`? This might not be the case if `p['e']` might be a non-boolean value; if so, there should be a comment explaining what's going on.

david-deboer: This key is only supposed to have boolean values (it is a shortcut so that the code doesn't have to sort through the hookup more than once after loaded).  I've made the change, since it is equivalent and if people want to convey other information there in the future it should still work.

PR #214, opened 2018-02-19
Tridrants

david-deboer:
This implements the hera "tridrant" sector stuff and cleans up the geo*.py code a bit.

### DIFF
--- a/hera_mc/geo_handling.py
+++ b/hera_mc/geo_handling.py
@@ -107,8 +107,8 @@ class Handling:
         Returns located cofa.
         """
         self.get_station_types(add_stations=True)
-        current_cofa = self.station_types['COFA']['Stations']
-        located = self.get_location(current_cofa, 'now', self.station_types)
+        current_cofa = self.station_types['cofa']['Stations']
+        located = self.get_location(current_cofa, 'now')
         if len(located) > 1:
             s = "{} has multiple cofa values.".format(str(current_cofa))
             warnings.warn(s)
--- a/hera_mc/geo_handling.py
+++ b/hera_mc/geo_handling.py
@@ -22,14 +22,6 @@ from pyproj import Proj
 from hera_mc import mc, part_connect, cm_utils, geo_location
 
 
-class FauxPlot():
-    def figure(x):
-        return
-
-    def plot(x, y, **kwargs):
-        return
-
-
 def cofa(session=None):
     """
     Returns location class of current COFA
@@ -98,10 +90,9 @@ class Handling:
             self.session = db.sessionmaker()
         else:
             self.session = session
-        if testing:
-            self.plt = FauxPlot()
-        else:
-            import matplotlib.pyplot as self.plt
+        self.testing = testing
+        if not testing:
+            import matplotlib.pyplot as plt
 
         self.station_types = None
         self.__stations_added_to_types = False
@@ -146,9 +137,9 @@ class Handling:
         station_data = self.session.query(geo_location.StationType).all()
         stations = {}
         for sta in station_data:
-            stations[sta.station_type_name] = {'Prefix': sta.prefix,
-                                               'Description': sta.description,
-                                               'Marker': sta.plot_marker, 'Stations': []}
+            stations[sta.station_type_name.lower()] = {'Prefix': sta.prefix.upper(),
+                                                       'Description': sta.description,
+                                                       'Marker': sta.plot_marker, 'Stations': []}
         if add_stations:
             self.__station_added_to_types = True
             locations = self.session.query(geo_location.GeoLocation).all()
@@ -202,9 +193,8 @@ class Handling:
         """
 
         query_date = cm_utils.get_astropytime(query_date)
-        ustn = station.upper()
         connected_antenna = self.session.query(part_connect.Connections).filter(
-            (func.upper(part_connect.Connections.upstream_part) == ustn) &
+            (func.upper(part_connect.Connections.upstream_part) == station.upper()) &
             (query_date.gps >= part_connect.Connections.start_gpstime))
         ctr = 0
         for conn in connected_antenna:
@@ -247,7 +237,7 @@ class Handling:
             raise ValueError('More than one active connection between station and antenna')
         return antenna_connected.upstream_part
 
-    def get_location(self, to_find_list, query_date, station_types=None):
+    def get_location(self, to_find_list, query_date):
         """
         Return the location of station_name or antenna_number as contained in to_find.
         This accepts the fact that antennas are sort of stations, even though they are parts
@@ -256,12 +246,9 @@ class Handling:
         ------------
         to_find_list:  station/antenna names to find (must be a list)
         query_date:  astropy Time for contemporary antenna
-        station_types: list of station_type prefixes (e.g. HH)
-        show_location:   if True, it will print the information
-        verbosity:  sets the verbosity of the print
         """
-        self.get_station_types(add_stations=True)
-        found_location = []
+        self.get_station_types(add_stations=False)
+        locations = []
         for L in to_find_list:
             station_name = False
             try:
@@ -276,8 +263,8 @@ class Handling:
                     a.desc = self.station_types[a.station_type_name]['Description']
                     hera_proj = Proj(proj='utm', zone=a.tile, ellps=a.datum, south=True)
                     a.lon, a.lat = hera_proj(a.easting, a.northing, inverse=True)
-                    found_location.append(copy.copy(a))
-        return found_location
+                    locations.append(copy.copy(a))
+        return locations
 
     def print_loc_info(self, loc_list, verbosity='h'):
         """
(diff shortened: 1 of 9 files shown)


hera_mc/geo_handling.py
```diff
@@ -117,8 +118,7 @@ def cofa(self):
     def get_station_types(self, add_stations=True):
         """
         adds a dictionary of sub-arrays (station_types) to the class
-             [prefix]{'Description':'...', 'plot_marker':'...', 'stations':[]}
-        also adds the "flipped" dictionary
+             [station_type_name]{'Prefix', 'Description':'...', 'plot_marker':'...', 'stations':[]}
 
         return dictionary with station information
```
pkgw: This function no longer returns anything.

hera_mc/geo_handling.py
```diff
@@ -127,23 +127,27 @@ def get_station_types(self, add_stations=True):
...
                        if False, just return station types
         """
-
+        if self.__stations_added_to_types:
+            return
+        if not add_stations and self.station_types is not None:
+            return
         station_data = self.session.query(geo_location.StationType).all()
         stations = {}
         for sta in station_data:
```
pkgw: This isn't a change in this PR, but it is better to write the loop as:

```python
for sta in self.session.query(geo_location.StationType):
```

While the `.all()` call buffers the results of the database query into one big list, iterating over the non-`.all()` query streams through the database results as they come back, which is more efficient. Not that the difference is large in this particular case. And in my opinion, avoiding the extra local variable here makes things somewhat clearer.

hera_mc/geo_handling.py
```diff
@@ -127,23 +127,27 @@ def get_station_types(self, add_stations=True):
         add_stations:  if True, add all of the stations to their types
                        if False, just return station types
         """
-
+        if self.__stations_added_to_types:
+            return
+        if not add_stations and self.station_types is not None:
+            return
         station_data = self.session.query(geo_location.StationType).all()
         stations = {}
```
pkgw: It might be clearer to rename this local variable to `station_types`.

hera_mc/geo_handling.py
```diff
@@ -127,23 +127,27 @@ def get_station_types(self, add_stations=True):
         add_stations:  if True, add all of the stations to their types
                        if False, just return station types
```
pkgw: Again, this function no longer returns anything.

Also, it might be helpful to add an example of when one might want to specify `add_stations = False`. It would simplify this function and some of its callers if the stations were always loaded up (i.e. if the `add_stations` argument were removed).

hera_mc/geo_handling.py
```diff
@@ -117,8 +118,7 @@ def cofa(self):
     def get_station_types(self, add_stations=True):
```
pkgw: As in a recent PR, you might consider renaming this to `ensure_station_types()`.

hera_mc/geo_handling.py
```diff
@@ -286,18 +272,32 @@ def print_loc_info(self, loc_list, verbosity='h'):
...
         return True
 
+    def parse_station_types_to_check(self, sttc, add_stations):
+        self.get_station_types(add_stations=add_stations)
+        if isinstance(sttc, (str, unicode)):
+            if sttc.lower() == 'all':
+                return self.station_types.keys()
+            else:
+                sttc = [sttc]
+        sttypes = []
```
pkgw: It might be better to make this a `set` rather than a list, in case the passed-in specification works out to include redundant station types.

hera_mc/geo_handling.py
```diff
@@ -286,18 +272,32 @@ def print_loc_info(self, loc_list, verbosity='h'):
...
+        for s in sttc:
+            if s.lower() in self.station_types.keys():
```
pkgw: This can just be `if s.lower() in self.station_types:`.

hera_mc/geo_handling.py
```diff
@@ -286,18 +272,32 @@ def print_loc_info(self, loc_list, verbosity='h'):
...
+                sttypes.append(s.lower())
+            else:
+                for k in self.station_types.keys():
+                    if s.upper() in self.station_types[k]['Prefix']:
+                        sttypes.append(k)
```
pkgw: This loop could be:

```python
for k, st in self.station_types.iteritems():
    if s.upper() in st['Prefix']:
        sttypes.append(k) # or sttypes.add(k) if sttypes becomes a set
```

Also, is this `s.upper() in self.station_types[k]['Prefix']` the desired behavior? I.e, I can pass an argument of `'H'` and it will match all station types whose prefixes include the letter `H`? A few example arguments in the docstring might be instructive.

hera_mc/geo_handling.py
```diff
@@ -308,75 +308,71 @@ def get_ants_installed_since(self, query_date, station_types_to_check='all'):
...
 
-        self.get_station_types(add_stations=False)
+        station_types_to_check = self.parse_station_types_to_check(station_types_to_check, add_stations=False)
         dt = query_date.gps
         found_stations = []
         for a in self.session.query(geo_location.GeoLocation).filter(
                 geo_location.GeoLocation.created_gpstime >= dt):
-            if (station_types_to_check == 'all' or
-                    self.flipped_station_types[a.station_type_name] in station_types_to_check):
+            if a.station_type_name.lower() in station_types_to_check:
```
pkgw: Note that if `parse_station_types_to_check` returned a set, this sort of check would be quicker. (Although we're talking about lists with a handful of items so it's really more about using the best tool for the job, rather than actual real-world performance.)

hera_mc/geo_handling.py
```diff
@@ -308,75 +308,71 @@ def get_ants_installed_since(self, query_date, station_types_to_check='all'):
...
                 found_stations.append(a.station_name)
         return found_stations
 
-    def plot_stations(self, stations_to_plot_list, query_date, state_args, testing=False):
+    def get_antenna_label(self, label_to_show, stn, query_date):
+        labeling = None
+        if label_to_show == 'name':
+            labeling = stn.station_name
```
pkgw: This could be:

```python
if label_to_show == 'name':
   return stn.station_name
```

And then the rest of the function could lose a level of indentation. It's generally a good thing to arrange control flow to minimize the levels of indentation (<=> levels of nested logic) that are needed. In general, in this function the `labeling` variable could be ditched in favor of just returning a value.

hera_mc/sys_handling.py
```diff
@@ -77,7 +77,7 @@ def get_dubitable_list(self, date='now', return_full=False):
 
         raise ValueError('Too many open dubitable lists ({}).'.format(len(fnd)))
 
-    def get_all_fully_connected_at_date(self, at_date, station_types_to_check=['HH']):
+    def get_all_fully_connected_at_date(self, at_date, station_types_to_check=['HH', 'HA', 'HB']):
```
pkgw: I see three or four places in the overall `hera_mc` codebase that include a hardcoded list of station types like this. If/when we add more station types to consider, we might get bugs if not all of the lists are updated. Would it be better to make this list a global variable somewhere? (One can write, e.g., `def my_func(station_types=cm_utils.default_station_types)`, if the `cm_utils` has been imported in the module.)

scripts/add_station.py
```diff
@@ -16,6 +16,24 @@
 
 default_coord_file_name = os.path.join(mc.data_path, 'HERA_350.txt')
 
+region = {'herahexw': [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20, 21,
```
pkgw: These lists could be condensed a lot through judicious use of the `range()` function. Since (in Python 2) `range()` returns lists, you can concatenate them like `range(0, 47) + range(50, 61) + ...`.

scripts/geo.py
```diff
@@ -12,171 +12,75 @@
...
+    show_state = args.show_state.lower()
+    if args.action == 'sin':
+        cutoff = at_date
+        at_date = cm_utils.get_astropytime('now')
 
     # start session and instances
     db = mc.connect_to_mc_db(args)
     session = db.sessionmaker()
-    h = geo_handling.Handling(session)
+    G = geo_handling.Handling(session)
```
pkgw: Single-letter local variable names should not be capitalized.

hera_mc/geo_handling.py
```diff
@@ -16,7 +16,7 @@
 import warnings
 from astropy.time import Time, TimeDelta
 from sqlalchemy import func
-
+import matplotlib.pyplot as plt
```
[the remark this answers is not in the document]
david-deboer: I don't understand this particular comment.

hera_mc/geo_handling.py
david-deboer: Did this, and then the very next loop was same, so changed too.

hera_mc/geo_handling.py
david-deboer: Thought about this from a performance point of view, then decided to do it since minimal hit and, as pointed out, simpler.

hera_mc/sys_handling.py
david-deboer: Why does this need to be `my_func?  I did a version of this.

scripts/add_station.py
david-deboer: Yah, but the work already done...

scripts/geo.py
david-deboer: Small push-back -- when setting up a local class using cap is allowed (well, obviously that is not said anywhere, except in this comment but...)

hera_mc/geo_handling.py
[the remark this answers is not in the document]
pkgw: Before, the pieces of code that imported matplotlib were all located inside the plotting functions, and so they were never triggered unless you tried to use those functions. That meant that you could `import hera_mc.geo_handling` even if you didn't have matplotlib installed. With the matplotlib import moved to the top level, you won't be able to import this module at all if you don't have matplotlib installed. Matplotlib is a pretty heavyweight dependency, and it could be genuinely difficult to install on something like a Raspberry Pi — i.e. the sort of mini-machine on which we quite plausibly will be using `hera_mc`. So I think it is better to maintain the old pattern and isolate all of the matplotlib imports to the functions that use them.

scripts/geo.py
pkgw: It's one of those things that just Is Not Done in Python ([cf PEP8](https://www.python.org/dev/peps/pep-0008/#function-and-variable-names)).

PR #220, opened 2018-03-14
fixed show hookup with none found issue

david-deboer:
Otherwise just showed a dissatisfying error.


hera_mc/cm_hookup.py
```diff
@@ -587,14 +587,19 @@ def show_hookup(self, hookup_dict, cols_to_show='all', show_levels=False, show_p
         headers = self.__make_header_row(hookup_dict['columns'], cols_to_show, show_levels)
         table_data = []
         numerical_keys = cm_utils.put_keys_in_numerical_order(sorted(hookup_dict['hookup'].keys()))
+        total_shown = 0
         for hukey in numerical_keys:
             for pol in sorted(hookup_dict['hookup'][hukey].keys()):
                 use_this_row = False
+                if hukey not in hookup_dict['hookup'].keys():
+                    continue
```
pkgw: I don't think this `if` statement can ever be true. The `for` loop iterates through `hookup_dict['hookup'][hukey]`, so if `hukey` is `not in hookup_dict['hookup'].keys()` you'll get an exception instead of entering the loop.

Also, again, it is better and more efficient to write `key in some_dict` rather than `key in some_dict.keys()`. The former tests for membership using the dictionary's efficient hashing method, whereas if you explicitly extract the `keys()` list, it has to generate that last and then search it linearly, which take `O(n)` time.

PR #224, opened 2018-05-20
Hera sigpath

david-deboer:
This makes the changes needed to incorporate the new architecture.


hera_mc/cm_handling.py
```diff
@@ -358,144 +330,73 @@ def get_connection_dossier(self, hpn_list, rev, port, at_date, exact_match=False
...
+                    pkey_tmp = {}
+                    for i, ck in enumerate(connection_dossier['conn'][prkey][sk]):
+                        po = getattr(connection_dossier['conn'][prkey][sk][ck], port_type[sk])
+                        pkey_tmp[po + '{:03d}'.format(i)] = ck
+                    connection_dossier['conn'][prkey]['paired'][sk] = [pkey_tmp[x] for x in sorted(pkey_tmp.keys())]
+                pad = len(connection_dossier['conn'][prkey]['paired']['down']) - len(connection_dossier['conn'][prkey]['paired']['up'])
+                if pad < 0:
+                    connection_dossier['conn'][prkey]['paired']['down'].extend([None] * abs(pad))
+                elif pad > 0:
+                    connection_dossier['conn'][prkey]['paired']['up'].extend([None] * abs(pad))
```
pkgw: I'm having some trouble understanding the logic of this stanza and convincing myself that the sort algorithm will pair all of the connections correctly. Some explanatory comments might help.

Also, in the function's docstring, I think it would be very helpful to paste an example dictionary that you might get out of this function:

```python
""".... An example return value might look like:

{
  'up': {
    'blah': <Connection object>,
    'other: <Connection object>,
  },
  'down': {
     'yadda': ...
  },
  'paired': {
     'up': [
        'key000',
        ...
      ],
      'down': [
         None,
         ....,
      ]
   }
}
"""
```

Also also, these dossier dictionaries are getting big and complicated enough that they should really become classes. That way their fields can be documented and they can be given methods for manipulating their contents. But that is not a change to undertake right now.

hera_mc/cm_hookup.py
```diff
@@ -38,14 +38,12 @@ def get_parts_from_hookup(part_name, hookup_dict):
     if part_name not in hookup_dict['columns']:
         return None
     parts = {}
-    part_ind = hookup_dict['columns'].index(part_name)
+    part_ind = hookup_dict['columns'].index(part_name) - 1 if hookup_dict['columns'].index(part_name) > 1 else 0
```
pkgw: Is this test correct? Right now we have that if `hookup_dict['columns'].index(part_name)` is 0 or 1, `part_ind = 0`. Otherwise, `part_ind = hookup_dict['columns'].index(part_name) - 1`. I don't understand the structure of the `columns` list and how that relates to the things in the `hookup` sub-structure, so maybe this is appropriate. If so some explanatory comments would help.

hera_mc/cm_hookup.py
```diff
@@ -270,54 +268,42 @@ def __get_part_types_found(self, huco, part_types_found):
...
         """
-        Given the current part and port_query (which is either 'all', 'e', or 'n')
-        this figures out which pols to do.  Basically, given 'all' and part it
+        - 1 - Signal Path
+        Given the current part and port_query (which is either 'pol' (or 'all'), 'e', or 'n')
+        this figures out which pols to do.  Basically, given 'pol' and part it
         figures out whether to return ['e'], ['n'], ['e', 'n']
 
+        - 2 - Physical
+        If the port_query is '@' or '[phys]ical' it returns all physical ports.  Doesn't fully work.
```
pkgw: I do not like this magical `@` sigil at all because the non-expert cannot guess its meaning. Can we replace it with something short but meaningful?

Also, can the comment "doesn't fully work" be expanded to indicate what aspect doesn't work and why not?

hera_mc/cm_handling.py
```diff
@@ -145,46 +145,34 @@ def get_part_dossier(self, hpn_list, rev, at_date, exact_match=False, full_versi
...
-                                [part.hpn], at_date, session=self.session)
-                        part_dossier[pr_key]['input_ports'], part_dossier[pr_key]['output_ports'] = \
-                            self.find_ports(part_dossier[pr_key]['connections'])
-                else:
-                    msg = "Should only be one part/rev for {}:{}.".format(part.hpn, part.hpn_rev)
-                    warnings.warn(msg)
+                part = copy.copy(part_query.first())  # There should be only one.
+                part.gps2Time()
+                pr_key = cm_utils.make_part_key(part.hpn, part.hpn_rev)
+                part_dossier[pr_key] = {'Time': at_date, 'part': part}
```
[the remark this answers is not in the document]
david-deboer: This was capitalized since it corresponded to the astropy Time.  I'm reluctant to change it now, since it might have propagated...

hera_mc/cm_hookup.py
david-deboer: It is but is confusing -- it matches the calls in the cm_info stuff.  I'd like to change both of these to be less confusing and more robust.  I'll make an issue.

scripts/add_station.py
```diff
@@ -35,6 +35,32 @@
...
 
+class Nodes:
+    node = {}
+    E = []
+    N = []
+
+    def __init__(self):
+        print('Nodes')
+
+    def import_nodes(self, file_name=None):
```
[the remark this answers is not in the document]
david-deboer: I completely redid add_station to get rid of historical baggage.

hera_mc/cm_hookup.py
david-deboer: I got rid of this as a hookup option (i.e. lessened my ambitions - to see these connections use parts.py conn.  Maybe I'll re-add them to hookup, but likely not.)

hera_mc/cm_handling.py
david-deboer: I COMPLETELY agree about the Class issue.  Tempting to do now, but better for another PR, which I will likely start.

hera_mc/cm_hookup.py
```diff
@@ -270,54 +268,42 @@ def __get_part_types_found(self, huco, part_types_found):
...
-            return p == pol
-        else:
-            raise ValueError("Invalid polarization.")
+        if port_query in PC.both_pols:
+            return [port_query]
+        raise ValueError('Invalid port_query')
 
     def __follow_hookup_stream(self, part, rev, pol):
         self.upstream = []
         self.downstream = []
```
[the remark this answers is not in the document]
david-deboer: We can look at this in a future PR.  If you think necessary, please raise in issue.

scripts/add_station.py
```diff
@@ -35,45 +31,46 @@
           'heraringb': [320, 321, 322, 323, 324, 328, 329, 332, 333, 336, 337, 340, 341, 345, 346, 347, 348, 349]}
 
 
-def get_coord_from_file(station_name, coord_filename):
-    coords = None
-    if station_name is not None:
-        fp = open(coord_filename, 'r')
+def Nodes():
```
pkgw: I kind of liked the simple classes here (à la issue #226) but this code is simple enough that it's not a big deal if you prefer a dict-based design. But, having a callable thing with a capitalized name like `Nodes()` that is a function rather than a class violates a virtually universal Python naming convention. This should be renamed to something like `read_nodes()` (and below, `Antennas` → `read_antennas()`).

hera_mc/cm_handling.py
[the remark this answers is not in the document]
pkgw: This is what `git grep` is for! It's an important tool for this kind of refactoring.

```
hera_mc $ git grep \'Time
hera_mc/cm_handling.py:                              {'Time':Time, 'part':CLASS,
hera_mc/cm_handling.py:                part_dossier[pr_key] = {'Time': at_date, 'part': part}
hera_mc/cm_handling.py:          'Time': astropytime,
hera_mc/cm_handling.py:        connection_dossier = {'Time': at_date, 'conn': {}}
hera_mc/cm_utils.py:                    raise ValueError('Time can only be hours[:minutes[:seconds]], not {}.'.format(_time))
scripts/mc_megaraid_monitor.py:        # it reports event times in the local time using the 'Time'
scripts/mc_megaraid_monitor.py:        abs_time_str = data.pop('Time', None)
```

Issue #229, opened 2018-06-27
python 3 compatibility


bhazelton:

pkgw:
At the moment I think we're only blocking on `hera_qm` becoming Python-3-compatible, at least in terms of the test suite passing. It might be good to disable that test for the time being and start CI'ing Python 3 anyway, so that we don't accidentally re-introduce Python-2-only-isms.

bhazelton:
That sounds reasonable, I'll make a PR

PR #231, opened 2018-07-27
make code Python3 compatible

bhazelton:
Note that python 3 testing isn't running in Travis yet because hera_qm isn't yet python 3 compatible. But that's the only error I find running nosetests in python 3 locally.

partially addresses #229 
I'm not closing the issue with this PR, that should wait until we have python 3 running on Travis.


bhazelton:
I'm a little bit worried about the scripts. I made the changes I thought needed to be made for python 3 compatibility, but I'm not sure how to tell that they all work. They aren't covered by tests.

pkgw:
Busy with a conference this week, but will look at this next week if that timescale is ok.

bhazelton:
Yes, that's fine. Thanks!

hera_mc/geo_handling.py
```diff
@@ -334,7 +335,7 @@ def plot_stations(self, stations_to_plot_list, query_date, **kwargs):
         query_date:  date to use to check if active
         kwargs:  arguments for marker_color, marker_shape, marker_size, show_label, xgraph, ygraph
         """
-        import matplotlib.pyplot as plt
+        # import matplotlib.pyplot as plt
```
pkgw: Why was this commented out? By my reading it breaks the code if `self.testing` is False.

hera_mc/geo_handling.py
```diff
@@ -371,7 +372,7 @@ def plot_station_types(self, query_date, station_types_to_use, **kwargs):
         kwargs:  marker_color, marker_shape, marker_size, show_label, xgraph, ygraph
         """
         from hera_mc import cm_hookup, cm_revisions
-        import matplotlib.pyplot as plt
+        # import matplotlib.pyplot as plt
```
pkgw: Ditto.

hera_mc/geo_handling.py
bhazelton: oops, that's a mistake, I took it out because of a weird thing I was seeing but I thought I'd put it back in. Sorry, I'll fix it.

pkgw:
OK, I've done some of the tweaks on things I noticed and test the scripts as best I can. Once the matplotlib thing is changed and CI passes I think this should be merged.

hera_mc/geo_handling.py
pkgw: I've got it all pulled up right now so I went ahead and made that change myself.

bhazelton: thanks!

bhazelton:
Thanks for the fixes @pkgw! I learned a lot reading through them.

PR #233, opened 2018-08-09
Add support for node m&c info

bhazelton:


hera_mc/mc_session.py
```diff
@@ -1040,6 +1040,195 @@ def get_roach_temperature(self, starttime, stoptime=None, roach=None):
...
+        special insertion method that will ignore records that are redundant
+        with ones already in the database. This makes it convenient to sample
+        the node sensor data densely on qmaster.
+
+        """
+        from .node import create_sensor
+
+        node_sensor_list = create_sensor()
+
+        if self.bind.dialect.name == 'postgresql':
```
pkgw: Perhaps this logic could be split into a generic method, if it's going to appear in multiple places?

bhazelton: yes it should!

pkgw:
Just checking, is this still in do-not-merge state?

bhazelton:
Yes, it's still not ready. Jack implemented some updates that I need to adjust this for and I need to test on site.

bhazelton:
@pkgw ok, I think this branch is ready now (once the tests pass)

hera_mc/mc_session.py
```diff
@@ -107,6 +107,36 @@ def _time_filter(self, table_object, time_column, starttime, stoptime=None,
 
         return result_list
 
+    def _special_insert(self, table_object, obj_list):
```
pkgw: This name is vague. Something like `_insert_ignoring_duplicates` would be better, I think.

Also, I think `table_object` is somewhat misnamed — if I'm reading correctly, this argument is a type. Something like `row_class`?

hera_mc/mc_session.py
```diff
@@ -1040,6 +1047,151 @@ def get_roach_temperature(self, starttime, stoptime=None, roach=None):
...
+        Add new node sensor data to the M&C database.
+
+        Parameters:
+        ------------
+        time: astropy time object
+            astropy time object based on a timestamp reported by node
+        node: integer
+            node number (integer running from 1 to 30)
+        top_sensor_temp: float
+            temperature of top sensor reported by node in Celcius
```
pkgw: Not a huge deal, but spelling: "celsius".

hera_mc/mc_session.py
```diff
@@ -1040,6 +1047,151 @@ def get_roach_temperature(self, starttime, stoptime=None, roach=None):
...
+        middle_sensor_temp: float
+            temperature of middle sensor reported by node in Celcius
+        bottom_sensor_temp: float
+            temperature of bottom sensor reported by node in Celcius
+        humidity_sensor_temp: float
+            temperature of the humidity sensor reported by node in Celcius
+        humidity: float
+            humidity measurement reported by node
```
pkgw: Units? Looks like this should be a relative humidity percentage?

hera_mc/mc_session.py
```diff
@@ -1040,6 +1047,151 @@ def get_roach_temperature(self, starttime, stoptime=None, roach=None):
                                  stoptime=stoptime, filter_column='roach',
                                  filter_value=roach)
 
+    def add_node_sensor(self, time, node, top_sensor_temp, middle_sensor_temp,
```
pkgw: I think the name of this table could get a bit confusing. A function named `add_node_sensor` sounds to me like we have a database that maintains a list of sensors that have been installed on nodes, and this function would be called when we've installed a new sensor.  `add_node_sensor_readings`? `add_node_telemetry`?

hera_mc/node.py
```diff
@@ -0,0 +1,259 @@
...
+    top_sensor_temp: temperature of top sensor reported by node in Celcius
+    middle_sensor_temp: temperature of middle sensor reported by node in Celcius
+    bottom_sensor_temp: temperature of bottom sensor reported by node in Celcius
+    humidity_sensor_temp: temperature of the humidity sensor reported by node in Celcius
+    humidity: percent humidity measurement reported by node
+    """
+    __tablename__ = 'node_sensor'
+    time = Column(BigInteger, primary_key=True)
+    node = Column(Integer, primary_key=True)
+    top_sensor_temp = Column(Float)
```
pkgw: Looking at the Alembic migration script, it looks like all of these fields will be nullable. Would it be better if they were non-nullable? I feel like it's generally good to start strict and loosen up if needed ... but on the other hand, I have trouble imagining realistic scenarios where nullability gets us in trouble. Maybe if we're getting bad data from the sensors, which end up getting represented as Nones, and then we don't notice that we're not taking any data for a while.

hera_mc/node.py
```diff
@@ -0,0 +1,259 @@
...
+        top_sensor_temp: float
+            temperature of top sensor reported by node in Celcius
+        middle_sensor_temp: float
+            temperature of middle sensor reported by node in Celcius
+        bottom_sensor_temp: float
+            temperature of bottom sensor reported by node in Celcius
+        humidity_sensor_temp: float
+            temperature of the humidity sensor reported by node in Celcius
+        humidity: float
+            humidity measurement reported by node
```
pkgw: Same comments as in the `mc.py` function.

scripts/mc_monitor_nodes.py
```diff
@@ -0,0 +1,71 @@
...
+                print('%s -- SQL error committing new node power status' % time.asctime(), file=sys.stderr)
+                traceback.print_exc(file=sys.stderr)
+                session.rollback()
+                continue
+            except Exception as e:
+                print('%s -- error committing new node power status' % time.asctime(), file=sys.stderr)
+                traceback.print_exc(file=sys.stderr)
+                continue
+    except KeyboardInterrupt:
+        pass
```
pkgw: This will mean that you can't exit the program by control-C'ing it, if you run it from the terminal. Is that the intended behavior? I feel like it is better to handle KeyboardInterrupt by exiting, but printing a one-line message "exiting on SIGINT" (e.g.) rather than a full traceback.

hera_mc/mc_session.py
bhazelton: `table_object` is referring to the Class that defines the table (inherits from DeclarativeBase). I'm happy to change the name, but I find the term `row` confusing. Maybe `table_class`?

hera_mc/mc_session.py
bhazelton: Thanks, that's embarrassing.

hera_mc/mc_session.py
bhazelton: yes, I changed it in node.py but forgot to change it here.

hera_mc/mc_session.py
bhazelton: yep, I'll change the function name to be clearer.

hera_mc/mc_session.py
bhazelton: `table_object` is also what I use for the same thing in the `_time_filter` function, so I think I should change it there as well.

hera_mc/node.py
bhazelton: So it sounds like other than the bottom temperature sensor, there will always be some value reported for all the sensors. But if the sensor goes bad or can't be reached the reported value will be `-99`. I'm trying to decide if it's more helpful to have `-99`s in this table representing bad readings or `None`s.

bhazelton: Ok, so these columns are going to stay nullable (per Slack conversation) because the sensors can give bad readings. Those readings will be reported to this code as `None`s. The columns in the `node_power_status` table will be made not nullable, however, because we don't expect to have missing readings for those.

bhazelton:
We should wait on the updates Jack is going to do to the node M&C code (which this code calls) and test against that updated code onsite before we merge this.

bhazelton:
Ok, Jack has made the changes and I've tested against them onsite. Everything seems to be working, so I think this PR is finally ready. @pkgw

bhazelton:
Thank you @pkgw!

PR #238, opened 2018-08-15
Update corr

david-deboer:
Updates the definition/return of the correlator input info to cminfo and updates `get_parts_from_type` (previously confusingly  called get_part_info).  Received Jack's ok...


hera_mc/part_connect.py
```diff
@@ -168,10 +169,9 @@ def update_part(session=None, data=None, add_new_part=False):
     for dkey, dval in six.iteritems(data_dict):
         hpn_to_change = dval[0][0]
         rev_to_change = dval[0][1]
-        # if rev_to_change[:4] == 'LAST':
+        # if rev_to_change.startswith('LAST'):
```
bhazelton: Why are we leaving in commented code?

hera_mc/part_connect.py
```diff
@@ -168,10 +169,9 @@ def update_part(session=None, data=None, add_new_part=False):
...
         #    rev_to_change = cm_revisions.get_last_revision(hpn_to_change, session)[0][0]
-        part_rec = session.query(Parts).filter((func.upper(Parts.hpn) == hpn_to_change.upper())
-                                               & (func.upper(Parts.hpn_rev) == rev_to_change.upper()))
+        part_rec = session.query(Parts).filter((func.upper(Parts.hpn) == hpn_to_change.upper()) & (func.upper(Parts.hpn_rev) == rev_to_change.upper()))
```
bhazelton: Is the only change here to make the line longer? Why?

hera_mc/part_connect.py
```diff
@@ -395,8 +395,7 @@ def add_part_info(session, hpn, rev, at_date, comment, library_file=None):
         db = mc.connect_to_mc_db(None)
         session = db.sessionmaker()
         close_session_when_done = True
-    part_rec = session.query(Parts).filter((func.upper(Parts.hpn) == hpn.upper())
-                                           & (func.upper(Parts.hpn_rev) == rev.upper()))
+    part_rec = session.query(Parts).filter((func.upper(Parts.hpn) == hpn.upper()) & (func.upper(Parts.hpn_rev) == rev.upper()))
```
bhazelton: ditto

hera_mc/sys_handling.py
```diff
@@ -282,19 +278,21 @@ def get_cminfo_correlator(self):
                 'cofa_lon': cofa_loc.lon,
                 'cofa_alt': cofa_loc.elevation}
 
-    def get_part_info(self, stn, at_date, part_name='post-amp'):
+    def get_part_from_type(self, stn, at_date, part_type='post-amp', include_ports=False):
```
bhazelton: Why do we have the same method name here and in cm_hookup? That seems a little confusing.

hera_mc/tests/test_sys_handling.py
```diff
@@ -63,7 +63,7 @@ def test_correlator_info(self):
 
         self.assertEqual(stn_types[index], 'herahexw')
 
-        self.assertEqual(corr_inputs[index], ('DF8B2', 'DF8B1'))
+        self.assertEqual(corr_inputs[index], ('input>DF8B2', 'input>DF8B1'))
```
bhazelton: This confuses me. Why do we want `'input>'` in the string?

david-deboer: For the new correlator we need the port and the part number, for the old correlator we only needed the part number and the old "port" was 'input'

hera_mc/cm_hookup.py
```diff
@@ -121,35 +121,60 @@ def add_correlator_levels(self, pol):
...
 
         returns:
-            parts[pol] = part number
-            location example: RI4A1E:A
-            pam number example: RCVR93:A (for a PAPER RCVR)
-            pam number example: PAM75101:B (for a HERA PAM)
+            parts:  dictionary keyed on polarization for actual installed part number of
+                    specified type within hookup as a string per pol
+                    if include_revs part number is e.g. FDV1:A
+                    if include_ports they are included as e.g. 'input<FDV:A>terminals'
```
bhazelton: I think the '<' and '>' are reversed in this docstring.

hera_mc/cm_hookup.py
```diff
@@ -121,35 +121,60 @@ def add_correlator_levels(self, pol):
         self.columns[pol].append('level')
         self.level[pol] = 'N/A'
 
-    def get_part_info(self, part_name):
+    def get_part_in_hookup_from_type(self, part_type, include_revs=False, include_ports=False):
         """
         Retrieve the value for a part name from a hookup
```
bhazelton: Should this docstring refer to "a part type" rather than "a part name"? It seems that you're distinguishing between types and names, but that doesn't seem to be consistent throughout...

hera_mc/tests/test_sys_handling.py
bhazelton: ok. I find the terminology a little confusing, but I can probably get used to it.

david-deboer:
Thanks - made the suggested changes.

PR #239, opened 2018-08-17
deprecate roach_temperature and paper_temperatures tables

bhazelton:


pkgw:
Shouldn't there be some associated changes to make the DB integrity checker ignore the associated DB tables if they exist?

bhazelton:
It appears that the database check only checks that the tables defined in the code are present with the right columns. It does not check that there are no extra tables. So if I leave the test alone and remove the roach_temperature table code, the tests pass.

Do we want to institute a check for extra tables with a white list of allowed extra tables or make no change?

pkgw:
In that case, I think it's fine as-is. I wouldn't be surprised if we chose the current implementation precisely to deal with this sort of circumstance.

PR #242, opened 2018-08-21
sorted connection printout

david-deboer:
all other printout lists except connections were sorted - now so are connections

### DIFF
--- a/hera_mc/cm_handling.py
+++ b/hera_mc/cm_handling.py
@@ -425,7 +425,8 @@ class Handling:
         elif verbosity == 'h':
             headers = ['uStart', 'uStop', 'Upstream', '<uOutput:', ':uInput>',
                        'Part', '<dOutput:', ':dInput>', 'Downstream', 'dStart', 'dStop']
-        for k, conn in six.iteritems(connection_dossier):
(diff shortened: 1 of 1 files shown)


david-deboer:
This is failing on getting server_status:

======================================================================
FAIL: test_add_server_status (hera_mc.tests.test_server_status.TestServerStatus)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/travis/build/HERA-Team/hera_mc/hera_mc/tests/test_server_status.py", line 82, in test_add_server_status
    self.assertTrue(result.isclose(expected))
AssertionError: False is not true
-------------------- >> begin captured stdout << ---------------------
column rtp_server_status.mc_time is string-like or int-like, values are not equal

bhazelton:
What's weird is it passes both the PR Travis test and the python 3 push test, but not the python 2 push test. And it passes on both python 2 & 3 on my local machine. I just restarted the failed job to see if it stays unhappy.

bhazelton:
Looks like re-running the test fixed the problem. I'm not sure what was wrong (and I'm a little disturbed by things that come and go mysteriously) but it's better now.

PR #246, opened 2018-08-29
Added cm_dataview to get antenna info over time.  Included julian date as option.

david-deboer:
This is a start to getting commissioning their antenna list.  It includes julian date in the cm_utils.get_astropytime function.

### DIFF
--- a/hera_mc/cm_dataview.py
+++ b/hera_mc/cm_dataview.py
@@ -4,8 +4,6 @@
 # Licensed under the 2-clause BSD license.
 
 """Allows some different data views of cm database.
-Currently this is only to look at the full connections of parts looped over time,
-either via db calls or pre-written files.
 """
 
 from __future__ import absolute_import, division, print_function
@@ -13,7 +11,7 @@ from __future__ import absolute_import, division, print_function
 from astropy.time import Time, TimeDelta
 import numpy as np
 
-from . import mc, cm_utils, cm_revisions, cm_hookup
+from . import mc, cm_utils, cm_revisions, sys_handling
 
 
 class Dataview:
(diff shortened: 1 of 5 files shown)


david-deboer:
Also added a delete_local_cache flag to hookup.py to easily delete the local hookup cache (which must be done after nosetests)

hera_mc/cm_dataview.py
```diff
@@ -27,149 +25,42 @@ def __init__(self, session=None):
...
-        Prints out correlator information from 'standard' fully_connected list
+        Return dated list of connected stations.
 
         Parameters:
         ------------
-        fully_connected:  list of fully connected stations provided by the correlator
-        methods in geo_handling
+        start:  start astropy.Time
+        stop:  stop astropy.Time
+        interval:  interval in days
```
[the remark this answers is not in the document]
david-deboer: Interval should be time_step -- I've changed documentation.

hera_mc/cm_dataview.py
```diff
@@ -27,149 +25,42 @@ def __init__(self, session=None):
...
+        filename:  filename to write to.  If stdout it will also return a dictionary
```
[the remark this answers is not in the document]
david-deboer: That was because the approach wasn't linear ;-)
I've changed it to be more sensible:  if run via script you always want a file and to see it (ok, maybe you don't want both but it is slow so always writing the file is prudent and it is a script to printing to the screen is probably harmless).  If called, it only returns the dictionary.

hera_mc/cm_dataview.py
```diff
@@ -26,41 +26,42 @@ def __init__(self, session=None):
...
         Parameters:
         ------------
-        start:  start astropy.Time
-        stop:  stop astropy.Time
-        interval:  interval in days
-        filename:  filename to write to.  If stdout it will also return a dictionary
-        station_types_to_check:  e.g. HH
+        start:  start (astropy.Time)
+        stop:  stop (astropy.Time)
+        time_step:  time_step in days between start/stop (float or int)
```
bhazelton: Sorry, this is still confusing to me. It looks like you want me to give you the number of days between the start & stop values (which you can just compute). Actually, I think it's the cadence at which you check which ants are connected. I do like the name `time_step` better than `interval`.

hera_mc/cm_dataview.py
```diff
@@ -27,149 +25,43 @@ def __init__(self, session=None):
...
-        Prints out correlator information from 'standard' fully_connected list
+        Return dated list of connected stations.
 
         Parameters:
         ------------
-        fully_connected:  list of fully connected stations provided by the correlator
-        methods in geo_handling
+        start:  start (astropy.Time)
+        stop:  stop (astropy.Time)
+        time_step:  time_step in days between start/stop (float or int)
```
bhazelton: Sorry, this is still confusing to me. It looks like you want me to give you the number of days between the start & stop values (which you can just compute). Actually, I think it's the cadence at which you check which ants are connected. I do like the name `time_step` better than `interval`.

bhazelton: I think the problem for me is the phrase `between start/stop'. If you clarify the docstrings on start and stop you can probably take that phrase out.

PR #248, opened 2018-08-31
changed raw_input to six.moves.input and slight update to install

david-deboer:
raw_input no longer works for Python 3.


scripts/cm_syslog.py
```diff
@@ -10,6 +10,7 @@
 from __future__ import absolute_import, division, print_function
 
 from hera_mc import mc, cm_utils, part_connect, sys_handling
+import sys
```
bhazelton: should this read `import six`?

PR #258, opened 2018-09-07
Fix Travis environment

bhazelton:
add h5py to Travis conda installation for qm code


plaplant:
Good, the h5py virus is spreading to more repos in the collaboration... :)

PR #263, opened 2018-09-14
hera_qm is now python 3 compatible, so no need to skip the tests

bhazelton:


hera_mc/tests/test_qm.py
```diff
@@ -183,11 +180,11 @@ def test_update_qm_list(self):
         r = self.test_session.get_metric_desc()
         results = []
         for result in r:
-            self.assertTrue(result.metric in self.metrics_dict.keys())
+            self.assertTrue(result.metric in list(self.metrics_dict.keys()))
```
pkgw: This could just be `self.assertTrue(result.metric in self.metrics_dict)`.

hera_mc/tests/test_qm.py
```diff
@@ -183,11 +180,11 @@ def test_update_qm_list(self):
...
             results.append(result.metric)
-        for metric in self.metrics_dict.keys():
+        for metric in list(self.metrics_dict.keys()):
```
pkgw: Since all that's being done with the `keys()` value is iterating over it, the listification isn't really necessary here.

hera_mc/tests/test_qm.py
bhazelton: thanks!

hera_mc/tests/test_qm.py
bhazelton: thanks!

PR #266, opened 2018-09-26
added active filter to part connection code

david-deboer:
This makes the active filter on the connection dossier display work and changes the default from all to active.


hera_mc/cm_handling.py
```diff
@@ -154,35 +157,60 @@ def get_entry(self, session):
         elif pad > 0:
             self.keys_up.extend([None] * abs(pad))
 
+    def add_filter(self, at_date, rev_type):
+        # Set default to use it
```
bhazelton: this comment doesn't clarify anything for me. Can you make a full docstring for this function? And it looks like whatever default you're talking about here isn't passed in, it's a property of the object, so please document how that comes in as well.

hera_mc/cm_handling.py
```diff
@@ -154,35 +157,60 @@ def get_entry(self, session):
...
+                print("For now, FULLY_CONNECTED is handled identically to ACTIVE")
+            for u, d in zip(self.keys_up, self.keys_down):
+                if u is not None:
+                    if not cm_utils.is_active(at_date, self.up[u].start_gpstime, self.up[u].stop_gpstime):
+                        self.up[u].skip_it = True
+                if d is not None:
+                    if not cm_utils.is_active(at_date, self.down[d].start_gpstime, self.down[d].stop_gpstime):
+                        self.down[d].skip_it = True
+
     def table_entry_row(self, headers):
```
bhazelton: a docstring would help here too.

hera_mc/cm_handling.py
```diff
@@ -154,35 +157,60 @@ def get_entry(self, session):
...
         tdata = []
         show_conn_dict = {'Part': self.entry_key}
 
         for u, d in zip(self.keys_up, self.keys_down):
-            if u is None:
+            if u is None or self.up[u].skip_it:
+                used_up = False
```
bhazelton: I'm not quite sure what `used_up` and `used_down` mean as variables.

hera_mc/cm_handling.py
```diff
@@ -387,23 +421,25 @@ def get_part_connection_dossier(self, hpn_list, rev, port, at_date, exact_match=
         Parameters
         -----------
         hpn_list:  the input hera part number [list of strings] (whole or first part thereof)
-        rev:  specific revision(s) or category(ies) ('LAST', 'ACTIVE', 'ALL', 'FULL')
+        rev:  specific revision(s) or category(ies) ('LAST', 'ACTIVE', 'ALL', 'FULL', specific)
               if list, must match length of hpn_list
         port:  a specifiable port name [string, not a list],  default is 'all'
-        at_date: reference date of dossier [anything get_astropytime can handle]
+        at_date: reference date of dossier, only used if rev==ACTIVE (and for now FULLY_CONNECTED...)
```
bhazelton: is `FULLY_CONNECTED` the same as `FULL` mentioned 3 lines above? If so, use the same name here.

hera_mc/cm_revisions.py
```diff
@@ -48,24 +43,33 @@ def get_revisions_of_type(hpn, rev_type, at_date=None, session=None):
     at_date:  astropy.Time to check for
     session:  db session, or hookup_dict if FULLY_CONNECTED
     """
+    if isinstance(hpn, six.string_types) and hpn.lower() == 'help':
+        print("""
+        Allowed revision types or revisions are (case doesn't matter and only need first 3 letters):
+            [ACT]IVE:  revisions active at_date ('now' as default) -- only one that uses 'at_date'
+            [LAS]T:  the last connected revision (could be active or not)
+            [FUL]LY_CONNECTED:  revision that is part of a fully connected signal path in current or supplied cache
```
bhazelton: Here you're clearer, but you actually suggest a 3rd option `FUL`. Maybe we can make this clearer throughout?

hera_mc/tests/test_parts.py
```diff
@@ -48,6 +48,7 @@ def test_update_new(self):
                 [ntp, 'X', 'start_gpstime', 1172530000]]
         part_connect.update_part(self.test_session, data, add_new_part=True)
         located = self.h.get_part_dossier([ntp], 'X', 'now', True)
+        print(located)
```
bhazelton: do we still want this print statement?

hera_mc/tests/test_sys_handling.py
```diff
@@ -35,6 +35,11 @@ def test_publish_summary(self):
         msg = self.sys_h.publish_summary()
         self.assertEqual(msg, 'Not on "main"')
 
+    def test_random_update(self):
+        si = sys_handling.StationInfo()
+        si.update_stn(None)
+        si.update_arrays(None)
```
bhazelton: should this test have some kind of assert? What is it testing?

hera_mc/tests/test_connections.py
```diff
@@ -80,6 +80,22 @@ def test_update_new(self):
...
+        x = self.h.get_part_connection_dossier('test_part2', 'active', 'all', at_date='now', exact_match=True)
+        y = list(x.keys())[0]
+        self.assertEqual(y, 'test_part2:Q')
+        self.h.show_connections(x)
+        old_time = Time('2014-08-01 01:00:00', scale='utc')
+        x = self.h.get_part_connection_dossier('test_part1', 'active', 'all', at_date=old_time, exact_match=True)
+        self.assertTrue(len(x) == 0)
+        x = self.h.get_part_connection_dossier('test_part2', 'active', 'all', at_date=old_time, exact_match=True)
+        self.assertTrue(len(x) == 0)
+        print(x)
```
bhazelton: extra print statement?

hera_mc/tests/test_parts.py
```diff
@@ -93,11 +93,24 @@ def test_cm_version(self):
...
     def test_get_revisions_of_type(self):
-        at_date = self.now
+        at_date = None
         rev_types = ['LAST', 'ACTIVE', 'ALL', 'A']
         for rq in rev_types:
             revision = cm_revisions.get_revisions_of_type('HH0', rq, at_date, self.test_session)
             self.assertTrue(revision[0].rev == 'A')
+        revision = cm_revisions.get_revisions_of_type('HH23', 'ACTIVE', 'now', self.test_session)
+        cm_revisions.show_revisions(revision)
+        revision = cm_revisions.get_revisions_of_type('help', 'help')
```
bhazelton: should there be some kind of assert here?

hera_mc/tests/test_sys_handling.py
david-deboer: These don't return anything -- they are very internal.  This just checks the code runs.

bhazelton: ok. Is there a way to test them that checks that the code does the right thing?

PR #271, opened 2018-10-22
Add most_recent filter

bhazelton:
Add an option to `ms_session._time_filter` to get the most recent entry. Propagate that option to get methods on mc_session that use it and add tests.

I realized the need for this while writing the correlator control code, but did it in another branch to keep the PRs organized.

### DIFF
--- a/hera_mc/mc_session.py
+++ b/hera_mc/mc_session.py
@@ -48,7 +48,7 @@ class MCSession(Session):
         return db_time
 
     def _time_filter(self, table_class, time_column, most_recent=None,
-                     starttime=startime, stoptime=None,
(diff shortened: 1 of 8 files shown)


bhazelton:
`most_recent=None` is the default value. I'm using `None` as a default because I want to do contextual defaulting (e.g. default `most_recent` to `True` if `starttime` is `None`), which is a common pattern.

bhazelton:
When `stoptime` is not specified, it just gets a the first value after `starttime`, it does not get all values between `starttime` and now. I can look at the logic a little more and see if I can make it simpler.

bhazelton:
Thanks!

Issue #264, opened 2018-09-21
correlator command interface & config support


bhazelton:

bhazelton:
done.

PR #275, opened 2018-10-26
Correlator command interface

bhazelton:
partially addresses #264

### DIFF
--- a/hera_mc/correlator.py
+++ b/hera_mc/correlator.py
@@ -14,7 +14,10 @@ from sqlalchemy import Column, BigInteger, Integer, Float, Boolean, String, Fore
 
 from . import MCDeclarativeBase
 
-DefaultRedisAddress = "redishost"
+# default acclen -- corresponds to just under 10 seconds
+DEFAULT_ACCLEN_SPECTRA = 147456
+
+DEFAULT_REDIS_ADDRESS = "redishost"
 
 # key is state type, value is method name in hera_corr_cm
 state_dict = {'taking_data': 'is_recording', 'phase_switching': 'phase_switch_is_on',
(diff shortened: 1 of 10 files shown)


hera_mc/correlator.py
```diff
@@ -0,0 +1,249 @@
...
+    acclen_spectra = Column(Integer)
+    integration_time = Column(Float)
+    tag = Column(String)
+
+    # TODO: this Foreign key arrangement requires an unneccesary column on this
+    # object (command) because this table only has rows when command='take_data'
+    # Is there a way to specify the constraint without this dummy column?
+    __table_args__ = (ForeignKeyConstraint(['time', 'command'],
+                                           ['correlator_control_command.time',
+                                            'correlator_control_command.command']), {})
```
bhazelton: @pkgw do you know of a way to set up this foreign key constraint without this unnecessary command column?

pkgw: No, I don't. This is not an area where I have much expertise at all, but I doubt it is possible to do so. My guess is that if you don't want the `command` column, you have to drop the explicit statement of the foreign key constraint. Which might be fine? It's kind of a judgment call, but given that these tables are likely to be append-only, I think it would be fairly safe to *not* formally enforce the constraint.

bhazelton:
@david-deboer or @pkgw I think this is ready for review when one of you has time.

@jack-h can you take a look at this too, particularly the `scripts/correlator_control.py`? That's the script that I'd hope we could use for all correlator commands.

jack-h:
Sorry to be unhelpful, but it turns out (I say this as if it wasn't a decision that I made before and forgot about) that the underlying method for determining how much data is recorded has two parameters --
1. How much time per file (i.e., how many integrations, but specified as a time)
2. How many files to record.
You can either change your duration specification to match this interface, or leave it as is and the correlator can do something reasonable -- eg., ask for 40 seconds of data and get 1 file with 4 x 10s integrations in, ask for 1 hour of data and get ~30 files all of reasonable <10, >1 GB size.
I'm happy with things as they are -- giving the user control of the amount of data in a file may or may not be something you actually want to do -- they get big *fast*.

The control script looks fine.

PR #276, opened 2018-10-29
fixed by retaining the requested revision

david-deboer:
This PR fixes an error when querying connections which occurred in some (most!) cases.


hera_mc/cm_revisions.py
```diff
@@ -82,6 +84,7 @@ def get_last_revision(hpn, session=None):
     session:  db session
     """
 
+    rq = 'LAS'
```
pkgw: I think it would be better to spell out the full words (LAST, ACTIVE, etc) here and below, to make things more understandable for new people who are reading the code.

hera_mc/cm_revisions.py
```diff
@@ -91,15 +94,15 @@ def get_last_revision(hpn, session=None):
         end_date = revisions[rev]['ended']
         if end_date is None:
             latest_end = cm_utils.get_astropytime('>')
-            no_end.append(Namespace(hpn=hpn, rev=rev, started=revisions[rev]['started'],
+            no_end.append(Namespace(hpn=hpn, rev=rev, rev_query=rq, started=revisions[rev]['started'],
```
pkgw: I also think it would be better just to write `rev_query='LAST'`, and avoid the variable, here and below. These functions aren't very long, but using the variable still puts unneeded distance between the actual value in question and its usage.

bhazelton:
It also looks like there are two new lines of code that aren't covered by tests:
cm_handling, line 312
cm_revisions, line 305

david-deboer:
I've made the changes.  cm_handling line 312 actually isn't needed and cm_revisions only has 223 lines

hera_mc/cm_handling.py
```diff
@@ -280,9 +280,17 @@ def get_part_type_for(self, hpn):
...
+        """
+        Makes sure that the hpn and revision requests are both lists and that they are
+        equal in length.  This is needed to handle the revision "categories"
+        """
+        if isinstance(hpn, list):
+            hpn_list = hpn
+        elif isinstance(hpn, six.string_types):
+            hpn_list = [hpn]
+        else:
+            raise ValueError("Wrong hpn type.")
```
bhazelton: This line is currently not being covered by tests.

hera_mc/cm_revisions.py
```diff
@@ -91,15 +93,15 @@ def get_last_revision(hpn, session=None):
...
+            no_end.append(Namespace(hpn=hpn, rev=rev, rev_query='LAST', started=revisions[rev]['started'],
                                     ended=revisions[rev]['ended']))
         elif end_date > latest_end:
             latest_rev = rev
             latest_end = end_date
     if len(no_end) > 0:
         return no_end
     else:
-        return [Namespace(hpn=hpn, rev=latest_rev, started=revisions[latest_rev]['started'],
+        return [Namespace(hpn=hpn, rev=latest_rev, rev_query='LAST', started=revisions[latest_rev]['started'],
```
bhazelton: This line is currently not being covered by tests.

PR #278, opened 2018-11-02
Add script to write files with node sensor values

bhazelton:
for @jack-h


bhazelton:
@david-deboer can you review this when you have time?

Issue #281, opened 2018-11-04
geo.py plots current antennas regardless of date


dannyjacobs:
For example the command
`geo.py -g --date=2016/11/08`
returns
![image](https://user-images.githubusercontent.com/63191/47968973-4d039d00-e02e-11e8-894b-5b3bb01c87c7.png)

dannyjacobs:
I think its line 390 in  geo_handling.plot_station_types
fig_num = self.plot_stations(self.station_types[st]['Stations'], query_date, **kwargs)

I think that which gets defined as "Stations" in the station_types is not recomputed depending on the date. Either the station_types should be recomputed or plot_stations should use the query_date is its passed to filter out things which don't yet exist.  
Ideally the filter would take place at one obvious place.

david-deboer:
I'm simplifying this (hopefully synonymous with fixing...) hopefully by tomorrow.

PR #282, opened 2018-11-04
Fix dishes plotted inthe past

dannyjacobs:
Fixes issue #281. Changes how get_location works. Now only returns dishes built prior to query_date.
I believe this is the desired behavior.


bhazelton:
This might be right, but I would have thought we'd want only the fully connected antennas at the date specified (which is not the same as the dishes having been built)...

bhazelton:
@dannyjacobs why are there 10 commits in this 2 line PR?

dannyjacobs:
Because I had a bunch of merge commits sitting on my master branch.  Looking at the final master-branch diff, I see only one file affected.

david-deboer:
I think we can withdraw this one for the new PR I'll put in in a few moments.

bhazelton:
will be fixed via #284

PR #285, opened 2018-11-09
Bg geo

david-deboer:
This now allows for 3 different geo plot backgrounds:  none, installed, all.  It also now allows for a file output.


bhazelton:
This is what I get when I just make the default plot. It's pretty ugly... Also there are no axes labels and the plot title is pretty unhelpful.

screen shot 2018-11-11 at 3 43 59 pm

dannyjacobs:
Looks pretty good with 
`scripts/geo.py -g --date 2018/11/08  -b layers --label=none -t HH,HA`
The marker size for the active ants could probably be tweaked, but I think its close enough to merge.

image

bhazelton:
@david-deboer it looks like there are 4 new uncovered lines

PR #289, opened 2018-11-16
Nosing

david-deboer:
100% coverage in cm_utils and cm_transfer.  Culled some unused code and fix a few things.


hera_mc/cm_transfer.py
```diff
@@ -48,7 +48,7 @@ def create(cls, time, git_hash):
         git_hash: String
             git hash of cm repository
         """
-        if not isinstance(time, Time):
+        if not isinstance(time, Time):  # pragma: no cover
```
bhazelton: is this line hard to cover?

hera_mc/cm_utils.py
```diff
@@ -22,13 +22,13 @@
 default_station_prefixes = ['HH', 'HA', 'HB']  # This is for defaults for sys etc.
 
 
-def get_cm_repo_git_hash(mc_config_path=None, cm_csv_path=None):
+def get_cm_repo_git_hash(mc_config_path=None, cm_csv_path=None):  # pragma: no cover
```
bhazelton: I'm confused because you have this `no cover` which should exclude the entire function and then another one on inside the function. I think this one should make the next one unnecessary, but I also think that we should be able to cover part of this function.

PR #290, opened 2018-11-19
fix testing setup to prevent too many connections

bhazelton:


bhazelton:
`test_db` is set as a global in `setup_package ` a few lines above my changes. You can see it by clicking on the light blue box with a symbol showing up and down arrows at the top left of the code changes panel.

bhazelton:
I didn't have to modify anything else, either on my machine or on Travis (which would have shown up as other code changes). Does it work on your machine?

PR #294, opened 2018-11-20
 Add write_to_file option to all functions that call _time_filter

bhazelton:


bhazelton:
Sorry this is so confusing, I'll try to clarify in docstrings but first I'll respond to you to see if my explanation makes sense.

The default is to return the most recent record(s), but it can be more than one record if they occur at the same time.

If you set a starttime and no stoptime, you get the first record(s) after that starttime, it can be more than one record if there are more than one at the same time.

If you want records from multiple times you need to set both starttime and stoptime.

I'm can add some options to be able to start with the earliest record or end with the latest record, I'll work on that.

The logic starting on line 154 is to get all the records that occur at the same time (if most recent is set or there's a starttime but no stoptime). I thought it was confusing that it would randomly return one of the records at a given time with no guarantee which one, so I thought it was better to return all of them that have the same time. The bit about setting query=first_query is just to handle the case where there are no records returned so there's no time to use for the second query.

bhazelton:
@david-deboer I think this is ready for your review.

PR #301, opened 2018-12-05
fixed installed antennas not writing to file and added some tests

david-deboer:
Found this when trying to reconcile cm and the site google sheet.

### DIFF
--- a/hera_mc/geo_handling.py
+++ b/hera_mc/geo_handling.py
@@ -363,14 +363,15 @@ class Handling:
             plt.title(fig_label)
         return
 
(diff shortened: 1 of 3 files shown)


bhazelton:
I'm super confused. What code changes in this PR fix the bug? I don't see anything about writing to a file...

david-deboer:
In geo.py, the fix is including the "if args.fg_action.startswith('I')" fixes the problem (very subtle). The problem is that the "installed" option was done as a background, so the file flag was not set (which is done on the foreground).

bhazelton:
wow, ok.

PR #302, opened 2018-12-11
Install doc

david-deboer:
just a few notes in the install doc based on reinstalling Postgres

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -14,7 +14,7 @@ Clone the following two repositories:
 * https://github.com/HERA-Team/hera_cm_db_updates
 
 Then install by:
-1. within the hera_mc directory type `python setup.py install`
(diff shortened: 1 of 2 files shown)


INSTALL.md
```diff
@@ -14,7 +14,7 @@ Clone the following two repositories:
 * https://github.com/HERA-Team/hera_cm_db_updates
 
 Then install by:
-1. within the hera_mc directory type `python setup.py install`
+1. within the hera_mc directory type `python setup.py install` or `pip install .`
```
bhazelton: Do we think `pip install .` is better? If so, let's just say to use that one.

david-deboer: It seems to pollute the hard drives less.  But either are fine and maybe someone doesn't have pip installed?

PR #303, opened 2018-12-14
Cleanup version and setup code, add changelog, increase version number

### LINKED ISSUE #299: Update version number.
We should probably update the version:  (a) it has come a long way since we've last done it (b) it is helpful if the version number changes so that you can check.

bhazelton:
fixes #299


bhazelton:
@ddeboer turns out yaml is only needed for testing, not for running the code. So I'll add it a slightly different way to setup.py

hera_mc/__init__.py
```diff
@@ -84,7 +84,9 @@ def NotNull(kind, **kwargs):
     return Column(kind, nullable=False, **kwargs)
 
 
-from .version import __version__  # noqa
+from . import version  # noqa
+__version__ = version.version
```
plaplant: Is there a reason why we have to define this here, and we can't wait until all the imports are done?

bhazelton: This is the only place the `__version__` attribute is defined. Oh! you mean move that line after the imports. Sure, I can do that.

bhazelton:
@plaplant I don't think we can, there's a bunch of other code above those imports that can't be moved after them

PR #304, opened 2018-12-18
Delete the hookup cache after testing.

### LINKED ISSUE #253: Have nose tests delete hookup_cache_1/2 files upon teardown
Currently they write the cm cache files, which persist.  The files are in .hera_mc

david-deboer:
and clean up the install.md

fixes #253

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -1,7 +1,8 @@
 HERA M&C Installation
 =====================
 
-HERA M&C requires `hera_mc` and postgreSQL.  Note:  if you are only using it to view configuration management, you may instead use SQLITE.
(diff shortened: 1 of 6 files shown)


INSTALL.md
```diff
@@ -34,22 +36,13 @@ Python Prerequisites
 - pandas
 - psutil
 - pyproj
+- pyyaml
```
bhazelton: should add: six
we also install pyuvdata and h5py in our travis setup, not sure if those are needed as well.

bhazelton: interestingly, we don't install pyyaml in Travis...

PR #305, opened 2018-12-19
Fix config monitor

bhazelton:
Fix a few bugs that were preventing configs from being recorded properly in M&C

### DIFF
--- a/hera_mc/mc_session.py
+++ b/hera_mc/mc_session.py
@@ -2428,6 +2428,7 @@ class MCSession(Session):
 
                         # add it to the config file table
                         self.add(config_file_obj)
+                        self.commit()
                     else:
                         command_list.append(config_file_obj)
 
@@ -2474,6 +2475,7 @@ class MCSession(Session):
                 getattr(corr_controller, corr.command_dict[command])
 
             self.add(command_obj)
+            self.commit()
             if command == 'take_data':
                 # update the starttime with the actual starttime of the correlator
                 take_data_args_obj = \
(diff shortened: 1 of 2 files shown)


bhazelton:
currently failing on an iers downloading issue... should be transient

hera_mc/mc_session.py
```diff
@@ -2062,13 +2063,15 @@ def add_correlator_config_from_corrcm(self, config_state_dict=None,
 
                     # add it to the config file table
                     self.add(config_file_obj)
+                    self.commit()
```
plaplant: Do the `commit()` calls actually perform the transactions we've set up? If so, do we need to do any error handling in case they don't stick?

hera_mc/tests/test_correlator.py
```diff
@@ -362,7 +362,13 @@ def test_config_errors(self):
     def test_add_correlator_config_from_corrcm_onsite(self):
 
         result = self.test_session.add_correlator_config_from_corrcm(testing=True)
+
         self.assertTrue(len(result) > 0)
+        if len(result) == 2:
```
plaplant: Would we ever get more than 2 results?

bhazelton: No. We can only get 2 at most, that happens if there's not a corresponding file already known about, so we need to make an entry in the correlator_config_file table as well as an entry in the correlator_config_status table.

hera_mc/mc_session.py
bhazelton: If they don't stick, the later parts of this function will fail as well, but there's not much that can be done at that point. I guess the only thing you might want to do is to remove the file from the librarian, but I'm not sure there's a good way to do that.

plaplant: Yeah, I think it's hard to do delete something from the librarian (on purpose...). I'm fine leaving this as is then.

PR #312, opened 2019-01-22
WIP: Add views with unix time columns

bhazelton:
to support plotting in grafana, based on issue #306

### DIFF
--- a/hera_mc/observations.py
+++ b/hera_mc/observations.py
@@ -79,5 +79,6 @@ class Observation(MCDeclarativeBase):
 
 
 class HeraObsView(MCDeclarativeBase):
-    view = True
+    view_base_table = 'hera_obs'
     __table__ = mc_view(Observation, ['starttime', 'stoptime'], 'hera_obs_view')
+    __tablename__ = 'hera_obs_view'
--- a/hera_mc/observations.py
+++ b/hera_mc/observations.py
@@ -14,7 +14,6 @@ from astropy.coordinates import EarthLocation
 from sqlalchemy import Column, BigInteger, Float
 from sqlalchemy.ext.hybrid import hybrid_property
 
-from . import geo_handling
 from . import MCDeclarativeBase, mc_view, DEFAULT_GPS_TOL, DEFAULT_DAY_TOL, DEFAULT_HOUR_TOL
 
 
@@ -78,7 +77,5 @@ class Observation(MCDeclarativeBase):
                    lst_start_hr=starttime.sidereal_time('apparent').hour)
 
 
-class HeraObsView(MCDeclarativeBase):
-    view_base_table = 'hera_obs'
-    __table__ = mc_view(Observation, ['starttime', 'stoptime'], 'hera_obs_view')
-    __tablename__ = 'hera_obs_view'
+HeraObsView = type('HeraObsView', (MCDeclarativeBase,),
+                   mc_view(Observation, ['starttime', 'stoptime']))
--- a/hera_mc/observations.py
+++ b/hera_mc/observations.py
@@ -15,7 +15,7 @@ from sqlalchemy import Column, BigInteger, Float
 from sqlalchemy.ext.hybrid import hybrid_property
 
 from . import geo_handling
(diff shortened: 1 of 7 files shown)


hera_mc/tests/test_server_status.py
```diff
@@ -46,9 +46,13 @@ def test_repr(self):
                 servstat = LibServerStatus(**exp_columns)
                 class_name = 'LibServerStatus'
 
-        rep_string = ('<' + class_name + '(test_host, ' + str(exp_columns['mc_time'])
-                      + ', 0.0.0.0, ' + str(exp_columns['mc_system_timediff'])
-                      + ', 16, 20.5, 31.4, 43.2, 32.0, 46.8, 510.4, 10.4)>')
+        rep_string = ('<' + class_name + '(hostname: test_host, mc_time: '
```
dannyjacobs: Why is this in this PR?

hera_mc/observations.py
```diff
@@ -76,3 +76,8 @@ def create(cls, starttime, stoptime, obsid, hera_cofa):
         return cls(obsid=obsid, starttime=starttime.gps, stoptime=stoptime.gps,
                    jd_start=starttime.jd,
                    lst_start_hr=starttime.sidereal_time('apparent').hour)
+
```
dannyjacobs: Do we have to create a new view table object for every table?  I was hoping we could handle this transparently so that new tables will show up without having to manually add the matching view, or without having to go through and manually add views for each table.

hera_mc/tests/test_server_status.py
bhazelton: Because I was trying to debug things and I needed it. I can move it to another PR if you want, but I do want this change because it makes it much easier to debug.

hera_mc/observations.py
bhazelton: Yes, we have to make one for each table. But I put all the hard work in the `__init__.py`, so it's a 3-liner to add it for any new table and I can add it to the list of steps for making a new table in the readme. There may be a more clever way to do it, but I don't know how.

hera_mc/tests/test_server_status.py
dannyjacobs: nope, its fine

hera_mc/observations.py
bhazelton: Yes it is. I made one and here's the output, noted as:
table name: time columns

Note that several of the listed columns are not in gps time and it's not super straightforward to infer from code which ones are and which ones aren't in gps seconds.
```
Tables with missing views:

lib_server_status: mc_time, mc_system_timediff, uptime_days
lib_status: time
autocorrelations: time
dubitable: start_gpstime, stop_gpstime
part_info: posting_gpstime
cm_version: update_time
subsystem_error: time, mc_time
parts: start_gpstime, stop_gpstime
lib_raid_status: time
lib_raid_errors: time
lib_remote_status: time, ping_time
rtp_server_status: mc_time, mc_system_timediff, uptime_days
rtp_status: time
weather_data: time
node_sensor: time
node_power_status: time
node_power_command: time
correlator_control_state: time
correlator_control_command: time
connections: start_gpstime, stop_gpstime
geo_location: created_gpstime
lib_files: time
rtp_process_event: time
rtp_process_record: time
rtp_task_resource_record: start_time, stop_time
ant_metrics: mc_time
array_metrics: mc_time
correlator_take_data_arguments: time, starttime_sec, starttime_ms, integration_time
correlator_config_status: time
correlator_config_command: time
```

bhazelton: I can make this test part of our standard tests if you like. Because of some of the vagaries of the column naming, I don't see a good way to use this to make views by default, but I can keep thinking about that, maybe I'll come up with something.

bhazelton:
@dannyjacobs can you verify that using this branch, grafana can see the `hera_obs_view`? I'm not perfectly sure that it'll all just work and I want to test it on this view before I make all of them.

bhazelton:
@dannyjacobs I fixed the alembic integration and I think this should now work for you as long as you do an `alembic upgrade head`. I tested it onsite and I can query the view.

PR #314, opened 2019-02-07
fixed node_sensor(s) name

david-deboer:
Fixed the node_sensor table name in the write_records_to_file script.


bhazelton:
This is failing because of some pep8 issues (there was an update to pycodestyle that's catching a few new things). I fixed them in PR #315, once that's merged you can rebase and the problems should be solved.

PR #317, opened 2019-02-21
updated the overlap checker

david-deboer:
When I updated the hookup and pulled it back down, the overlap checker found overlaps that just shared an end point (we don't want it to find contiguous but non-overlapping ones).  It did not flag those earlier, which I find very confusing since it did not relate to any new connection.  I've rewritten the overlap checker to fix that (and made it more "elegant").

### DIFF
--- a/hera_mc/cm_health.py
+++ b/hera_mc/cm_health.py
@@ -14,26 +14,16 @@ from . import part_connect as PC
 from . import cm_revisions
 
 
-def check_for_overlap(intervals):
+def check_for_overlap(interval):
     """
     intervals:  gps_time intervals to test for overlap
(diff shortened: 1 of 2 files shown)


bhazelton:
Is there a test in the test suite that specifically fails before this fix and doesn't after?

david-deboer:
I don’t see one — with no code change the check for duplicates began showing results that shared a point at a limit.  I can’t really reproduce.  With this code I can make it find or not find those cases by changing the <=, >= to <, > and back.  I’d like to make this change, and then I’ll monitor how it responds when the cm tables get updated.

bhazelton:
When you find and fix a bug, it's good practice to make a test that fails as a result of the bug and passes after your fix.

PR #323, opened 2019-03-01
Implement ability to view all notes and to filter on dates

david-deboer:
Per the Wed 2/27 telecon, I added the ability to look at all of the notes and to sort by dates.

### DIFF
--- a/hera_mc/cm_handling.py
+++ b/hera_mc/cm_handling.py
@@ -29,7 +29,7 @@ class PartDossierEntry():
                'input_ports': 'Input', 'output_ports': 'Output',
                'part_info': 'Note', 'geo': 'Geo', 'post_date': 'Date', 'lib_file': 'File'}
 
-    def __init__(self, hpn, rev, at_date, early, sort_notes_by='part'):
+    def __init__(self, hpn, rev, at_date, notes_start_date, sort_notes_by='part'):
         if isinstance(hpn, six.string_types):
             hpn = hpn.upper()
         if isinstance(rev, six.string_types):
@@ -37,8 +37,8 @@ class PartDossierEntry():
         self.hpn = hpn
         self.rev = rev
         self.entry_key = cm_utils.make_part_key(hpn, rev)
-        self.time = at_date
-        self.early = early
+        self.at_date = at_date
+        self.notes_start_date = notes_start_date
         self.sort_notes_by = sort_notes_by
         self.part = None  # This is the part_connect.Parts class
         self.part_info = []  # This is a list of part_connect.PartInfo class entries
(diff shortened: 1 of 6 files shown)


hera_mc/cm_handling.py
```diff
@@ -363,23 +382,30 @@ def get_part_dossier(self, hpn, rev, at_date, exact_match=False, full_version=Tr
         hpn:  the input hera part number [string or list-of-strings] (whole or first part thereof)
         rev:  specific revision(s) or category(ies) ('LAST', 'ACTIVE', 'ALL', 'FULL', specific)
               if list, must match length of hpn
-        at_date:  reference date of dossier only used for ACTIVE
+        at_date:  reference date of dossier (and stop_date for displaying notes)
```
bhazelton: I thought we discussed this and decided not to conflate these two concepts into one keyword.

scripts/parts.py
```diff
@@ -24,18 +24,24 @@
...
                         dest='exact_match', action='store_true')
-    parser.add_argument('--notes', help="(For action=part_info):  Displays part notes with posting dates and not other info",
+    parser.add_argument('--notes', help="<For action=part_info>:  Displays part notes with posting dates and not other info",
                         action='store_true')
+    parser.add_argument('--sort_notes_by', help="<For action=part_info --notes> Can sort the notes listing by part \
+                        ('part') or posting time ('post') [part]", default='part')
     cm_utils.add_verbosity_args(parser)
     cm_utils.add_date_time_args(parser)
+    parser.add_argument('--notes_start_date', help="<For part_info notes> start_date for notes [<]", default='<')
+    parser.add_argument('--notes_start_time', help="<For part_info notes> start_time for notes", default=0.0)
```
bhazelton: For a script, I'd think it'd be a little easier to be able to enter a date and time together rather than having to specify two separate keywords. Astropy Time can certainly handle just being passed a date string with no time info -- it just uses midnight as the time if it's not specified, which I think is the same as your default here.

hera_mc/cm_handling.py
david-deboer: I spent some time with this and decided that keeping the 'at_date' and changing 'early' to 'start_date' was more clear in actual use.  Otherwise, if you want to filter on notes you have to add another identical date to the call (so you'd have two of the same), which seems silly since the context is relatively clear.  The other approach would be to rename 'at_date' to 'start_date', which would make it very confusing in every context, including this one.

scripts/parts.py
david-deboer: That would make it inconsistent with every other cm script.  If you think we should change this, can you make an issue then I'll go through and change all?

bhazelton: I don't feel super strongly about this and I agree consistency is more important. Happy to review a PR if you decide you want to make the change across all of them.

hera_mc/cm_handling.py
bhazelton: ok. I'll let your actual users weigh in via issues if they find it confusing.

PR #324, opened 2019-03-04
Mc pdf

dannyjacobs:
adds the pdf for m&c table definitions


bhazelton:
This PR doesn't actually have the pdf in it. It also has a ton of merge commits that add nothing. Closing it in favor of #326 (which has the pdf added)

PR #331, opened 2019-03-19
Add snap status table

bhazelton:
To track the snap status as reported in the correlator redis database


bhazelton:
@dannyjacobs you might have an opinion about this.

bhazelton:
I checked out the coverage report, it appears to be getting confused because of line number changes (or something) -- all of the lines it labels as newly uncovered were previously uncovered. And the new lines that aren't covered ARE covered on site, but the tests that cover them can't be run on Travis (because they involves talking to the correlator redis DB).

PR #333, opened 2019-03-23
Add antenna status table

bhazelton:
To track the antenna status as reported in the correlator redis database (from the SNAPs).


bhazelton:
@dannyjacobs you might want to look at this one too.

bhazelton:
@david-deboer The snap stuff isn't really in this PR, it was an artifact of working off of the previous branch and that PR not being merged yet. Now that it is merged, I've rebased and only the new stuff shows up in the changes.

Issue #328, opened 2019-03-18
Hookup queries on SNAP and NODE don't return all.


david-deboer:
The query (e.g. hookup.py -p SNP) will return all SNAPs hooked up, but won't return all connections since there are multiple e and n ports (NODE as well).  Fortunately there is no "pipeline" mode this seems to impact, but it should be fixed.

david-deboer:
This is (mostly) addressed in the the current PR -- however, it does not address the "node issue" adequately, so I'll keep this issue open.

PR #335, opened 2019-03-23
Update hookup to allow multiple paths.

david-deboer:
This address issue #328.  It punts a bit on getting a full node return, which will likely need to be a special module to add.  This modifies cm_hookup quite substantially (hopefully better;-).  It quarantines all of the ad hoc stuff related to specific hookups into one place (cm_sysdef), which makes it much cleaner.


hera_mc/cm_sysdef.py
```diff
@@ -0,0 +1,128 @@
...
+                        }
+corr_index = {"parts_hera": 5, "parts_paper": 9, "parts_rfi": 3}
+all_pols = ["e", "n"]
+redirect_part_types = ['node']
+
+
+def handle_redirect_part_types(part):
+    print("Redirected parts require that you determine the set of parts externally and re-run.")
+    print("\t\tFor this part type:  'parts.py conn -p {}'".format(part.hpn))
+    print("\t\tThen re-run hookup with a csv list of parts.")
```
bhazelton: I'm not sure I really understand this. If all you're going to do is rerun using the list passed out by the suggested command, why not do it automatically in the code (rather than having a human in the loop)?

hera_mc/cm_sysdef.py
```diff
@@ -0,0 +1,128 @@
...
+
+
+def next_connection(options, rg):
+    """
+    This checks the options and returns the next part.
+    """
+
+    port = {'up': 'out', 'down': 'in'}
+    this = {'up': 'down', 'down': 'up'}
+    that = {'up': 'up', 'down': 'down'}
```
bhazelton: looks like what you mean by "that" is really the next part? then I'd suggest using "next". "This" makes sense, "that" confuses me (and probably other people reading the code).

hera_mc/cm_hookup.py
```diff
@@ -276,50 +250,50 @@ def determine_hookup_cache_to_use(self, force_new=False):
...
         port_query:  a specifiable port name to follow or 'all',  default is 'all'.
         exact_match:  boolean for either exact_match or partial
-        levels:  boolean to include correlator levels
-        force_new:  boolean to force a full database read as opposed to checking file
-                    this will also rewrite the cache-file
-        force_specific:  boolean to force this to read/write the file to use the supplied values
-                         Setting this makes get_hookup provide specific hookups (mimicking the
-                         action before the cache file option was instituted)
-        force_specific_at_date:  date for hookup check -- use only if force_specific
+        levels:  boolean to include correlator levels (Currently not working.)
```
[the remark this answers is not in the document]
david-deboer: This has been helpful in the past, to have the hookup printout display the autos.  I'd like to keep that ability (with your new stuff, that should now be possible).

hera_mc/cm_hookup.py
```diff
@@ -276,50 +250,50 @@ def determine_hookup_cache_to_use(self, force_new=False):
...
+        force_new_cache:  boolean to force a full database read as opposed to checking file
+                          this will also rewrite the cache-file
+        force_db:  boolean to force this to read/write the file to use the supplied values
```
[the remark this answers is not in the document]
david-deboer: I tried to clarify the docstring

hera_mc/cm_hookup.py
```diff
@@ -433,76 +414,46 @@ def _follow_hookup_stream(self, part, rev, pol):
             hu.append(pn)
         return hu
 
-    def _recursive_go(self, direction, part, rev, port, pol):
+    def _recursive_go(self, rg):
```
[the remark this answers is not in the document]
david-deboer: It was a legacy word, which I've changed with 'connect'

hera_mc/cm_sysdef.py
david-deboer: This should be an issue in the future, since it is not straightforward in this scope.  I would like to implement that function, but it will need a bit of thought/coding.  For now, the "helpful" message may "help" (to remind me at least).  It also tells the user that the request won't be complete.

bhazelton: ok. Do you want to make an issue?

hera_mc/cm_hookup.py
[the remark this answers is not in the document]
bhazelton: maybe. The new stuff I've done recently doesn't include the autos, but that could be coming.

hera_mc/cm_hookup.py
```diff
@@ -272,27 +272,28 @@ def get_hookup(self, hpn_list, rev='ACTIVE', port_query='all', exact_match=False
...
         levels:  boolean to include correlator levels (Currently not working.)
-        force_new_cache:  boolean to force a full database read as opposed to checking file
-                          this will also rewrite the cache-file
-        force_db:  boolean to force this to read/write the file to use the supplied values
-                   Setting this makes get_hookup provide specific hookups (mimicking the
-                   action before the cache file option was instituted)
+        force_new_cache:  boolean to force a full database read as opposed to the cache file.
+                          This will also rewrite the cache file in the process
+        force_db:  boolean to force use of the database, rather than the cache file.  This is primarily
+                   needed to force a search if the date is not the current date (see date below)
```
bhazelton: So do you need to set both `force_db` and `force_db_at_date` or does setting `force_db_at_date` also set `force_db`? If the only use for `force_db` is for using it with `force_db_at_date`, why not only have the `force_db_at_date` keyword, which had db in the name already.

hera_mc/cm_sysdef.py
david-deboer: I updated issue #328

hera_mc/cm_hookup.py
david-deboer: Excellent point - I don't need both.  I fixed that.

hera_mc/cm_hookup.py
```diff
@@ -330,20 +303,22 @@ def get_hookup(self, hpn_list, rev='ACTIVE', port_query='all', exact_match=False
...
         if not requested_list_OK_for_cache:
-            s = "Hookup request list does not match cache file - using database."
-            d = {'hpn_list (request)': hpn_list, 'hookup_list_to_cache': self.hookup_list_to_cache}
-            cm_utils.log(s, params=d)
-        if force_specific or not requested_list_OK_for_cache:
+            force_db = True
+            at_date = 'now'
+        if force_db_at_date is not None:
+            force_db = True
+            at_date = force_db_at_date
```
bhazelton: Looks like these 2 lines aren't covered by tests at the moment.

PR #338, opened 2019-04-05
Sysdef gen

### LINKED ISSUE #328: Hookup queries on SNAP and NODE don't return all.

david-deboer:
This tries to fully fix the hookup recursion and handle the "redirect" parts to close the issue.

Closes #328


bhazelton:
you mention closing an issue, but which issue does it close? If it does close an issue, you should mention it in the first comment in the PR like: `closes #<number>` so that the issue is closed when the PR is merged.

david-deboer:
Closes #328

hera_mc/cm_sysdef.py
```diff
@@ -6,123 +6,229 @@
...
+# Initialize the dictionaries
+corr_index = sys_init(checking_order, None)
+all_pols = sys_init(checking_order, [])
+redirect_part_types = sys_init(checking_order, [])
+single_pol_labeled_parts = sys_init(checking_order, [])
+
+# Redefine dictionary as needed
+corr_index['parts_hera'] = 6
+corr_index['parts_paper'] = 10
+corr_index['parts_rfi'] = 4
```
bhazelton: what do these lines do? I'm unclear about why they're needed right after you did an initialization and on what the hardcoded integers mean.

hera_mc/cm_sysdef.py
```diff
@@ -6,123 +6,229 @@
...
     part:  current part dossier
     port_query:  the ports that were requested ('e' or 'n' or 'all')
+    hookup_type:  if not None, will use specified hookup_type
+                  otherwise it will look through in order
     """
-
-    all_pols_lo = [x.lower() for x in all_pols]
+    if hookup_type is None:
+        hookup_type = find_hookup_type(part.part_type, None)
+    global this_hookup_type, all_pols, pind
```
bhazelton: globals worry me a little, why are they required here?

hera_mc/sys_handling.py
```diff
@@ -220,7 +221,7 @@ def get_fully_connected_location_at_date(self, stn, at_date):
                 station_info.stop_date = fctime['end']
         return station_info
 
-    def get_cminfo_correlator(self):
+    def get_cminfo_correlator(self, hookup_type=None):
```
bhazelton: What is the default if it's left as `None`? This should be clarified in docstrings here and for the other functions that have this keyword. The allowed options should also be clarified.

hera_mc/cm_hookup.py
```diff
@@ -202,22 +197,21 @@ class Hookup:
     hookup_list_to_cache and if the cache file is current relative to the cm_version
     """
     hookup_list_to_cache = cm_utils.all_hera_zone_prefixes
-    if six.PY2:
+    if six.PY2:  # pragma: no cover
```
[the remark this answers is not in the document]
david-deboer: It does python 3 second, so it doesn't give you "credit" for that statement (i.e. coveralls final report shows it as uncovered).  I'm striving for 100%!

hera_mc/cm_sysdef.py
david-deboer: cm_sysdef contains all of the "hard-coded" aspects of a system (and only in cm_sysdef!). This is needed to tell which part position is the correlator so cm_corrinfo knows where to look.

hera_mc/cm_hookup.py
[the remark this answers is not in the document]
bhazelton: coveralls should be combining the two reports. I'll see if I can figure out what's going on.

hera_mc/cm_sysdef.py
david-deboer: I agree - this should probably become a class.   (I may do one more clean-up PR for this).  This gets called multiple times in a hookup pass and the globals are specified at the start.  It might well be that I don't need those declarations, but in the absence of surety I opted to include it.

hera_mc/cm_hookup.py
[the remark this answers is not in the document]
bhazelton: It is combining them. If you look at the files, it shows a `2x` on most lines because they were hit in python 2 & 3, and only a `1x` on the python 3 line because it's only hit once. Can you take out this comment and so we can see what's going on? If it's not being hit by the python 2 tests we should figure out why not.

hera_mc/cm_sysdef.py
bhazelton: ah, ok. Maybe add a more descriptive comment?

hera_mc/sys_handling.py
bhazelton: I like what you added to this function. Can you add it to other API level functions (i.e. not starting with underscores) that have this keyword?

hera_mc/cm_sysdef.py
bhazelton: Ok. I think it would be better to make it an object, what I'm most worried about is knock-on effects where it hangs around and affects later calls, but if you've tested that I can live with it short term (maybe make an issue to convert it to an object?).

PR #340, opened 2019-04-09
Sysdef class and add test data

### LINKED ISSUE #252: Update hera_mc test data
(a) more antennas
(b) new signal chain

david-deboer:
closes #252 

This makes sysdef a class and adds 3 antenna hookups to the test data.


bhazelton:
It looks like your test data changes didn't require a change to `test_sys_handling.test_correlator_info`, but I would expect it to -- that was part of the point of the issue. Do you know why it's not affecting that test?

PR #341, opened 2019-04-10
Iss332

### LINKED ISSUE #332: add method to check for duplicate connections, differing only in time
We currently have some of these in the onsite database. SNAPs `SNPA000024` and `SNPA000008` have duplicate downstream (node) connnections that differ only in their time.

david-deboer:
Closes #332 
This will now only return the active connections (if at_date is not None).  Before it returned all connections, as long as the part itself was active.


david-deboer:
It also changes a couple of filenames to be more consistent/less confusing (hopefully).  It will require coordinating with the correlator code when we put it on qmaster.

hera_mc/cm_health.py
```diff
@@ -22,7 +22,7 @@ def check_for_overlap(interval):
     """
     maxx = max([x for x in interval[0] + interval[1] if x is not None]) + 100
     for i, ival in enumerate(interval):
-        interval[i] = [x if x is not None else maxx for x in ival]
+        ival[1] = maxx if ival[1] is None else ival[1]
```
bhazelton: this is changing the functionality of this code, not just renaming things. Is it doing what you want?

david-deboer: I started looking at the issue to address a "duplication" issue, so to make sure I checked and changed this line.  it worked before but I always struggled with the logic and this is easier.

bhazelton:
If this PR fixed the issue, then test_correlator should have errored on line 1363.

PR #349, opened 2019-05-25
Cumulative plot for "dataview"

david-deboer:
This generates data to see what antennas are fully connected over time (e.g. May 2019 NSF presentation).  It slightly changes how multiple connections are handled (warning, not error).


hera_mc/cm_dataview.py
```diff
@@ -10,8 +10,21 @@
...
 from astropy.time import Time, TimeDelta
 import numpy as np
+import os.path
+import sys
 
-from . import mc, cm_utils, cm_revisions, cm_sysutils
+from . import mc, cm_utils, cm_revisions, cm_sysutils, geo_handling
+
+
+def read_antennas():
```
bhazelton: missing docstring

hera_mc/cm_dataview.py
```diff
@@ -26,7 +39,7 @@ def __init__(self, session=None):
         else:
             self.session = session
 
-    def ants_by_day(self, start, stop, time_step, output=None, station_types_to_check='default', output_date_format='jd'):
+    def connected_by_day(self, start, stop, time_step, output=None, station_types_to_check='default', output_date_format='jd', hookup_type='parts_hera'):
```
bhazelton: docstring needs a Returns section. See here for the format: https://numpydoc.readthedocs.io/en/latest/format.html

hera_mc/cm_dataview.py
```diff
@@ -59,9 +72,43 @@ def ants_by_day(self, start, stop, time_step, output=None, station_types_to_chec
...
+        """
+        Return dated list of connected stations.
+
+        Parameters:
+        ------------
+        start:  start (astropy.Time)
+        stop:  stop (astropy.Time)
+        time_step:  desired time_step resolution in days or fractions thereof (float or int)
+        output:  Optional filename to write (and shows on screen).  If None, only returns dictionary.
+        output_date_format: jd or ymd
```
bhazelton: Please add a Returns section.

hera_mc/geo_handling.py
```diff
@@ -194,12 +194,16 @@ def find_antenna_at_station(self, station, query_date):
...
+                if ctr:
+                    print("Already connected: {} -> {}".format(antenna_connected.upstream_part, antenna_connected.downstream_part))
+                    print("...trying to add:  {} -> {}".format(conn.upstream_part, conn.downstream_part))
                 antenna_connected = copy.copy(conn)
                 ctr += 1
         if ctr == 0:
             return None, None
         elif ctr > 1:
-            raise ValueError('More than one active connection between station and antenna')
+            warnings.warn('More than one active connection between station and antenna')
```
bhazelton: can you explain the use case here? Seems pretty bad (and hard to interpret) to me if there are multiple connections. Also, your docstring says it raises a ValueError in this case, not a warning. The docstring should be updated.

hera_mc/geo_handling.py
```diff
@@ -194,12 +194,16 @@ def find_antenna_at_station(self, station, query_date):
         ctr = 0
         for conn in connected_antenna:
             if conn.stop_gpstime is None or query_date.gps <= conn.stop_gpstime:
+                if ctr:
+                    print("Already connected: {} -> {}".format(antenna_connected.upstream_part, antenna_connected.downstream_part))
+                    print("...trying to add:  {} -> {}".format(conn.upstream_part, conn.downstream_part))
```
bhazelton: This confuses me. This method is supposed to "check to see what antenna is at a station". I would not expect to get print statements about trying to add things from this method.

hera_mc/tests/test_dataview.py
```diff
@@ -26,9 +27,25 @@ def test_ant_by_day(self):
...
+        start = cm_utils.get_astropytime(1184354550)
+        stop = cm_utils.get_astropytime(1184354600)
+        interval = 10.0
+        filename = None
+        n = dv.connected_by_day(start, stop, interval)
+        self.assertEqual(len(n.keys()), 1)
+        filename = os.path.expanduser('~/hera_mc_dataview_test.txt')
+        n = dv.connected_by_day(start, stop, interval, output=filename, output_date_format='iso')
+        self.assertEqual(n, None)
+        os.remove(filename)
```
bhazelton: this file will not be removed if this test fails. Is that ok? We could handle this in the package setup/teardown (by making a folder to put test output into in setup and removing the folder in teardown) if we should.

hera_mc/tests/test_geo_location.py
```diff
@@ -143,7 +143,8 @@ def test_find_antenna_station_pair(self):
         self.assertTrue(ant == 'A23')
         ant, rev = self.h.find_antenna_at_station('BB23', 'now')
         self.assertEqual(ant, None)
-        self.assertRaises(ValueError, self.h.find_antenna_at_station, 'HH68', 'now')
```
bhazelton: was this testing the error that's become a warning? We should check that the warning is raised using the `checkWarnings` method in `test/__init__.py`

david-deboer: Untitled-1

david-deboer: See above for sample data.  The purple dashed line is from cm_dataview: "ants_by_day" and the black and green solid lines are form cm_dataview: "connected_by_day".  It is a brute force slow step through time.

hera_mc/tests/test_dataview.py
david-deboer: I tried teardown first, but it hung (remember, this happened in the past too).

bhazelton: hmm, ok. I might try to move us in that direction in the future, but we can leave it alone for now. You could use a try/finally block to ensure the file is deleted if you want.

hera_mc/geo_handling.py
bhazelton: Is the top line docstring description accurate? If so, what are these prints supposed to convey to me?

hera_mc/geo_handling.py
david-deboer: I fixed the docstring.  A warning and error convey the same information, and the warning is preferable since it can handle all non-warned ones (which may take some time and you wouldn't want to repeat).  Maintenance of this is an on-going thing.

hera_mc/geo_handling.py
david-deboer: Confusion warranted ;-).  I've fixed that.

hera_mc/tests/test_geo_location.py
david-deboer: I've added the checkWarning test

bhazelton: good. This plot will never be seen because it's inside a code review thread. Can you put it directly on the PR as a new comment?

hera_mc/geo_handling.py
bhazelton: Ok. Is there an error somewhere that will force you (not the random user trying to make this plot) to address it? Errors are much more effective at causing change than warnings.

david-deboer: I feel confident that a warning will suitably motivate me.

bhazelton: ok

david-deboer:
Untitled-1
The purple dashed line is from cm_dataview: "ants_by_day" and the black and green solid lines are form cm_dataview: "connected_by_day". It is a brute force slow step through time

PR #350, opened 2019-05-28
Csvoutput

david-deboer:
This gives the option for csv output for the hookup listing.  It is preparatory to improving the presentation of the data for commissioning.

### DIFF
--- a/hera_mc/cm_hookup.py
+++ b/hera_mc/cm_hookup.py
@@ -623,7 +623,7 @@ class Hookup(object):
         return s
 
     def show_hookup(self, hookup_dict, cols_to_show='all', state='full', ports=False, revs=False,
-                    file=None, output_format='ascii'):
+                    file_name=None, output_format='ascii'):
         """
         Print out the hookup table -- uses tabulate package.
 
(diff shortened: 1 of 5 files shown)


hera_mc/cm_utils.py
```diff
@@ -297,6 +297,18 @@ def html_table(headers, table):
     return s
 
 
+def csv_table(headers, table):
```
bhazelton: this function needs a docsting. Also, please do not use single letter variable names.

hera_mc/cm_hookup.py
```diff
@@ -634,8 +634,8 @@ def show_hookup(self, hookup_dict, cols_to_show='all', state='full', ports=False
         ports:  boolean to include ports or not
         revs:  boolean to include revisions letter or not
         state:  show the full hookups only, or all
-        file:  file to use, None goes to stdout
-        output_format:  set to html for the web-page version, or ascii
+        filename:  file name to use, None goes to stdout
+        output_format:  set to html for the web-page version, or table, or csv
```
bhazelton: what does the `table` format mean for file output?

david-deboer: It writes the formatted table to that file.

bhazelton: so it's a human-readable text file? Something explaining that needs to be in the docstring.

PR #353, opened 2019-05-31
Fix duplication check in cm_health

david-deboer:
The old version tested over a too limited test for duplicated connections.  This does it much more comprehensively.


david-deboer:
There was already duplicated data which tests the new and improved version.

bhazelton:
Ok, good. Are there tests on this branch that you have seen fail on master?

hera_mc/cm_health.py
```diff
@@ -15,15 +15,19 @@
...
     """
+    Checks to see if two intervals overlap.  Notionally for time, but will work
+    for any ints/floats.
+
     Parameter
     ---------
-    intervals:  gps_time intervals to test for overlap
-                format [[span_1_low, span_1_high], [span_2_low, span_2_high]]
-                if the high values are None, they get set to the max value + 100 sec
+    interval :  intervals to test for overlap
```
bhazelton: the type should be python types (list, str, int, ndarray....)

hera_mc/cm_health.py
```diff
@@ -15,15 +15,19 @@
...
+               format [[span_1_low, span_1_high], [span_2_low, span_2_high]]
+               if the high values are None, they get set to the max value + 100 sec
 
     Return
     ------
-    boolean:  True of overlap
+    boolean
+                True of overlap
```
bhazelton: What does this mean? I thought it meant it would return True if there was an overlap found. I find the word "of" confusing...

david-deboer:
The problem with the old version was that they passed on master.  With the new test, I found and corrected about 8 duplications (all but one from the parts_paper era).

PR #355, opened 2019-06-07
epoch time conversion for grafana

dannyjacobs:
I added this function to the site DB to make plotting easier on grafana


david-deboer:
I don't get why that definition is in Docs?  (I also don't understand how it works, but presumably you point grafana to it somehow?)

dannyjacobs:
This is just a database function.  It could potentially be added to the alchemy layer, for now I just wanted to note somewhere what I added to the site db.

bhazelton:
It's good to have some documentation on it, but I'd really like to see it added to alchemy if possible so it lives with the code. I just made an issue to track that.

PR #356, opened 2019-06-13
Improve correlator command script

bhazelton:
Better docstings, raise error if `acclen_spectra` isn't an integer multiple of 2048.

### DIFF
--- a/hera_mc/mc_session.py
+++ b/hera_mc/mc_session.py
@@ -2269,6 +2269,7 @@ class MCSession(Session):
             only applies if command is 'take_data': Length of time to take data for, in seconds
         acclen_spectra: integer
             only applies if command is 'take_data': Accumulation length in spectra.
+            Must be an integer divisible by 2048.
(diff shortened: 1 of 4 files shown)


hera_mc/mc_session.py
```diff
@@ -2377,6 +2378,11 @@ def correlator_control_command(self, command, starttime=None, duration=None,
             else:
                 if acclen_spectra != corr.DEFAULT_ACCLEN_SPECTRA:
                     warnings.warn('Using a non-standard acclen_spectra!')
+
+            if (not isinstance(acclen_spectra, (int, np.integer))
+                    or np.round(acclen_spectra / 2048.) * 2048 - acclen_spectra != 0):
```
jack-h: Is there a reason this can't just be (acclen_spectra % 2048 != 0)?

bhazelton: No, I'll change it.

PR #377, opened 2019-08-08
Multiple part start times

david-deboer:
The code couldn't handle a stopped part reappearing.  This branch allows that.  It also cleans up much of the low level parts stuff.  It adds test coverage for parts content in cm_partconnect.


bhazelton:
The numpy docstring style is here: https://numpydoc.readthedocs.io/en/latest/format.html

Note that the description of the parameter goes on the next line with an indent, while the type goes on the line with the name of the parameter. Also note that there's a space before and after the colon that separates the parameter name and the type.

bhazelton:
More details from a Slack exchange with Dave:
> I (a) change the existing record to the new start time and None end time, (b) I add a note to part_info with the previous information, so it is associated with that part.
> This wasn’t in there because it wasn’t intended to happen — a part is a part  However, if one does get “stopped” it is good to have a sensible way to restart it.  A few PAMs got stopped that are back in the system and need to get restarted.
>We could undo the stop time, but I like to collect any information we have, even if it wasn’t based on an intended process.

PR #383, opened 2019-08-22
Add ant autospectra

mkolopanis:
Added hook for the "autocorrelation" status now in HeraCorrCM dicts.
value is a list of floats, represented as a string (like histogram and histogram_bin_centers).

updated a few docstrings in correlator.py and mc_session.py


Docs/mc_definition.tex
```diff
@@ -615,6 +615,8 @@ \subsubsection{antenna\_status}
 eq\_coeffs & string & digital EQ coefficients for this antenna, used for keeping the bit occupancy in the
             correct range. list of floats (one per freq. channel) represented as a string. Note this these are
             not divided out anywhere in the DSP chain (!). \\\hline
+autocorrelation & string & the antenna pol autocorrelation as a function of frequency.
+                list of floats represented as a string  \\\hline
```
bhazelton: Thanks for adding this! Can you update the pdf as well?

hera_mc/correlator.py
```diff
@@ -653,27 +653,57 @@ class AntennaStatus(MCDeclarativeBase):
...
+        EM temperature sensor reading for this antenna in degrees Celsius
+    eq_coeffs : String
+        Digital EQ coefficients for this antenna, list of floats stored as a string
+    histogram_bin_centers : String
+        ADC histogram bin centers, list of ints stored as a string
+    histogram : String
+        ADC histogram counts, list of ints stored as a string
+    autocorrelation : String
+        Real valued autocorrelation value for the given antenna pol
+        list of floats stored as string
```
bhazelton: Wow, thanks!

bhazelton: How long are these lists generally?

mkolopanis: Truth be told I am not 100% sure, if they are the same length as the 'autos' from the redis database they will be `int(2048 // 4 * 3) = 1536` entries long.

bhazelton: ok. I think we'll just need to keep an eye on the database size. We might need to archive older values and remove them from the onsite db if it gets too big.

bhazelton:
@dannyjacobs do you have opinions about merging this now or waiting?

mkolopanis:
I can update the docstrings to make it very clear. Do we also want to re-label the entry to be like "snapspectra"?

dannyjacobs:
Yes, please rename the variables. 

Also it looks like the data are being stored in "antenna_status" as strings.  This strikes me as sub-optimal from a data volume point of view.  How much space does a spectrum take up and what is the projected total over, say 3 years?

mkolopanis:
I'm not sure as to the actual size, jack might know being that he wrote it into the hera_corr_cm module and I haven't been able to get my hands on them. Being an array, is there a better way to put them into a table?

dannyjacobs:
Its probably the raw spectrum size, 8k.

mkolopanis:
There are `Array` type columns, but that won't really help with size I bet. And I've notice there do not seem to be any instances of that.

bhazelton:
@mkolopanis I'm ok with array as long as it works properly. No reason not to use the right thing. I'll take a peek and Postgres and make sure it's kosher there. The string conversion thing might have been because of old issues.

bhazelton:
Oh, interesting. It's only postgres that allows them. So then the issue is, does it break our ability to use sqlite. @david-deboer I can't remember, does the sqlite option make all the tables? and if it does, will using an array column type break things?

mkolopanis:
We actually may pull this PR too. The more @dannyjacobs and I talk, we think the use case for these might exist without the need for M&C

mkolopanis:
We've decided to use the interface directly through `hera_corr_cm` for this.

PR #398, opened 2019-09-20
add get_snap_hostname_from_serial method

bhazelton:
@mkolopanis can you see if this does what you want?


mkolopanis:
Sorry this took so long. This does exactly what was envisioned for the function.

PR #404, opened 2019-10-09
update to use chart_studio for plotly ploting

mkolopanis:
plotly is moving to use `chart_studio` for the plotly call. We will need to install this package on the executing computer too. 
available on [conda with custom channel](https://anaconda.org/plotly/chart-studio) or pip
see https://plot.ly/python/getting-started-with-chart-studio/

### DIFF
--- a/hera_mc/autocorrelations.py
+++ b/hera_mc/autocorrelations.py
@@ -75,7 +75,8 @@ class Autocorrelations(MCDeclarativeBase):
 
 
 def plot_HERA_autocorrelations_for_plotly(session):
-    from plotly import graph_objs as go, plotly as py
+    from plotly import graph_objects as go
+    from chart_studio import plotly as py
 
     hera_ants = [9, 10, 20, 22, 31, 43, 53, 64, 65, 72, 80, 81, 88, 89, 96, 97, 104, 105, 112]
     # Fill in arrays from the DB.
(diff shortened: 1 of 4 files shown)


mkolopanis:
@bhazelton are these failures expected/common/understood? I am not too familiar with the `server_status` module and cannot see how the plotly import change could cause these

bhazelton:
It looks like I have a slightly unstable test due to a timing race condition. I'll make a PR to fix it, but I'll also try restarting the failed tests because they may come good.

hera_mc/autocorrelations.py
```diff
@@ -75,7 +75,8 @@ def __repr__(self):
 
 
 def plot_HERA_autocorrelations_for_plotly(session):
-    from plotly import graph_objs as go, plotly as py
+    from plotly import graph_objects as go
+    from chart_studio import plotly as py
 
     hera_ants = [9, 10, 20, 22, 31, 43, 53, 64, 65, 72, 80, 81, 88, 89, 96, 97, 104, 105, 112]
```
bhazelton: Is there a way we can get this out of config management rather than hardcoding it?

mkolopanis: oh probably. I didn't even look at that line honestly just went on replacing the plotly import. I'l see what I can do with these hard-coded antennas.

mkolopanis:
This iteration filters on anteannas the DB sees as "fully hooked up" not sure if it is ideal, but also I'm not sure how often this particular plot is used.

mkolopanis:
I'm not sure if these ever were covered, and now I added more uncovered lines to a plotting script :thinking:

PR #406, opened 2019-10-10
make sure server_status test doesn't fail due to small time differences

bhazelton:


bhazelton:
@mkolopanis I think this PR should fix the intermittent test failures you were seeing.

bhazelton:
They're floored gps seconds, so they should never be out by more than 1. They can be off by one because of small time differences.

PR #410, opened 2019-10-21
NOMERGE - Fix part/connection aspect along lines of hookup fix

david-deboer:
This simplifies the cm_handling for parts and connections using the "ActiveData" class put in under the hookup pull.  After approved, I'd like to test the branch on site before merging into Master.


david-deboer:
Can someone help to have itertools included for Python 2.7?

hera_mc/cm_handling.py
```diff
@@ -453,168 +112,173 @@ def get_part_from_hpnrev(self, hpn, rev):
...
+                    if key.upper().startswith(h_hpn):
+                        k_hpn, k_rev = cm_utils.split_part_key(key)
+                        all_hpn.append(k_hpn.upper())
+                        all_rev.append(h_rev)
+            match_list = cm_utils.match_list(all_hpn, all_rev, upper=True)
+        return match_list
+
+    def _set_ports(self, inp):
+        """
+        Handles 3 distinct phases of get_dossier:  initialize, get ports, set allowed
```
bhazelton: Since that these three phases do not share code and return immediately (rather than going on to another phase internally) shouldn't they be separate methods?

hera_mc/cm_dossier.py
```diff
@@ -4,15 +4,250 @@
 # Licensed under the 2-clause BSD license.
 
 """
-This contains the HookupEntry class which serves as a "dossier" for hookup entries.
-
+This contains the Entry classes which serves as a "dossier" for part entries,
+connections entries, and hookup entries.
```
bhazelton: This sounds like there should be 3 objects, but I only see 2. Am I missing something?

hera_mc/cm_utils.py
```diff
@@ -230,6 +233,19 @@ def listify(X):
     return [X]
 
 
+def match_list(a_obj, b_obj, upper=False):
+    if not isinstance(a_obj, list):
+        a_obj = [a_obj]
```
bhazelton: Do you only want to check for lists? If an array is passed then this will make a 1-element list that contains the array in that one element.

hera_mc/cm_utils.py
```diff
@@ -230,6 +233,19 @@ def listify(X):
     return [X]
 
 
+def match_list(a_obj, b_obj, upper=False):
```
bhazelton: Needs a docstring which should mention the trickiness you're doing below where you're replicating `b_obj` to have the same length as `a_obj` if `b_obj` isn't a list.

hera_mc/mc_session.py
```diff
@@ -2887,52 +2886,15 @@ def _get_node_snap_from_serial(self, snap_serial, session=None):
             return None, None
 
         if len(conn_dossier.keys()) > 1:
-            warnings.warn('Multiple active dossiers returned for snap serial '
-                          '{snn}. Setting node and snap location numbers to '
-                          'None'.format(snn=snap_serial))
+            warnings.warn('Multiple {} snaps were found, please specify the revision. '
+                          'Setting node and snap location numbers to None'
+                          .format(snap_serial))
```
bhazelton: There's no way to specify the revision to this method. We could add one, but this method is not generally going to be called by humans, it's going to be called by automated code that is just trying to resolve a snap serial number from the correlator redis into a node and snap location number. So I think we want the defaulting to give you the right information and I don't really see how the code could/should specify a revision.

Also, I find the change from `active dossiers` to `snaps` to be confusing. Presumably a single snap serial number doesn't resolve to several snaps, but if the database has issues it can resolve to several connections/dossiers/whatever-you-call-them.

If this can never happen because of the improvements you've put in place that make the database more stable, lets take it out. On the other hand, if it can happen but the way it happens is misrepresented in the test I wrote, we should fix the test to make it match a more real world scenario.

hera_mc/tests/test_correlator.py
```diff
@@ -1464,9 +1464,7 @@ def test_snap_status_errors(mcsession):
 
 
 def test_get_node_snap_from_serial_nodossier(mcsession):
-    node, snap_loc_num = checkWarnings(mcsession._get_node_snap_from_serial,
-                                       ['foo'],
-                                       message='No active dossiers')
+    node, snap_loc_num = checkWarnings(mcsession._get_node_snap_from_serial, ['foo'])
```
bhazelton: This is going backwards, you need to check that the message you get is the correct message.

hera_mc/tests/test_correlator.py
```diff
@@ -1484,79 +1482,35 @@ def test_get_node_snap_from_serial_multiple_locs(mcsession):
     connection.start_gpstime = 1230375618
     mcsession.add(connection)
     mcsession.commit()
-    node, snap_loc_num = checkWarnings(
-        mcsession._get_node_snap_from_serial, ['SNPD000703'],
-        message='Multiple snap location numbers returned')
-    assert node == 701
-    assert snap_loc_num is None
+    pytest.raises(ValueError, mcsession._get_node_snap_from_serial, ['SNPD000703'])
```
bhazelton: What error is this line testing? You should specify the text of the expected error not just that an error is raised. It looks like you took out the branch of the method that this was supposed to test. If you took it out because it's unreachable, we can just remove this test.

hera_mc/tests/test_correlator.py
```diff
@@ -1484,79 +1482,35 @@ def test_get_node_snap_from_serial_multiple_locs(mcsession):
...
-    connection.downstream_input_port = 'loc3'
-    connection.start_gpstime = 1230375718
+    connection.downstream_input_port = 'loc2'
+    connection.start_gpstime = 1230375618
     mcsession.add(connection)
     mcsession.commit()
-    node, snap_loc_num = checkWarnings(
-        mcsession._get_node_snap_from_serial, ['SNPD000703'],
-        message='Multiple node connections returned')
+    node, snap_loc_num = checkWarnings(mcsession._get_node_snap_from_serial, ['SNPD000703'])
```
bhazelton: You need to specify the warning message you expect -- that's the whole point of checkWarnings.

hera_mc/tests/test_correlator.py
```diff
@@ -1484,79 +1482,35 @@ def test_get_node_snap_from_serial_multiple_locs(mcsession):
...
-        mcsession._get_node_snap_from_serial, ['SNPD000703'],
-        message='Multiple node connections returned')
-    assert node is None
-    assert snap_loc_num is None
-
-
-def test_get_node_snap_from_serial_multiple_noderevs(mcsession):
-    """Test multiple node revisions."""
-    # add part for new revision
+def test_get_node_snap_from_serial_multiple_locs(mcsession):
```
bhazelton: This test has the same name as the test before. Make sure the name isn't a duplicate and matches what is being tested.

hera_mc/tests/test_parts.py
```diff
@@ -81,9 +79,10 @@ def test_update_part(parts, capsys):
     assert 'does not exist as a field' in captured.out.strip()
     data = [[parts.test_part, parts.test_rev, 'hpn_rev', 'Z']]
     cm_partconnect.update_part(parts.test_session, data)
-    dtq = Time('2017-07-01 01:00:00', scale='utc')
-    located = parts.cm_handle.get_part_dossier(
-        hpn=[parts.test_part], rev='Z', at_date=dtq, exact_match=True)
+    dtq = Time('2019-09-01 01:00:00', scale='utc')
+    located = parts.cm_handle.get_dossier(
+        hpn=[parts.test_part], rev=None, ports=None, at_date=dtq, exact_match=True)
+    print(located)
```
bhazelton: should this print statement be taken out?

hera_mc/cm_active.py
```diff
@@ -110,8 +111,9 @@ def load_connections(self, at_date=None):
         """
         Retrieves all active connections for a given at_date.
 
-        Sets this object's at_date and loads the active parts on to the object.
+        Loads the active parts onto the class and sets the class at_date.
         If at_date is None, the existing at_date on the object will be used.
+        This method will ValueError if a duplicate is found.
```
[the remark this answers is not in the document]
david-deboer: Done.

hera_mc/cm_handling.py
david-deboer: I went back and forth on this, and decided that since it was a private method it was ok to keep them bundled (which I preferred for future understandability, on my part at least).

hera_mc/cm_dossier.py
david-deboer: You are right - I missed this after updating from removing the whole confusing PartConnection thing.

hera_mc/cm_handling.py
bhazelton: ok.

hera_mc/cm_utils.py
david-deboer: For this and comment below I made quite a few changes that occurred while I was writing the docstrings plus added pytesting for new stuff in cm_utils

hera_mc/tests/test_correlator.py
david-deboer: How do you check a specified message?  Looks like ones uses `with pytest.raises`...?

bhazelton: Here's an example from a pyuvdata test:
```    with pytest.raises(ValueError) as cm:
        uvdata_data.uv_object.check()
    assert str(cm.value).startswith('All antennas in ant_1_array must be in antenna_numbers')
```

bhazelton: s=Sorry, formatting took a couple tries. The assert is **not** indented under the `with` statement.

david-deboer: Deleted this error. check since new code doesn't allow it.

hera_mc/tests/test_correlator.py
david-deboer: Renamed to incorporate info that multiple revs are checked.

hera_mc/tests/test_parts.py
david-deboer: Yes

hera_mc/tests/test_correlator.py
david-deboer: Added text check.

hera_mc/tests/test_parts.py
david-deboer: Yes

hera_mc/tests/test_correlator.py
david-deboer: I'm not entirely sure what this one is about.  I did change the name to better describe the test.

hera_mc/mc_session.py
david-deboer: Added rev to call.  Defaults to None.

I'm not sure what is meant by the `active dossiers` to `snaps`

david-deboer:
There is a pycodestyle error that I can't find to fix!

plaplant:
@david-deboer I've rebased this branch off of master after merging in #415. Please delete your local version.

PR #412, opened 2019-10-21
RTP Processing Record script

mkolopanis:
adds a script for creating RTPProcessRecords from the command line for new makeflow based rtp

### DIFF
--- a/scripts/add_rtp_process_record.py
+++ b/scripts/add_rtp_process_record.py
@@ -43,7 +43,7 @@ if __name__ == "__main__":
             obsid=obsid.gps,
             pipeline_list=args.pipeline_list,
             rtp_git_version=hera_opm_version_info['version'],
-            rtp_git_hash=hera_opm_version_info['git_hashs'],
+            rtp_git_hash=hera_opm_version_info['git_hash'],
             hera_qm_git_hash=hera_qm_version_info['version'],
             hera_qm_git_version=hera_qm_version_info['git_hash'],
             hera_cal_git_hash=hera_cal_version_info['version'],
--- a/scripts/add_rtp_process_record.py
+++ b/scripts/add_rtp_process_record.py
@@ -1,0 +1,54 @@
+#! /usr/bin/env python
+# -*- mode: python; coding: utf-8 -*-
+# Copyright 2019 the HERA Collaboration
+# Licensed under the 2-clause BSD license.
+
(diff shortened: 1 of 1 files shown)


scripts/add_rtp_process_record.py
```diff
@@ -0,0 +1,54 @@
...
+    obsid = Time(np.unique(uv.time_array)[0], scale='utc', format='jd')
+
+    db = mc.connect_to_mc_db(args)
+    with db.sessionmaker() as session:
+        session.add_rtp_process_record(
+            time=Time.now(),
+            obsid=obsid.gps,
+            pipeline_list=args.pipeline_list,
+            rtp_git_version=hera_opm_version_info['version'],
+            rtp_git_hash=hera_opm_version_info['git_hashs'],
```
plaplant: Small typo: dict key should be `'git_hash'`.

plaplant:
@bhazelton I know that hera_cal writes its version information into the calibration solution files, and I believe that hera_qm does something similar for metrics/flag files. (And pyuvdata obviously does if it writes files out.) However, given that right now, we're not writing calibration solutions or metrics files, I'm not sure where to get this data from. It's arguably not as important if we're not even doing heavy lifting with these repos, but I think this approach is okay.

What would you recommend as an improvement? Should each task print it to the log before it starts doing work?

mkolopanis:
I can definitely add pass throughs for all those hashes. However with the
current implementation of rtp I'm not sure it is feasible to have them
passed as arguments. It would result in either a lot of inline python calls
in the script that calls this or in tracking the hashes in every step of a
makeflow workflow since the environment needs to be activated at every
step.

Open to suggestions/ideaz

mkolopanis:
I've been thinking it might almost make more sense for the git hashes/etc to go in a `RTPTaskResourceRecord` so you know which code version it is associated with.

plaplant:
Perhaps, but I'm assuming our environment on site will be stable enough that we're not re-installing versions in the middle of the night. I think this is fine as is. @bhazelton how do you feel about it?

bhazelton:
I can live with it as long as it's only ever called from the same environment that the processing is done under.

mkolopanis:
It is meant to just be ran as part of the workflow for each file. No reason the environment _should_ change between steps. It might be possible to check for that the expected environment is actually activated in all the wrappers and exit if it is not?

PR #414, opened 2019-10-24
Add dependencies to install.md

bhazelton:
After one of my grad students struggled to get it installed (some of the packages give weird errors that do not reference their package names exactly, so it wasn't as easy as just adding each package as needed).


david-deboer:
This repo garnered the same invisible pycodestyle error as the other one...

plaplant:
The build seems to be failing because the IERS server is down for maintenance (for 6 months!), so some tests that rely on astropy time objects are raising errors. The astropy team has merged in a fix for this, but as of right now they have not released a new version to conda/PyPI. Hopefully they make a new release soon...

In the meantime, we should make the tests robust to not being able to download the IERS table. We do this on, e.g., pyuvdata already, so hopefully it's as simple as adapting the approach there.

bhazelton:
Unfortunately, the basics of the pyuvdata fix is already in this repo (see test/conftest.py) and it's still erroring. I will try to dig into this, but help would be very appreciated!

plaplant:
@bhazelton heads up, I've rebased this branch off of master after merging in #415.

PR #416, opened 2019-10-26
Remove default True to lsttrig command line option

jack-h:
Since it is a store_true flag, it should default
to False.


bhazelton:
@dannyjacobs ping on this

PR #419, opened 2019-10-29
Fix-partconn update

david-deboer:
This is the "fixed" part conn branch.


hera_mc/mc_session.py
```diff
@@ -2975,14 +2975,16 @@ def add_corr_snap_versions_from_corrcm(self, corr_snap_version_dict=None,
             self._insert_ignoring_duplicates(SNAPConfigVersion,
                                              snap_version_list)
 
-    def _get_node_snap_from_serial(self, snap_serial, session=None):
+    def _get_node_snap_from_serial(self, snap_serial, rev=None, session=None):
```
bhazelton: This still doesn't make sense to me (as I explained in the PR that you closed and replaced with this one). This internal function is used when we get a snap serial number from the correlator redis. I have no idea what to do with the revision number here, I'll never have one. Is it really possible to have multiple revisions plugged into the currently running system with the same serial number such that I would need to disambiguate them? I think that would be a mistake in cm, which should be caught as an error.

david-deboer: I see.  Yes - multiple revisions of the same part are allowed simultaneously.  This can be an exception, where I can take out the rev option and raise an error if it finds multiple revisions?

bhazelton: No, I definitely don't want an exception here because it would prevent recording anything into the database. What should happen is that a warning should be issues and the two returns should be none.

I do think it should cause an error in the CM code if you can have 2 revisions simultaneously active that have the same serial number. That seems like an obvious mistake. But I don't want it to cause an error here where it would prevent database logging of the snap status info.

bhazelton: Do multiple revisions of the same part share a serial number?!?

david-deboer: It will now set a warning and return Nones.

PR #421, opened 2019-10-31
Add load state to M&C

mkolopanis:
Trying to plumb through the load state from corr_cm into M&C. I may have missed a few spots, please check.

Also added the actual corr_cm noise_diode_enable/disable functions to acceptable correlator commands in addition to noise_diode_on/off but happy to remove if we only want the aliases (on/off)


hera_mc/correlator.py
```diff
@@ -38,6 +39,10 @@
                 'phase_switching_off': 'phase_switch_disable',
                 'noise_diode_on': 'noise_diode_enable',
                 'noise_diode_off': 'noise_diode_disable',
+                'noise_diode_enable': 'noise_diode_enable',
+                'noise_diode_disable': 'noise_diode_disable',
```
bhazelton: why are you adding these lines?

hera_mc/correlator.py
```diff
@@ -56,6 +61,18 @@
                      'noise_diode_off': {'state_type': 'noise_diode',
                                          'state': False,
                                          'allowed_when_recording': False},
+                     'noise_diode_enable': {'state_type': 'noise_diode',
+                                            'state': True,
+                                            'allowed_when_recording': False},
+                     'noise_diode_disable': {'state_type': 'noise_diode',
+                                             'state': False,
+                                             'allowed_when_recording': False},
```
bhazelton: And these?

hera_mc/correlator.py
mkolopanis: Ah yes happy to remove, but noise_diode_enable/disable are the actual corr_cm commands, and honestly I am more familiar with corr_cm than M&C so I tried to run noise_diode_enable once and had to look up the M&C command. 

So, I thought it might be nice to have the actual corr_cm commands but it _is_ just more to keep track of now.
I'm not opposed to them being different things in the end and just deleting these lines and the others you commented on.

mkolopanis:
@bhazelton  I think I have addressed all the comments from our last discussion, whenever you get a chance.

PR #424, opened 2019-11-01
Fix RTPProcessRecord script

mkolopanis:
casts the obsid in add_rtp_process_record to expected format. int(floor( time in gps seconds))


mkolopanis:
Also just have a question I think about what we want our expected behavior to be. Currently the script has `nargs='+'` for the filename format, but then assumes we only have 1 file. Should we update to loop through files, or just set `nargs=1`?

plaplant:
Oh never mind, I see your concern now. Yes, let's change `nargs=1` explicitly.

PR #427, opened 2019-11-05
New ant status

bhazelton:
Adds columns to antenna_status table:
- fem_switch (antenna, load, noise)
- fem_lna_power (is fem powered)
- fem_imu_theta (one of 2 tilt axes)
- fem_imu_phi (second tilt axis)


hera_mc/correlator.py
```diff
@@ -913,6 +927,10 @@ def create(cls, time, antenna_number, antenna_feed_pol, snap_hostname,
         if antenna_feed_pol not in ['e', 'n']:
             raise ValueError('antenna_feed_pol must be "e" or "n".')
 
+        if (fem_switch is not None
+                and fem_switch not in ['antenna', 'load', 'noise']):
```
mkolopanis: since corr_cm runs in python 3 there _shouldn't_ be any bytes vs string issue with this. I'll report back after maintenance day.

PR #428, opened 2019-11-05
Make megaraid script py3 compatible

plaplant:
The script that checks the megaraid status of the pots didn't work in python3. Now it does.


mkolopanis:
test run of this was successful

PR #431, opened 2019-11-05
update time_conversion doc

mkolopanis:
added an overloaded gps2epoch for bigint types. Should really look into making an sqlalchemy version so these could be re-added if we need automatically.


mkolopanis:
Yeah I'm not 100% sure _how_ to do that but once I do I'm more than willing to make it happen.

PR #433, opened 2019-11-07
Improve IERS handling for tests

### LINKED ISSUE #417: try IERS mirror in py3
Could be good to add a try the IERS_A_MIRROR in python3 if the initial IERS_A download fails. 
Something like the following:
```
 if six.PY3:
            # python 3 offers a mirror for the download url.
            try:
                iers.IERS.iers_table = iers.IERS_A.open(iers.IERS_A_URL_MIRROR)
                t1 = Time.now()
                t1.ut1
            except(urllib.error.URLError):
                iers.conf.auto_max_age = None
        else:
            iers.conf.auto_max_age = None
```

bhazelton:
Based on an improvement trialed recently in pyuvdata.

fixes #417

### DIFF
--- a/hera_mc/tests/conftest.py
+++ b/hera_mc/tests/conftest.py
@@ -19,13 +19,14 @@ test_db = None
 def setup_and_teardown_package():
     global test_db
 
-    # Try to download the latest IERS table. If the download succeeds, run a
-    # computation that requires the values, so they are cached for all future
-    # tests. If it fails, turn off auto downloading for the tests and turn it
-    # back on once all tests are completed (done by extending auto_max_age).
+    # Do a calculation that requires a current IERS table. This will trigger
+    # automatic downloading of the IERS table if needed, including trying the
+    # mirror site in python 3 (but won't redownload if a current one exists).
+    # If there's not a current IERS table and it can't be downloaded, turn off
+    # auto downloading for the tests and turn it back on once all tests are
+    # completed (done by extending auto_max_age).
     # Also, the checkWarnings function will ignore IERS-related warnings.
     try:
-        iers.IERS_A.open(iers.IERS_A_URL)
         t1 = Time.now()
         t1.ut1
     except(urllib.error.URLError, IOError):
--- a/hera_mc/tests/conftest.py
+++ b/hera_mc/tests/conftest.py
@@ -29,7 +29,7 @@ def setup_and_teardown_package():
     try:
         t1 = Time.now()
         t1.ut1
-    except(urllib.error.URLError, IOError):
+    except(urllib.error.URLError, IOError, iers.IERSRangeError):
         iers.conf.auto_max_age = None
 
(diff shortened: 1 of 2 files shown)


bhazelton:
@mkolopanis There was an error on this branch because IERS wasn't available. I couldn't figure out why it was happening so I added some print statements and now the error is gone. Could you test it with your setup?

mkolopanis:
I can try!

mkolopanis:
I'm have a _lot_ of trouble getting psql set up in this docker image, it just does not want to run for the default user

bhazelton:
Oh, we had a docker container a long time ago that we got this working in (back when we were trying to emulate the onsite computing environment in docker). Let me see if I can dig anything up that might be useful.

bhazelton:
I'm not at all sure this will help because it's so old, but Peter wrote it up so it might: https://github.com/HERA-Team/hera-images

mkolopanis:
did finally get it running in the miniconda docker, seems to work fine :thinking: not sure why it errored before even, obviousy you weren't either.

bhazelton:
hmm. I'm glad it passed but worried about it coming back when IERS fails again. I guess I could increase the tolerance further to add some protection. What do you think?

mkolopanis:
Increasing the tolerance to a still sane level isn't a bad idea (and we even do that when the auto_max_age is none), but it would be nice to know _how_ the issue came up in the first place. I actually just had it occur, but it was a time when the iers table downloaded. The difference in LST is the size of when we expect to not have the iers table

bhazelton:
Oh that's weird. The funny thing is that when I saw the error, it looked to me it was with IERS down and from the printed info it seemed like the error was smaller than the 10 arc seconds that it was allowed, so I was confused and added the print statements. What was the size of the error you just saw with IERS present?

mkolopanis:
Yeah I reran with capture off, downloaded IERS table from the ftp mirror, saw:
```
hera_mc/tests/test_utils.py .tolerance (in arcseconds): 1e-08
lst diff (in arcseconds): -9.999999999998366
```

bhazelton:
Oh, I just got the same thing on my machine. Maybe I need to better understand what's going on with this test. (Danny wrote it not me).

mkolopanis:
If i change the date in the test to 2019 the error goes away. I wonder if there is an extrapolation issue back to 2015?

bhazelton:
I guess there could be. Maybe we should pick the date based on whether `iers.conf.auto_max_age` is `None`, but that feels hacky. I wish I understood IERS better.

mkolopanis:
I have just tried 2015, 2017, 2018 and all fail within 10arcseconds even with IERS. Does seem very odd

bhazelton:
Ok, I’ll dig into this more and try to understand what’s going on.

bhazelton:
Oh! At the top of the test, the bin size is set to 10 arcseconds. They must be falling just on either side of a bin edge.

PR #439, opened 2019-11-22
fixed connection issue and removed user-run check

david-deboer:
This had a subtle and vexing error when reading the active data multiple times.  It also had a time-boundary problem for active parts.  This was found because I have been adding more comprehensive checking.  I've also removed the user-run checks in the interest to make the code more maintainable (there was no remaining reason to keep the check flag within hookup.py).

### DIFF
--- a/hera_mc/cm_active.py
+++ b/hera_mc/cm_active.py
@@ -23,7 +23,6 @@ class ActiveData:
     ----------
     at_date : str, int, float, Time, datetime
     """
-    close_enough = 2.0  # Seconds within which the dates are close enough
 
     def __init__(self, session=None, at_date='now'):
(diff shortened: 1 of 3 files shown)


bhazelton:
Is there no need to ever run this check anymore?

david-deboer:
Yes, but I've pulled checking out of here and have a more comprehensive process.  I ended spending several hours yesterday checking and made some corrections to the database (this was when I found this error).

bhazelton:
Ok. Where is the comprehensive process? I only see code removing checking.

david-deboer:
I've moved it out of this repo to the cm_db_updates repo.  I do a lot of checking there when I update the repo.

PR #440, opened 2019-11-22
Cminfo redux:  Updates the cminfo and fully connected station to new faster version

david-deboer:
The cm_sysutils methods for get_fully_connected_XXX (and therefore the cminfo method) was very inefficient and unnecessarily looped several times over stations.  Being slow was hampering the correlator.  This fixes that.  Two interfaces changed:

get_all_fully_connected_at_date --> get_connected_stations
the object lost the 'started'/'stopped' attributes for a more general 'timing' attribute

I've changed instances in the repo, but I don't know if any other repo uses these (although I strongly doubt that they do).

### DIFF
--- a/hera_mc/autocorrelations.py
+++ b/hera_mc/autocorrelations.py
@@ -82,7 +82,7 @@ def plot_HERA_autocorrelations_for_plotly(session, offline_testing=False):
     from chart_studio import plotly as py
 
     hsession = cm_sysutils.Handling(session)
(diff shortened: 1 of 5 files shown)


bhazelton:
@dannyjacobs @mkolopanis This looks like it could be a breaking change if you call these methods in any of your code.

mkolopanis:
thanks for the heads up, definitely is a breaking change for some of our code. I'll make the adjustments after it is merged.

bhazelton:
@mkolopanis it's merged.

PR #441, opened 2019-11-22
Use flake8 for linting instead of pycodestyle

bhazelton:
Flake8 catches more things, including things that cause real bugs. It's a superset of pycodestyle + other things.


hera_mc/geo_handling.py
```diff
@@ -321,7 +321,9 @@ def get_ants_installed_since(self, query_date, station_types_to_check='all'):
             if a.station_type_name.lower() in station_types_to_check:
                 found_stations.append(a)
                 if self.fp_out is not None and not self.testing:
-                    self.fp_out.write('{:6} {:.2f} {:.2f} {:.4f} {:.4f}\n'.format(station_name, a.easting, a.northing, a.lon, a.lat))
+                    self.fp_out.write(
+                        '{:6} {:.2f} {:.2f} {:.4f} {:.4f}\n'
+                        .format(a.station_name, a.easting, a.northing, a.lon, a.lat))
```
bhazelton: I think this was an actual bug (`station_name` was never defined). I made a guess at how to fix it, but since it wasn't caught by tests before, I'm not confident in the fix. @david-deboer can you check that this is right and figure out why it wasn't causing tests to fail before?

david-deboer: Testing didn't catch it because the write fp_out was not set.  Your fix would fix it, but I actually tend to think we should remove the file write option (but that should be a different PR).  For now, let's go with your fix and I'll look into a future PR.

bhazelton: do you want to make an issue?

bhazelton:
we've been moving to flake8 in a number of our repos (RASG & HERA) because it's better. It includes everything caught by pycodestyle (it imports pycodestyle) but also catches things that are more important because they can lead to bugs, like variables that are defined but never used. In fact I found one critical pyuvdata bug by going to flake8 -- a variable was defined but not used when it should have been.

For Atom, you'll want to switch to using the `linter-flake8` package. You don't need to set any preferences on that package because our preferences are set in the setup.cfg file that's updated in this PR.

hera_mc/geo_handling.py
david-deboer: Made issue

david-deboer:
I see that most of the new uncovered stuff relates to Python 2 (when should we officially stop supporting?), but the qm.py and version.py "misses" could be covered.

bhazelton:
python2 support may have to remain for awhile, depending on the correlator interface. We're actively running that down.

bhazelton:
I think the misses are actually a mistake, caused by coveralls not merging the python 2 & 3 reports.

david-deboer:
But not the qm.py and version.py ones I don't think.

david-deboer:
When we post this, we should make a more general announcement about flake8, so that others can configure their editors if desired.

ci/hera_mc_test.yaml
```diff
@@ -0,0 +1,31 @@
...
+  - conda-forge
+  - plotly
+dependencies:
+  - alembic
+  - astropy
+  - cartopy
+  - chart-studio
+  - coveralls
+  - flake8
+  - flake8-docstrings
```
mkolopanis: I know you _can_ get this from conda, but it feels weird to me to get some flake8 plugins from conda and some from pip.

hera_mc/tests/test_autocorrelations.py
```diff
@@ -5,22 +5,9 @@
...
-
 import pytest
-import yaml
-import numpy as np
-from astropy.time import Time, TimeDelta
-
-from hera_mc import mc, autocorrelations, cm_sysutils
-import hera_mc.correlator as corr
-from hera_mc.data import DATA_PATH
-from ..tests import onsite, checkWarnings
```
mkolopanis: that's a _lot_ of unused imports

scripts/mc_log_autocorrelations.py
```diff
@@ -14,13 +14,17 @@
...
 from hera_mc import autocorrelations, mc
 
 
 # Yay hardcoded config settings:
 
 redis_host = 'redishost'
-antnums = xrange(128)
+
+sm.range
+antnums = sm.range(128)
```
mkolopanis: do we want this rogue `sm.range` call here?
also this will make antnums an iterator won't it? do we want `list(sm.range(128))`

bhazelton: yes on both counts. Thanks!

ci/hera_mc_test.yaml
bhazelton: ok. I can change it but it doesn't seem weird to me. I get everything I can from conda and only use pip for what I can't get from conda.

hera_mc/cm_active.py
```diff
@@ -73,9 +73,10 @@ def load_parts(self, at_date=None):
         at_date = cm_utils.get_astropytime(at_date)
         gps_time = self.set_times(at_date)
         self.parts = {}
-        for prt in self.session.query(partconn.Parts).filter((partconn.Parts.start_gpstime <= gps_time)
-                                                             & ((partconn.Parts.stop_gpstime >= gps_time)
-                                                             | (partconn.Parts.stop_gpstime == None))):  # noqa
+        for prt in self.session.query(partconn.Parts).filter(
+            (partconn.Parts.start_gpstime <= gps_time)
+            & ((partconn.Parts.stop_gpstime >= gps_time)
+               | (partconn.Parts.stop_gpstime == None))):  # noqa
```
mkolopanis: This must be due to the max line-length change, but I personally find it very difficult to discern where `for` statement ends. Could we move the `):` to a new line perhaps?

hera_mc/cm_active.py
```diff
@@ -106,14 +107,17 @@ def load_connections(self, at_date=None):
...
         self.connections = {'up': {}, 'down': {}}
         check_keys = {'up': [], 'down': []}
-        for cnn in self.session.query(partconn.Connections).filter((partconn.Connections.start_gpstime <= gps_time)
-                                                                   & ((partconn.Connections.stop_gpstime > gps_time)
-                                                                   | (partconn.Connections.stop_gpstime == None))):  # noqa
-            chk = cm_utils.make_part_key(cnn.upstream_part, cnn.up_part_rev, cnn.upstream_output_port)
+        for cnn in self.session.query(partconn.Connections).filter(
+            (partconn.Connections.start_gpstime <= gps_time)
+            & ((partconn.Connections.stop_gpstime > gps_time)
+               | (partconn.Connections.stop_gpstime == None))):  # noqa
```
mkolopanis: same comment regarding the `):` I'm not adamant about it, but would like some discussion at least.

hera_mc/cm_active.py
```diff
@@ -195,7 +201,8 @@ def load_geo(self, at_date=None):
         at_date = cm_utils.get_astropytime(at_date)
         gps_time = self.set_times(at_date)
         self.geo = {}
-        for ageo in self.session.query(geo_location.GeoLocation).filter(geo_location.GeoLocation.created_gpstime <= gps_time):
+        for ageo in self.session.query(geo_location.GeoLocation).filter(
+                geo_location.GeoLocation.created_gpstime <= gps_time):
```
mkolopanis: this one is not as bad, but up to your discretion.

hera_mc/cm_handling.py
```diff
@@ -318,7 +319,8 @@ def get_specific_connection(self, cobj, at_date=None):
         fnd = []
         for conn in self.session.query(partconn.Connections).filter(
                 (func.upper(partconn.Connections.upstream_part) == cobj.upstream_part.upper())
-                & (func.upper(partconn.Connections.downstream_part) == cobj.downstream_part.upper())):
+                & (func.upper(partconn.Connections.downstream_part)
+                   == cobj.downstream_part.upper())):
```
mkolopanis: marking another

hera_mc/cm_dossier.py
```diff
@@ -199,11 +201,13 @@ def table_row(self, columns, ports):
...
             for up, down in conns:
-                if up is None or up.upstream_output_port.upper() in ports or up.downstream_input_port.upper() in ports:
+                if (up is None or up.upstream_output_port.upper() in ports
+                        or up.downstream_input_port.upper() in ports):
                     new_up.append(up)
                 else:
                     new_up.append(None)
-                if down is None or down.upstream_output_port.upper() in ports or down.downstream_input_port.upper() in ports:
+                if (down is None or down.upstream_output_port.upper() in ports
+                        or down.downstream_input_port.upper() in ports):
```
mkolopanis: two other minor ones, at your discretion.

ci/hera_mc_test.yaml
mkolopanis: I used to be like this too, then i installed `black` through conda-forge and it was broken. I'm not dead set on the change, maybe just being grumpy

hera_mc/cm_active.py
bhazelton: sure. I will probably get used to that with black. Still is hard for me to read, but we're going that way it seems.

hera_mc/cm_active.py
bhazelton: There appears to be no indent that flake8 will accept for the closing bracket on the next line for this one.

hera_mc/cm_handling.py
bhazelton: flake8 won't let me do this one either. I can't figure out why.

bhazelton: Oh! I know why it worked on the earlier one -- because we have a `# noqa` on that line because sql requires `==` no `is`.

bhazelton: Given that I can't do it on the others, I'm not going to do it on the earlier ones. You're welcome to try and see if you can come up with something aesthetically pleasing to you that flake8 will accept.

bhazelton:
@mkolopanis can you figure out why coverage has dropped in `qm.py`? I see that something's lost coverage but I can't see how the changes I made would cause that. Maybe `hera_qm` has been updated since the run that this PR is being compared to?

hera_mc/cm_handling.py
mkolopanis: That's fine :smile: I'm not super adamant about it anyway.

mkolopanis:
Interesting. I think I see the test that  _should_ catch those lines but you're right it was not touched in this PR at all. Could add a test in there though. I feel like it should have come up in the pytest pr

hera_mc/cm_handling.py
bhazelton: update: I figured out how to do what flake8 and you wanted for this one and the above ones, but not for the ones below.

bhazelton:
I just checked all the files that coveralls reports as having lost coverage in any way against the codecov build. It turns out that codecov does not report coverage on the `__init__.py` file (a known bug) which is concerning, but other than that, all the things coveralls reports as having changed match between the codecov build on this branch and the coveralls master branch build. So they appear to be spurious and are making me unhappy with coveralls.

mkolopanis:
Huh very confusing. I'm willing to sign off on this branch again with that information, and then we can move to codecov if that is desired.

mkolopanis:
but I've also had a considerable investment with github actions, so if we need a third person now that makes sense.

bhazelton:
@david-deboer how do you feel about merging this PR? I'm willing to wait until you get yours in if you'd like.

david-deboer:
I don't have a strong feel - I'm interested in the easiest for me... ;-)

PR #444, opened 2019-11-26
Fixes script error on viewing revisions.

david-deboer:
This PR fixes an error in the revision viewing script, but as a consequence led to "unifying" revision stuff a bit.  WIP since need to add a few tests, but like the online coverage viewer.  Will probably have some merge conflicts with the flake8 PR(?)


bhazelton:
Do you really want to be adding and dropping columns in Alembic?

david-deboer:
I renamed a column that was no longer appropriately titled

bhazelton:
you can do a rename in Alembic, which is better than dropping and remaking the column.

bhazelton:
https://alembic.sqlalchemy.org/en/latest/ops.html#alembic.operations.Operations.alter_column

hera_mc/cm_active.py
```diff
@@ -205,23 +205,42 @@ def load_geo(self, at_date=None):
...
         """
-        Returns a list of active revisions for a given hpn.
+        Returns a list of active revisions for given hpn.
 
         Parameters
         ----------
-        hpn : str
-            A HERA part number (not a list)
+        hpn : str or list
+            A HERA part number, or list of part numbers or fractions thereof.
```
bhazelton: this should specify it's the beginning that's matched (startswith) not any "fraction" of the part.

hera_mc/cm_active.py
```diff
@@ -205,23 +205,42 @@ def load_geo(self, at_date=None):
...
-                hpn_revs.append(part.hpn_rev)
-        return hpn_revs
+        from argparse import Namespace
+        hpn = [x.upper() for x in cm_utils.listify(hpn)]
+        rev_dict = {}
+        for hloop in hpn:
+            rev_dict[hloop] = {}
+            for part in self.parts.values():
+                phup = part.hpn.upper()
+                use_this_one = (phup == hloop) if exact_match else phup.startswith(hloop)
```
bhazelton: what if multiple parts start with the specified string?

hera_mc/cm_active.py
david-deboer: Done.

hera_mc/cm_active.py
david-deboer: If there are multiple parts, it lists them all

alembic/versions/b022867d09e3_changed_library_file_to_reference_in_.py
```diff
@@ -0,0 +1,28 @@
...
+
+# revision identifiers, used by Alembic.
+revision = 'b022867d09e3'
+down_revision = 'e83aa47e530b'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
```
bhazelton: please remove these comments about auto generation. They aren't true, you haven't kept any of the autogenerated code.

hera_mc/cm_active.py
bhazelton: ok. maybe clarify that in the docstring?

david-deboer: Done

hera_mc/cm_active.py
```diff
@@ -205,23 +205,43 @@ def load_geo(self, at_date=None):
             key = cm_utils.make_part_key(ageo.station_name, None)
             self.geo[key] = ageo
 
-    def revs(self, hpn):
+    def revs(self, hpn, exact_match=False):
         """
-        Returns a list of active revisions for a given hpn.
+        Returns a list of active revisions for the provided hpn list.  The returned
+        list is the concatentated set of revisions found for the provided list.
```
bhazelton: "And can include more than one entry for each input entry if `exact_match` is not set and multiple parts start with the provided input entry". Right?

david-deboer: No - it can provide more than one entry even if there is only one hpn supplied, since there may be more than one active revision.

bhazelton: what if the hpn supplied is incomplete (just the first few characters) and can be matched to multiple parts?

david-deboer: It provides the concatenation of all.  This is the exact purpose of it.  If I want to know what revisions of a particular part are in the system I query on hpn='PAM' and it tells me all of the relevant revisions.

bhazelton: That makes perfect sense, but that is not clear to me at least from the current docstring. I read the docstring and think you get one active revision per item in the input list.

david-deboer: It has to check all of them to check for all active revisions.  I can also check on a single part etc.

hera_mc/cm_active.py
```diff
@@ -205,23 +205,48 @@ def load_geo(self, at_date=None):
...
+                if use_this_one:
+                    prup = part.hpn_rev.upper()
+                    rev_dict[hloop].setdefault(prup, Namespace(hpn=hloop, rev=prup, number=0,
+                                               started=part.start_gpstime, ended=part.stop_gpstime))
+                    rev_dict[hloop][prup].number += 1
+                    if part.start_gpstime < rev_dict[hloop][prup].started:
+                        rev_dict[hloop][prup].started = part.start_gpstime
+                    if rev_dict[hloop][prup].ended is not None:
+                        if part.stop_gpstime is None or part.stop_gpstime > rev_dict[hloop][prup].ended:
+                            rev_dict[hloop][prup].ended = part.stop_gpstime
```
bhazelton: Looks like lines 244, 246 & 247 aren't covered by tests.

hera_mc/cm_handling.py
```diff
@@ -235,19 +237,30 @@ def get_dossier(self, hpn, rev=None, at_date='now', notes_start_date='<', exact_
...
-        active.load_info()
-        active.load_geo()
+        if active is None:
+            active = cm_active.ActiveData(self.session, at_date=at_date)
+        elif at_date is not None:
+            date_diff = abs(at_date - active.at_date).sec
+            if date_diff > 1.0:
+                raise ValueError("Supplied date and active date do not agree ({}sec)".format(date_diff))
+        else:
+            at_date = active.at_date
```
bhazelton: These lines (242-247) aren't covered by tests.

hera_mc/cm_active.py
david-deboer: Testing these lines would rely on the order of stepping through dictionary keys.

hera_mc/cm_handling.py
david-deboer: Done

hera_mc/cm_active.py
bhazelton: you could use an ordered dict for the test I guess.

david-deboer: Given simplicity and position/use of the snippet I'd rather not introduce a whole other concept.

mkolopanis: I agree it would be very nice to test these lines. I think I can see a way forward for it too. 

We only call this revs issue in one test in `test_parts`. I would argue to make `active.parts` and ordered dict _only in the test_. That way we can control the order of the revisions to account for these fringe cases and it does not require any bending over backwards in the actual code.

We can call `active.parts = OrderedDict( self.parts ordered the way we want it)`

bhazelton:
woohoo! 100% diff coverage! Thanks @mkolopanis !

PR #445, opened 2019-12-06
Add support for white rabbit monitoring.

bhazelton:


bhazelton:
This has been tested on site and it works.

bhazelton:
@mkolopanis where are we on getting rid of python 2? There's some awful unicode error that I'll debug if I have to but would rather not...

mkolopanis:
Uh I don't think we've started on this repo. If `hera_mc` doesn't run on any node but `qmaster` then we should be good to go whenever. But if it runs on an intra-correlator node I'm not sure we can ever _stop_ supporting py2.

bhazelton:
I just made python 2.7 conda environment on my machine and ran this and did not get the error. @mkolopanis are you willing to try on your machine to see if its a linux thing?

bhazelton:
@dannyjacobs I can't leave that table up when we're on the master branch because then the db is inconsistent with the code.

bhazelton:
@dannyjacobs @david-deboer This is passing now if you'd like to take a look

PR #446, opened 2019-12-06
Remove support for python 2

bhazelton:
Waiting on confirmation that the correlator is ready for this.


bhazelton:
Yes, it is hard to review because it touches nearly every file in the repo. I'm sorry!

Remember, all this code is already being tested on python 3, so any integer divisions etc are already tested. The `from __future__ import...` lines were making python 2 behave more like python 3 (importing from the future). So I removed them because they only affected python 2 and I don't expect any problems there.

I replaced all calls to six with what six used for python 3 under the hood, so those should behave exactly as before under python 3. We really only needed six because it gave us a way to write code that was compatible with both.

In utils I removed the `_reraise_context` function which was a way of achieving exception chaining in python 2. We don't need it in python 3 because that's built into python 3. In test_utils I removed the associated test for `_reraise_context`. That test was the only test that needed pytest to be imported, so when I took out the test, pytest was an unused import.

PR #449, opened 2019-12-13
Sorthookup

david-deboer:
This allows sorting of the hookup output, which is invaluable for comparing to e.g. the config googlesheet.


hera_mc/tests/test_sysutils.py
```diff
@@ -56,7 +56,7 @@ def test_other_hookup(sys_handle, mcsession, capsys):
     hu = hookup.get_hookup('HH', 'all', at_date='now', exact_match=True, use_cache=True)
     assert len(hu) == 0
     hu = hookup.get_hookup('cache', pol='all', hookup_type='parts_hera')
-    out = hookup.show_hookup(hu, state='all', output_format='csv')
+    out = hookup.show_hookup(hu, state='all', sortby='station', output_format='csv')
     assert '1230375618' in out
```
bhazelton: It doesn't look like this actually tests that the sort ordering worked.

david-deboer: Modified to verify expected order.

PR #450, opened 2019-12-16
Hookup methods - this allows for "methods" to pick sets of antennas

david-deboer:
The only actual code changes are those lines in cm_hookup that involve _proc_hpnlist, the other changes are reorganizing and improving docstrings.  Since I really reorganized it so that I can more readily navigate, it looks like there are many changes.  I pulled stuff out needed for the "methods" that had been hiding in a script and made a new geo_sysdef.py file.


hera_mc/cm_hookup.py
```diff
@@ -130,13 +118,20 @@ def get_hookup_from_db(self, hpn, pol, at_date, exact_match=False, hookup_type=N
         Parameters
         -----------
         hpn : str, list
-            List/string of input hera part number(s) (whole or first part thereof)
+            List/string of input hera part number(s) (whole or 'startswith')
+            If string
+                - 'cache' returns the entire cache file
```
bhazelton: is 'cache' actually allowed in this method? It looks like this docstring was copied from `get_hookup` but that one explicitly handles 'cache' and this one doesn't.

Also, while you're re-arranging code, any function or method that is called by another function or method should come **before** the function or method that is calling it. So these two methods should be swapped.

hera_mc/cm_sysdef.py
```diff
@@ -295,3 +295,18 @@ def get_ports(self, pol, part_type):
                     elif port[0].lower() == pol[0].lower():
                         port_dict[dir].append(port.upper())
         return port_dict
+
+    def node(self, node_nums):
```
bhazelton: Is this also just re-arranged from somewhere else or is it new code? I don't see it anywhere else. It needs a docstring in either case.

hera_mc/geo_sysdef.py
```diff
@@ -0,0 +1,72 @@
...
+                       266, 267, 268, 269, 270, 271, 272, 273, 274, 275, 276,
+                       281, 282, 283, 284, 285, 286, 287, 288, 289, 290, 291,
+                       295, 296, 297, 298, 299, 300, 301, 302, 303, 304, 305,
+                       308, 309, 310, 311, 312, 313, 314, 315, 316, 317, 318],
+          'heraringa': [325, 326, 327, 330, 331, 334, 335, 338, 339, 342, 343, 344],
+          'heraringb': [320, 321, 322, 323, 324, 328, 329, 332, 333, 336, 337,
+                        340, 341, 345, 346, 347, 348, 349]}
+
+
+def read_nodes():
```
bhazelton: this function needs a docstring.

hera_mc/geo_sysdef.py
```diff
@@ -0,0 +1,72 @@
...
+            node_E = float(line.split(':')[1].split(',')[0])
+            node_N = float(line.split(':')[1].split(',')[1])
+            ants = line.split(':')[2].split(',')
+            ants = [int(x) for x in ants]
+            nodes[node_num] = {'E': node_E, 'N': node_N,
+                               'elevation': default_elevation, 'ants': ants}
+    return nodes
+
+
+def read_antennas():
```
bhazelton: this one also needs a docstring.

hera_mc/cm_hookup.py
david-deboer: Good catch!  I've swapped the order.

hera_mc/cm_sysdef.py
david-deboer: This is the new code for this PR.  I've added a docstring

hera_mc/geo_sysdef.py
david-deboer: I've added docstrings

PR #453, opened 2019-12-18
GitHub redis

mkolopanis:
Adds the redis service to github for additional code validation. Makes new decorator `requires_redis` many `onsite` tests moved to `requires_redis` instead.


mkolopanis:
This will require a new dump file every time a feature is added. Should I add a section to the readme?

bhazelton:
This is huge! Thank you @mkolopanis! I think adding something to the readme might be nice.

mkolopanis:
I have added a small section in the readme, please let me know if there is more you would like.

PR #454, opened 2019-12-18
Adds new apriori Statuses

### LINKED ISSUE #448: Redefine apriori status values
A discussion at SALF brought up the idea we may need to redefine the known values for the apriori antenna status table. I think the decision was a _new_ discussion needs to happen to find a way forward for this table.

@bhazelton @nimarazavi @dannyjacobs

mkolopanis:
Adds statuses and closes #448 
Keeps old arpiori status values for backwards compatibility but sends a DeprecationWarning if one is set.

### DIFF
--- a/scripts/update_apriori.py
+++ b/scripts/update_apriori.py
@@ -10,11 +10,26 @@ Script to update the apriori antenna status.
 from hera_mc import mc, cm_utils, cm_partconnect
 
 
-if __name__ == '__main__':
+if __name__ == "__main__":
     parser = mc.get_mc_argument_parser()
-    parser.add_argument('-p', '--hpn', help="HERA part number")
-    parser.add_argument('-s', '--status', help="New apriori status.",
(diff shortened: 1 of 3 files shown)


scripts/update_apriori.py
```diff
@@ -10,11 +10,26 @@
...
+            "dish_maintenance",
+            "dish_ok",
+            "RF_maintenance",
+            "RF_ok",
+            "digital_maintenance",
+            "digital_ok",
+            "passed_checks",
+            "needs_checking",
+            "known_bad",
+            "not_connected",
```
bhazelton: Do we want to make the deprecated values easy to add via a script?

david-deboer: In scripts/update_apriori.py, the old options are still there (I guess you get a deprecated message later), but the new versions aren't there.  Also, I think that we should include underscores in the new values.

PR #459, opened 2020-01-21
Add details about what environments to update when tables are updated.

bhazelton:


bhazelton:
@david-deboer can you check this? It's what we came up with on the telecon yesterday but wanted to make sure you didn't know of any we're missing.

bhazelton:
Thanks. I think we established on Monday that hera-snap-head shares a file system with hera-corr-head, so they don't both need to be updated. Is that right @dannyjacobs @dgorthi ?

bhazelton:
Danny has established that `hera-snap-head` does not share a file system with `hera-corr-head`. So the next question is: does `hera-snap-head` need to have hera_mc installed? @dannyjacobs @dgorthi @david-deboer

bhazelton:
@dgorthi confirmed on slack that `hera-snap-head` needs to have hera_mc installed.

PR #479, opened 2020-03-26
Node hookup

david-deboer:
This PR adds a "node-view" and cleans up/removes some confusing aspects of ports.


hera_mc/cm_hookup.py
```diff
@@ -537,7 +508,7 @@ def _get_connection(self, current):
         options = list(self.active.connections[current.direction][current.key].keys())
         try:
             current.type = self.active.parts[current.key].hptype
-        except KeyError:
+        except KeyError:  # pragma: no cover
```
bhazelton, line 511: What is the use case where this can happen? Why not test it?

hera_mc/cm_utils.py
```diff
@@ -658,6 +665,20 @@ def csv_table(headers, table):
     return s_table
 
 
+def table_table(headers, table_data, output_format=None):
```
bhazelton, line 668: That's a pretty confusing function name.

hera_mc/cm_active.py
```diff
@@ -32,6 +32,7 @@ def __init__(self, session=None, at_date='now'):
         self.info = None
         self.apriori = None
         self.geo = None
+        self.pytest_param = False
```
bhazelton, line 35: Putting test-only activated code like this in the real code feels a bit clunky. Is this really the best way?

hera_mc/cm_sysutils.py
```diff
@@ -424,3 +424,299 @@ def get_apriori_antenna_status_for_rtp(self, status, at_date='now'):
...
+    ant_num = cm_utils.listify(ant_num)
+    ant_node = {}
+    for pn in ant_num:
+        pnint = cm_utils.peel_key(str(pn), 'NPR')[0]
+        ant_node[pnint] = [_find_ant_node(pnint, na_from_file)]
+        ant_node[pnint].append(_find_ant_node(pnint, na_from_hookup))
+    return ant_node
+
+
+def print_which_node(ant_node):
```
bhazelton, line 535: I find this function name confusing -- it sounds like it prints something but it doesn't, it returns a formatted string.

Docs/mc_definition.tex
```diff
@@ -146,12 +146,10 @@ \subsection{Signal Chain Hookup Connections}
 \begin{figure}[H]
 \includegraphics[width=0.8\textwidth]{hookup.pdf}
 \centering
-\caption{Block diagram of hookup.  The line labels indicate the port names/connections.}
+\caption{Block diagram of hookup.  The line labels indicate the port names/connections.  Black solid lines are the signal chain.}
 \label{fig:hookup}
 \end{figure}
 
-\subsection{Physical Connections}
-Possibly not the most aptly named ``physical'' connections are other connections that we wish to track for various operational reasons (e.g. tracing a problem to a single PAM chassis etc).  Unlike for hookup connections, these are not defined anywhere and new ones may be included as needs arise.  They are shown in Fig. \ref{fig:hookup} in blue.
```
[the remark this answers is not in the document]
david-deboer: Correct - there really isn't one (ie I don't differentiate) anymore except through various system architecture definitions in sys_def.

hera_mc/cm_sysutils.py
```diff
@@ -424,3 +424,238 @@ def get_apriori_antenna_status_for_rtp(self, status, at_date='now'):
...
+    return info
+
+
+def _get_macip(info):
+    data = []
+    for this_note in info:
+        if this_note.startswith('MAC'):
+            data.append(this_note.split('-')[1].strip())
+        elif this_note.startswith('IP'):
+            data.append(this_note.split('-')[1].strip())
```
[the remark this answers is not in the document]
david-deboer: Well, for no good reason....  I've made it one clause.  I also pulled that out to a method for marginal reasons (this could get more sophisticated in what it searches, so seemed worthwhile to just pull out now).

hera_mc/cm_hookup.py
david-deboer: There are a couple of places I have this for the following reason:  the way this works now is via the get_active stuff, which disallows it from happening.  I could see someone (maybe) wanting to use this from a different call and I wanted to retain the check, but not put in the extra test infrastructure (a whole other module) to test.

hera_mc/cm_utils.py
david-deboer: Yes indeed.  I've renamed general_table_handler

hera_mc/cm_active.py
david-deboer: I'm not sure I'd go with "best", but maybe "reasonable" in this instance.  I originally passed a variable, but I actually thought that was worse in this instance.

hera_mc/cm_sysutils.py
david-deboer: I struggle with this - does the module "print" or return a print string -- the latter is preferable to feed text to e.g. Matt K.  Any advice on what to rename for the distinction?  Ideally _everything_ would do it consistently, but I don't want to go over the entire codebase.

bhazelton: how about "format_which_node" or "string_which_node" for things that return strings, reserving "print" for methods that actually print things.

hera_mc/cm_active.py
bhazelton: I guess I'm struggling to understand the purpose of this, maybe just add a detailed comment about what it achieves that you can't achieve otherwise.

hera_mc/cm_hookup.py
bhazelton: Ok. I can't tell you how many times I've discovered that my error handling code doesn't work as expected (either doesn't get invoked when I expected it to or doesn't handle the problem as expected) until I add tests that actually exercise it.

hera_mc/cm_sysutils.py
david-deboer: How about this belt and suspenders approach?
def print_which_node(ant_node):
    print(formatted__which_node__string(ant_node))


def formatted__which_node__string(ant_node):

hera_mc/cm_hookup.py
david-deboer: I don't disagree, and I've contorted to check most stuff.  The other option is to remove the exception and let it error.  Since it is operational, it seemed better to catch it.

hera_mc/cm_sysutils.py
bhazelton: it's a little silly but it is at least clear.

bhazelton:
looks like tests aren't passing now.

PR #481, opened 2020-04-02
Corrf logger

mkolopanis:
Adds a deamon service meant to listen to `mc-log-channel` pubsub channel and add all logs above the input threshold into `subsystem_error` table in M&C. 

onsite testing still required

### DIFF
--- a/scripts/mc_listen_to_corr_logger.py
+++ b/scripts/mc_listen_to_corr_logger.py
@@ -11,11 +11,16 @@ from __future__ import absolute_import, division, print_function
 import json
 import redis
 import socket
+import logging
 
 from astropy.time import Time
 
 from hera_mc import mc
 from hera_mc.correlator import DEFAULT_REDIS_ADDRESS
+
+
+allowed_levels = ["DEBUG", "INFO", "NOTIFY", "WARNING", "ERROR", "CRITICAL"]
+logging.addLevelName(logging.INFO + 1, "NOTIFY")
 
 
 parser = mc.get_mc_argument_parser()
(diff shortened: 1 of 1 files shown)


scripts/mc_listen_to_corr_logger.py
```diff
@@ -0,0 +1,115 @@
...
+args = parser.parse_args()
+db = mc.connect_to_mc_db(args)
+
+hostname = socket.gethostname()
+redis_pool = redis.ConnectionPool(host=args.redishost)
+
+if args.level not in allowed_levels:
+    print("Selected log level not allowed. Allowed levels are:", allowed_levels)
+    print("Defaulting to WARNING")
+    level = logging.WARNING
```
bhazelton, line 61: why are you choosing a different default here than the argument default?

mkolopanis: Honestly this was copy-pasta, and now that you mention it, makes little sense. I'll update to the argument default.

mkolopanis: actually this should be impossible to hit. I've update the parser to use the "choices" kwarg which restricts users from sending in a level that is not in "allowed_levels"

mkolopanis:
proof it reports it's status to M&C 
image

mkolopanis:
some entries in the database 
image

scripts/mc_listen_to_corr_logger.py
bhazelton: great.

PR #482, opened 2020-04-02
Mc monitor daemons

mkolopanis:
Moving daemon monitoring over from corr_cm.

This script is almost and exact duplicate of the daemon_monitor from corr_cm with small formatting PEP changes and a larger try/catch loop for this daemon erroring.


mkolopanis:
testing looks good so far. Disabled other daemon, see reporting still on grafana
image

PR #486, opened 2020-05-19
Add missing entries to the antenna_status table docstring

bhazelton:


bhazelton:
@david-deboer can you review this so we can get it in?

PR #487, opened 2020-05-29
Fix linking for new flake8 version

bhazelton:
The tests are failing on master because they're using a newer flake8 version that's catching new things. This fixes them.


bhazelton:
You can't in general, but apparently `sorted` can take a generator.

PR #494, opened 2020-07-28
Geoloc

david-deboer:
This adds X,Y,Z information to the geo_handling stuff (in addition to E/N and lng/lat/elevation).
It also allows for csv and txt file output types.

### DIFF
--- a/hera_mc/geo_handling.py
+++ b/hera_mc/geo_handling.py
@@ -177,16 +177,15 @@ class Handling:
         import os.path as op
         self.file_type = fname.split('.')[-1]
         write_header = False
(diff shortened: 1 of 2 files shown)


hera_mc/geo_handling.py
```diff
@@ -389,15 +438,21 @@ def get_ants_installed_since(self, query_date, station_types_to_check='all'):
         """
         station_types_to_check = self.parse_station_types_to_check(station_types_to_check)
         dt = query_date.gps
+        d2r = pi / 180.0
```
bhazelton, line 441: here too.

PR #500, opened 2020-08-19
Add rosetta

david-deboer:
This adds a new "rosetta" table that tracks hera part numbers to "logical" part numbers, e.g. SNPC000072 - heraNode0Snap0


hera_mc/cm_active.py
```diff
@@ -174,6 +176,49 @@ def load_info(self, at_date=None):
             self.info.setdefault(key, [])
             self.info[key].append(info)
 
+    def load_rosetta(self, at_date=None):
+        """
+        Retrieve the current 'rosetta' mappings.
+
+        Since the mappings are all "current" there is no date.  They also don't
```
bhazelton, line 183: This docstring needs to be fixed -- there is a date.

david-deboer: I also am adding an "add_part_rosetta" function for scripting

PR #502, opened 2020-08-21
change to update rosetta

david-deboer:
The earlier "add_rosetta" wasn't fully thought out with the late change to adding timed entries.  This redoes is a la apriori_antenna.  The change allows it to more easily be used in the cm_db_updates workflow.

### DIFF
--- a/hera_mc/cm_partconnect.py
+++ b/hera_mc/cm_partconnect.py
@@ -47,38 +47,60 @@ class PartRosetta(MCDeclarativeBase):
                 '{self.stop_gpstime}>'.format(self=self))
 
 
-def add_part_rosetta(session, hpn, syspn, start_date, stop_date=None):
+def update_part_rosetta(hpn, syspn, at_date, date2=None, session=None):
     """
-    Add part information into database.
+    Update rosetta relationship in part_rosetta table.
+
+    If hpn <-> syspn relationship is active, will close it at at_date.
+    If hpn <-> syspn relationship is NOT active, it will start it at at_date.
+    If date2 is None, at_date is used as either start or stop, as appropriate.
+    Note that hpn is cast to upper, but syspn is not modified.
 
     Parameters
     ----------
-    session : object
-        Database session to use.  If None, it will start a new session, then close.
     hpn : str
         HERA part number
     syspn : str
         System part number
-    start_date : any format that cm_utils.get_astropytime understands
-        Date to use for the start
-    stop_date : any format that cm_utils.get_astropytime understands
-        Date to use for the stop
+    at_date : any format that cm_utils.get_astropytime understands
+        Date to use for action: start for new is relationship not active, else stop.
+    date2 : any format that cm_utils.get_astropytime understands
+        If not None, use for stop
+    session : object
+        Database session to use.  If None, it will start a new session, then close.
     """
+    hpn = hpn.upper()
+    at_date = int(cm_utils.get_astropytime(at_date).gps)
+    if date2 is not None:
+        date2 = int(cm_utils.get_astropytime(date2).gps)
+
     close_session_when_done = False
     if session is None:  # pragma: no cover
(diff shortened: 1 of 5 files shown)


hera_mc/cm_partconnect.py
```diff
@@ -47,38 +47,60 @@ def __repr__(self):
...
+            old_rose = trial
+    if ctr > 1:
+        raise ValueError("Multiple rosetta relationship active for {} - {}".format(hpn, syspn))
+    if old_rose is None:
+        new_rose = PartRosetta()
+        new_rose.hpn = hpn
+        new_rose.syspn = syspn
+        new_rose.start_gpstime = at_date
+        new_rose.stop_gpstime = date2
+        session.add(new_rose)
```
bhazelton, line 99: This seems potentially dangerous to me, in that it seems that if I ask for a pairing that doesn't exist it makes a new one without checking to see if there is an existing one for that hpn already. So it can generate multiple relationships between an hpn and different syspns, for example. Its that desired?

david-deboer: I had left open the possibility of more than one syspn active, but in retrospect that is probably not desirable.  One may have more than one syspn per hpn though.  So, I removed the filter on hpn and added a warning if trying to add a stop date where one exists.  Also updated checking for active relation.

bhazelton: What's the use case for more that one syspn per hpn? I'm thinking about this through the filter of SNAP hostnames. I never want more than one SNAP hostname per physical SNAP I don't think.

david-deboer: Currently, many snaps have multiple hostnames.  This is mostly historical I think, but given this fact I kept open the possibility.

bhazelton: ah, ok.

bhazelton:
@david-deboer looks like this needs to be rebased. Also, tests aren't passing.

hera_mc/tests/test_parts.py
```diff
@@ -103,21 +103,45 @@ def test_rosetta(mcsession, capsys):
...
+    assert int(active.rosetta['SNPC000709'].stop_gpstime) == 1280278818
+    # Add a test part to fail on update part
+    rose = cm_partconnect.PartRosetta()
+    rose.hpn = 'SNPC000701'
+    rose.syspn = 'heraNode0Snap701'
+    rose.start_gpstime = Time('2019-07-15 01:00:00', scale='utc').gps
+    mcsession.add(rose)
+    mcsession.commit()
+    pytest.raises(ValueError, cm_partconnect.update_part_rosetta, 'SNPC000701',
+                  'heraNode0Snap701', stop_at, None, mcsession)
```
bhazelton, line 126: You should check that it raises the expected error message like this:
```
    with pytest.raises(ValueError, match="Your expected error message here"):
          cm_partconnect.update_part_rosetta('SNPC000701', 'heraNode0Snap701', stop_at, None, mcsession)
```

hera_mc/tests/test_parts.py
```diff
@@ -103,21 +103,45 @@ def test_rosetta(mcsession, capsys):
...
+    rose2 = cm_partconnect.PartRosetta()
+    rose2.hpn = 'SNPC000712'
+    rose2.syspn = 'heraNode712Snap712'
+    rose2.start_gpstime = Time('2019-07-01 01:00:00', scale='utc').gps
+    rose2.stop_gpstime = Time('2020-08-01 01:00:00', scale='utc').gps
+    mcsession.add(rose2)
+    mcsession.commit()
+    with pytest.warns(UserWarning):
+        cm_partconnect.update_part_rosetta('SNPC000712', 'heraNode712Snap712',
+                                           '2020/01/02', session=mcsession)
```
bhazelton, line 144: And here you should check that the warning is what you expect using the `match` keyword like I did above.

PR #506, opened 2020-08-27
change to use setuptools_scm for versioning

bhazelton:
This follows the approach we use in RASG repos.


bhazelton:
@mkolopanis can you take a look at this?

setup.py
```diff
@@ -31,13 +25,13 @@
...
     'author_email': "hera-sw@lists.berkeley.edu",
-    'version': version.version,
+    "use_scm_version": {"local_scheme": branch_scheme},
     'packages': ['hera_mc', 'hera_mc.tests'],
     'scripts': glob.glob('scripts/*'),
     'include_package_data': True,
     'install_requires': ["numpy", "astropy", "sqlalchemy", "psycopg2",
                          "alembic", "python-dateutil", "tabulate", "pandas",
-                         "psutil", "cartopy", "pyyaml", "redis"],
+                         "psutil", "cartopy", "pyyaml", "redis", "setuptools_scm"],
```
mkolopanis, line 34: Do we actually need `setuptools_scm` as an install dependency? I know it is a build dependency and we have that declared in the `pyproject.toml`, but I wonder if it is necessary here or not.

mkolopanis (approved):
This looks good. Clean cut just like the RASG repos. I have one very small comment that doesn't really matter so I'll just go ahead and approve it.

setup.py
bhazelton: I think we need it as an install dependency because it's imported in the `__init__`.py.

mkolopanis: fair enough! :D let's get this done.

PR #507, opened 2020-09-08
fix docs

bhazelton:
The documentation on the antenna_status table was quite out of date (and also had some just plain mistakes). The white rabbit status table and the hera autos table were completely missing. This PR fixes those tables and also more consistently marks non-nullable columns throughout the docs. I also added a pointer to our documentation in the readme.


bhazelton:
Once the tex looks good I'll update the pdf in this PR as well.

david-deboer (approved):
Looks good - I didn't verify every not-null etc...

PR #509, opened 2020-09-18
Slim install

david-deboer:
This distills the packages needed for install to various options for slimmer installations.


bhazelton:
@mkolopanis can you review this? I wrote as much or more of it than Dave did, so outside eyes would be good.

hera_mc/cm_dossier.py
```diff
@@ -187,6 +186,7 @@ def table_row(self, columns, ports):
         list
             A row or rows for the tabulate display.
         """
+        from itertools import zip_longest
```
mkolopanis, line 189: I'm a little confused here. Isn't iterools in the standard libraray? I suppose it doesn't matter too much, but just feels weird to move the import.

mkolopanis (commented):
I think this is all pretty good. I have one little comment about a standard lib package.

hera_mc/cm_dossier.py
bhazelton: I agree, I think it should be moved back @david-deboer

mkolopanis (approved):
Looks good thanks for the adjustment.

PR #510, opened 2020-09-19
Weather dog

david-deboer:
This adds a watch dog to monitor node temperatures


dannyjacobs (approved):
A good start.  Don't forget to add this to the list of running cron jobs.  And note there where to add email addresses.

PR #511, opened 2020-09-21
Mctime

david-deboer:
This changes the `mc-gps.py` script to `mc_clock.py`, which is easier and better...  It now accepts civil, gps, and Julian inputs and outputs all variants back.


mkolopanis (commented):
This seems simple enough. But I'm not really a big fan of renaming/removing the old script without letting users know where to go to get the new one. It's a bit of a pain, but we might need to support both for a little bit. maybe @bhazelton has an opinion on the matter too?

david-deboer:
I can add it back in, however previous installers will have it in their local script bin (which is, I know, a sketchy way to support...)

scripts/mc_gps.py
```diff
@@ -4,7 +4,7 @@
 # Licensed under the 2-clause BSD license.
 
 """
-Prints Time info for given date/time
+Prints Time info for given date/time.  This is being deprecated for mc_clock.py
```
mkolopanis, line 7: While this information is useful for developers, the script itself should probably issue a warning when someone tries to use it. Telling them that it is being deprecated, what to use in its stead, and when they can expect this script to disappear.

mkolopanis (commented):
Thanks for adding the script back, but can we please add a little more information for the user by adding a warning?

david-deboer:
I added a print statement and put mc_gps back into the repo.

mkolopanis:
I do not see any print statements in the old mc_gps.py script beyond the 6 that output the converted times.

david-deboer:
I must not have committed the print statement?  I know I had it in there at some point.

scripts/mc_gps.py
```diff
@@ -17,6 +17,9 @@
     cm_utils.add_date_time_args(parser)
     args = parser.parse_args()
 
+print("DEPRECATION WARNING")
+print("mc_gps.py is being deprecated in favor of mc_clock.py, which is more flexible.")
+print("mc_gps.py will be retained until January 2021")
```
mkolopanis, line 22: I would prefer if an actual DeprecationWarning was issued with this text.

mkolopanis (approved):
Thanks @david-deboer dealing with all the nit picking.

PR #513, opened 2020-09-22
added reference to get_notes in cm_hookup

### LINKED ISSUE #512: Add PartInfo reference to notes
Currently the "reference" field from PartInfo is not propagated to the `get_hookup` and `get_notes` functions. An optional argument to add this could be especially useful for the website.

david-deboer:
Adds the reference field to the get_notes method in get_hookup:  {'notes': <note>, 'ref': <reference>}

This will close #512

### DIFF
--- a/hera_mc/cm_hookup.py
+++ b/hera_mc/cm_hookup.py
@@ -310,7 +310,9 @@ class Hookup(object):
                 gps_times = sorted(list(hu_notes[hkey][ikey].keys()))
                 for gtime in gps_times:
                     atime = cm_utils.get_time_for_display(gtime)
-                    entry_info += "\t{} ({})  {}\n".format(ikey, atime, hu_notes[hkey][ikey][gtime])
+                    this_note = ("{} ({})".format(hu_notes[hkey][ikey][gtime]['note'],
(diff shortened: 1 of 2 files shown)


mkolopanis:
I think this current implementation will completely break interfaces. maybe it should be an option keyword that changes the entries to be this new dict?

david-deboer:
I agree - we don't know who else uses this.  I'll make that change.

mkolopanis (commented):
This should suffice I think. Though I am actually a little confused how the `False` statement is being tested. Are there additional tests for `get_notes` that just were not edited?

david-deboer:
There are other tests for it that didn't need updating.

mkolopanis:
can you also please add verbiage in your top level comment for the PR that indicates it will close #512

mkolopanis:
a little finicky, but you have to use specific phrases for github to catch on like `closes #512`. see the full list [here](https://docs.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword). I think if you just remove the word "issue" it should be good.

mkolopanis (approved):
Thanks @david-deboer

PR #515, opened 2020-09-24
Fix info bug

david-deboer:
Checks to make sure that a comment field isn't just blank space (as can errantly happen in the googlesheet)


bhazelton (approved):
Looks straightforward.

PR #516, opened 2020-09-30
this improves the node dossier function and fixes a hookup bug

david-deboer:
This adds useful node functionality to the dossier.py to leave off an argument to get all active nodes.  It also fixes a bug that cropped up in cm_hookup.  Typing `dossier.py node` gives a table of connections, MACs, IPs, and connected antennas etc.

### DIFF
--- a/hera_mc/cm_sysutils.py
+++ b/hera_mc/cm_sysutils.py
@@ -559,14 +559,16 @@ def formatted__which_node__string(ant_node):
     return print_str
 
 
-def node_info(node_num, session=None):
+def node_info(node_num='active', session=None):
     """
     Generate information per node.
 
     Parameters
     ----------
-    node_num : list of int or str (can be mixed)
-        Node numbers, as int or hera part number
+    node_num : list of int or str (can be mixed), or str
+        Node numbers, as int or hera part number.
+        If 'active', use list of active nodes.
+        if 'all', use list of all.
 
     Returns
     -------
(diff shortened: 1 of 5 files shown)


hera_mc/cm_sysutils.py
```diff
@@ -576,6 +578,10 @@ def node_info(node_num, session=None):
     hu = cm_hookup.Hookup(session)
     na_from_file = node_antennas('file', session=session)
     na_from_hookup = node_antennas(hu, session=session)
+    if node_num == 'active':
+        node_num = sorted(list(na_from_hookup))
+    elif node_num == 'all':
+        node_num = sorted(list(na_from_file))
```
bhazelton, line 584: Not a big deal, but I don't think you need the `list`, I think `sorted` will take care of that.

hera_mc/tests/test_sysutils.py
```diff
@@ -280,6 +286,8 @@ def test_sysdef(sys_handle, mcsession):
 
 
 def test_sysutil_node(capsys, mcsession):
+    xni = cm_sysutils.node_info(['N91'], mcsession)
+    assert not len(xni['N91']['ants-file'])
```
bhazelton, line 290: same complaint here

bhazelton (commented):
Looks reasonable, a couple minor quibbles.

bhazelton (approved):

PR #517, opened 2020-10-05
added using ints for dossier node

david-deboer:
This implements what I had intended to do earlier (and included in the help) to allow nodes to be specified by just integers rather than node hpn in the 'dossier.py' node script.

### DIFF
--- a/hera_mc/cm_utils.py
+++ b/hera_mc/cm_utils.py
@@ -172,16 +172,16 @@ def stringify(X):
     return str(X)
 
 
-def listify(X, None_as_list=False, prefix=None, padding=None):
+def listify(to_list, None_as_list=False, prefix=None, padding=None):
     """
     "Listify" the input, hopefully sensibly.
 
     Parameters
     ----------
-    X
+    to_list
         Thing to be listified.
     None_as_list : bool
-        If False return None, otherwise return [None]
+        If False return None if to_list is None, otherwise return [None]
     prefix : str or None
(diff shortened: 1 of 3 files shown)


hera_mc/cm_utils.py
```diff
@@ -194,13 +198,24 @@ def listify(X, None_as_list=False):
         else:
             return None
     if isinstance(X, str):
```
bhazelton, line 200: `X` is not a good variable name in python. It should be multiple characters and not be capitalized per PEP8.

hera_mc/tests/test_cm_utils.py
```diff
@@ -76,6 +76,12 @@ def test_listify(input, expected):
     assert x is None
     x = cm_utils.listify(None, None_as_list=True)
     assert x[0] is None
+    x = cm_utils.listify('0-4', prefix='Test')
+    assert x[1] == 'Test1'
+    x = cm_utils.listify('0,3,4', prefix='Test', padding=3)
+    assert x[2] == 'Test004'
+    x = cm_utils.listify('N0-A', prefix='Test', padding=3)
+    assert x[0] == 'TestN0-A'
```
bhazelton, line 84: You should test that you get all the expected values and no others.

bhazelton (commented):
Looks pretty good, just a couple comments.

bhazelton (approved):
Looks good, thanks!

PR #519, opened 2020-10-12
had to fix the rosetta script and removed force upper

david-deboer:
The script did not match the method.  I also stopped forcing the part number to upper.


bhazelton (approved):
Looks fine. FYI, if you remove a line it may cause the overall coverage percent to go down, but that doesn't block the PR as long as the diff coverage is not negative. But adding tests to cover more lines of code is always welcome.

PR #520, opened 2020-10-20
added operationalError check in valid_db method

david-deboer:
It seems that periodically hera_mc can't connect to the database
[
from nb:
OperationalError: (psycopg2.OperationalError) could not translate host name "qmaster" to address: No address associated with hostname

(Background on this error at: http://sqlalche.me/e/13/e3q8)
]

This tries to connect twice with 5 sec breaks in between.

### DIFF
--- a/hera_mc/db_check.py
+++ b/hera_mc/db_check.py
@@ -56,12 +56,22 @@ def is_valid_database(base, session):
     True if all declared models have corresponding tables and columns.
 
     """
+    from sqlalchemy.exc import OperationalError
     if base is None:
         from . import MCDeclarativeBase
         base = MCDeclarativeBase
 
     engine = session.get_bind()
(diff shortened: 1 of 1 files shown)


hera_mc/db_check.py
```diff
@@ -56,12 +56,22 @@ def is_valid_database(base, session):
     True if all declared models have corresponding tables and columns.
 
     """
+    from sqlalchemy.exc import OperationalError
```
bhazelton, line 59: Let's move this to the top of the file. The only reason to do imports in functions is if they require libraries that are not required by the package.

bhazelton (commented):
Seems reasonable. There are some other options detailed in the linked documentation but I think it's fine to try this first.

bhazelton (approved):

PR #522, opened 2020-10-29
split send-email

david-deboer:
This was going to be a bigger PR to add some other watch dogs that I think are somewhat deprecated in the new scheme.  However, I did want to keep the "mail" method broken out from the initial, so this is to do that.

### DIFF
--- a/hera_mc/watch_dog.py
+++ b/hera_mc/watch_dog.py
@@ -116,7 +116,7 @@ def hosts_ethers(path='/etc', To=None, testing=False):
     meth, neth = read_hosts_ethers_file('ethers', host=hostname, path=path, testing=testing)
     if 'Error' in meth.keys():
         return send_message(meth['Subject'], meth['Message'], to_addr=To, skip_send=testing)
-    print("Now need to do the actually checking.")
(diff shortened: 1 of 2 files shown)


hera_mc/watch_dog.py
```diff
@@ -6,6 +6,40 @@
 """System watch-dogs."""
 
 
+def send_message(subject, msg, to_addr=None, from_addr='hera@lists.berkeley.edu', skip_send=False):
```
dannyjacobs, line 9: send_email might be more accurate. Is it ever not an email? Like a SMS or slack?

dannyjacobs (commented):
Making things difficult for myself.

hera_mc/watch_dog.py
david-deboer: I honestly have no preference, but you are right that that method is for email, so I've renamed.

david-deboer:
The actual purpose of this PR has been superseded, but I want to retain breaking out the send-email functionality.

bhazelton (approved):
Looks fine to me, but I am not familiar with the smtp stuff.

PR #530, opened 2021-02-18
fixed changed arg in wd script

david-deboer:
I had changed a keyword in the watch-dog script somewhere along the line which only reared its head when installed on site...


mkolopanis (approved):
This looks simple enough. Looks like we never piped through the keyword change from another PR? Makes me wonder if we aren't giving scripts a test run on PRs when we are running the on-site testing, that would probably be a good idea. @bhazelton do we need anything more on this? I'll give a provisional approval either way.

PR #518, opened 2020-10-07
Corr config

david-deboer:
This puts the info in the correlator configurations into a set of M&C tables. The configurations were already being tracked via hashes linked to files in the librarian, but this gets the info out into tables so it can be queried directly.

Still to do:
- ~update the documentation to include these new tables.~

PR #531, opened 2021-02-19
Fix bugs in code and tests that were causing code and tests to fail onsite

bhazelton:
These are fixes following the merge of PR #518 which led to both code errors and testing errors onsite. 

Tests pass with these fixes.

### DIFF
--- a/hera_mc/mc_session.py
+++ b/hera_mc/mc_session.py
@@ -1832,6 +1832,9 @@ class MCSession(Session):
 
         if dryrun:
             return command_list
+        else:
+            self.commit()
+            return
 
     def get_node_power_command(self, most_recent=None, starttime=None,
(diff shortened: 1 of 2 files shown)


hera_mc/mc_session.py
```diff
@@ -2967,6 +2970,9 @@ def add_correlator_config_from_corrcm(self, config_state_dict=None,
 
                     # add it to the config file table
                     self.add(config_file_obj)
+                    # commit to ensure that the hash is in the config file table before
+                    # anything that uses that hash as a Foreign key is added.
+                    self.commit()
```
bhazelton, line 2975: This is the most critical fix and the source of the code (not just testing) errors onsite.

plaplant (approved):
Looks good, thanks for the quick fix @bhazelton!

PR #533, opened 2021-02-19
fix accidental nodeControl reference change

bhazelton:
This fixes a bug that was caused by copy/paste errors when changing the imports in `mc_session.py` in PR #518. This part of the code is control code, so it is not covered by tests either on CI or onsite so was only caught from cronjob errors.


david-deboer:
Thanks!

david-deboer (approved):

PR #537, opened 2021-03-01
Get region

david-deboer:
This adds code to return the "region" of an antenna or antenna list.  The regions are [N, E, W] for the core antennas and [A, B] for the outriggers (A is the inner ring, B the outer).  Query inputs can be 123, '123', 'HH123', 'A123', etc. If given a list it returns a list, else it returns a string.  If given a part name ('HA325') for an outrigger antenna it will check the prefix and issue a warning if not in agreement.

### DIFF
--- a/hera_mc/geo_sysdef.py
+++ b/hera_mc/geo_sysdef.py
@@ -57,7 +57,9 @@ def _get_region(this_ant):
         Antenna number for which to check.
     """
     prefix = ''
-    if not isinstance(this_ant, int):
+    try:
+        this_ant = int(this_ant)
+    except ValueError:
         peeled = cm_utils.peel_key(this_ant, 'NPR')
         this_ant = peeled[0]
         prefix = peeled[1].upper()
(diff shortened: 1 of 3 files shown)


hera_mc/geo_sysdef.py
```diff
@@ -43,6 +44,63 @@
...
+        A ring (inner)
+        B ring (outer)
+
+    Parameter
+    ---------
+    ants : str, int or list
+            Antenna number or HERA part number of antenna.
+    """
+    return_as = 'list'
+    if not isinstance(ants, list):
```
jsdillon, line 95: You might want to also allow the input to be tuples or numpy arrays.

hera_mc/tests/test_geo_location.py
```diff
@@ -39,6 +39,15 @@ def test_geo_sysdef():
     assert int(a['HH123']['E']) == 540945
 
 
+def test_geo_regions():
```
jsdillon, line 42: It'd be good to see more tests. I'd try having lists with multiple entries. I'd try all the different types you mentioned in the PR, e.g. `123, '123', 'HH123', 'A123'`

jsdillon (changes requested):
Just a few suggestions but basically fine.

hera_mc/geo_sysdef.py
david-deboer: I made it more flexible to allow other list-y options, using TypeError vs ValueError.  Did similar in the _get_region module.

hera_mc/tests/test_geo_location.py
david-deboer: Added

jsdillon (approved):

PR #538, opened 2021-03-04
Add table to track observations run through RTP

### LINKED ISSUE #527: make a table to track RTP starting
*obsid
*submitted_time (GPS second integer)
jd (integer)
obs_tag ("engineering"/"science")
submitted (boolean)

We'll write a daemon that grabs all non-submitted jobs, breaks into groups of integer JDs and launches RTP on each group. RTP needs to update submitted time or make a new row if it already exists (for a re-run).

plaplant:
This PR adds a new table (`RTPLaunchRecord`) which keeps track of observations that have been run through RTP. So far the PR defines the new table, and methods to add records and update them. ~Still needed are scripts to do this from the command line, as well as a related cron job to automatically launch jobs for recent observations.~ Scripts for adding/updating entries and launching RTP have been added in a834d77 and 0a57a8e.

There are also some incidental changes around [deprecated numpy datatypes](https://numpy.org/doc/stable/release/1.20.0-notes.html#using-the-aliases-of-builtin-types-like-np-int-is-deprecated), which seemed like such a small change that I lumped them into this PR.

Closes #527.


hera_mc/mc_session.py
```diff
@@ -1410,6 +1412,82 @@ def get_rtp_task_resource_record(self, most_recent=None, starttime=None,
...
+        result = result[0]
+        # get the current number of rtp_attempts and increment
+        rtp_attempts = result.rtp_attempts
+        rtp_attempts += 1
+        # update result
+        result.rtp_attempts = rtp_attempts
+        result.submitted_time = submitted_time
+        self.commit()
+        return
+
```
bhazelton, line 1490: We probably want a `get_rtp_launch_record` method as well.

plaplant: Thanks, added a few different ways to search in 3850513 and 65c6983.

scripts/mc_launch_rtp.py
```diff
@@ -0,0 +1,149 @@
...
+    ),
+)
+
+args = ap.parse_args()
+db = mc.connect_to_mc_db(args)
+
+if args.jd is None:
+    # launch a separate job for each un-launched day
+    jd_list = []
+    query = mc.query(RTPLaunchRecord).filter(RTPLaunchRecord.rtp_attempts == 0)
```
bhazelton, line 76: do you want to use your getting methods here instead?

scripts/mc_launch_rtp.py
```diff
@@ -0,0 +1,149 @@
...
+    results = query.all()
+    for result in results:
+        if result.jd not in jd_list:
+            jd_list.append(result.jd)
+else:
+    jd_list = [args.jd]
+
+for jd in jd_list:
+    filelist = []
+    query = mc.query(RTPLaunchRecord).filter(RTPLaunchRecord.jd == jd)
```
bhazelton, line 86: same question here

scripts/mc_rtp_launch_record.py
```diff
@@ -0,0 +1,52 @@
...
+    int_jd = int(np.floor(starttime.jd))
+    obsid = int(np.floor(starttime.gps))
+    obs_tag = uv.extra_keywords["tag"]
+
+    with db.sessionmaker() as session:
+        obs = session.get_obs(obsid)
+        if len(obs) == 0:
+            print(f"observation {obsid} not in M&C, skipping")
+            continue
+        query = mc.query(RTPLaunchRecord).filter(RTPLaunchRecord.obsid == obsid)
```
bhazelton, line 42: and here.

scripts/mc_rtp_launch_record.py
```diff
@@ -0,0 +1,52 @@
...
+import numpy as np
+from astropy.time import Time
+
+from pyuvdata import UVData
+
+from hera_mc import mc
+from hera_mc.rtp import RTPLaunchRecord
+
+ap = mc.get_mc_argument_parser()
+ap.description = """Add or update the RTP Launch Record for an observation."""
```
bhazelton, line 15: When will this script be used? I'm not sure I understand the use case.

bhazelton:
We also need to update the M&C definition tex doc to include this new table.

dannyjacobs:
@plaplant We're reviewing this in the commissioning meeting to try and get it online soon as we can. Some questions that came up:
 1. How do new observations get added to this table?  Does the correlator do it at the same time it adds entries to the main "obsid" table?
2. How do we make sure that we don't launch RTP on nights that haven't finished being recorded yet?

scripts/mc_launch_rtp.py
plaplant: Sure, I can do that. I had written this script before I wrote the `get_*` methods, so I'll use them here instead.

plaplant:
@dannyjacobs to answer your questions:
1. The idea is that the `mc_rtp_launch_record.py` script will be called by the correlator after the `mc_add_observation.py` script (which adds a given observation to M&C).
2. Right now my plan is "run the RTP launch script well after data taking is finished". It's not a perfect system, but I'm not sure there's a better solution. (The previous redis-based keys method was brittle in cases where the correlator didn't shut down gracefully. Perhaps we could come up with a better solution after the current observing run.)

I'll address the other comments on this PR later today.

bhazelton:
@plaplant Are we going to have a daemon on the commissioning repo that calls the launch script once per day at the right time? or is it going to be a cronjob?

scripts/mc_launch_rtp.py
plaplant: This was done in ee9a244.

scripts/mc_rtp_launch_record.py
plaplant: done in ee9a244

scripts/mc_launch_rtp.py
plaplant: done in ee9a244

scripts/mc_rtp_launch_record.py
plaplant: I added some documentation in d2cf9c3. Let me know if it can be improved.

dannyjacobs:
I've added a rough description of this change to http://hera.pbworks.com/w/page/118389669/Real-Time%20Processor. The script and table names are not correct, but its a start.

bhazelton (approved):
This looks good to me.

bhazelton:
I'm merging this now, but not installing on site yet. @plaplant will do that when he has time (remember to update/restart all the things listed in the readme!)

PR #545, opened 2021-04-01
Moves conversion to sqlite to this repo.

david-deboer:
The sqlite stuff was in the cm updates repo, which is not where it should be.  This moves it to hera_mc and adds testing.


bhazelton (approved):
Looks reasonable to me. Good to get this all under testing!

PR #547, opened 2021-04-02
Future proof the db_check code by ensuring that it uses only the sqlalchemy public API.

### LINKED ISSUE #546: Make database validity checking more robust
Right now, the following function that checks whether the database matches the expected schema is relatively brittle.

https://github.com/HERA-Team/hera_mc/blob/fc114c065a9a9adab4eb0fd900d778ba53a3a4fe/hera_mc/db_check.py#L36

It relies on hidden attributes of the `declarative_base` API, which have moved between v1.3.x and v1.4.x of SQLAlchemy. Unfortunately, this means that we're stuck on an old version until this is resolved. Ideally our solution should use public API calls, so that we're slightly more insulated against similar changes in the future.

bhazelton:
Fix errors seen with new release of sqlalchemy (v1.4).

Closes #546

### DIFF
--- a/hera_mc/db_check.py
+++ b/hera_mc/db_check.py
@@ -4,8 +4,6 @@
 """Database consistency checking functions."""
 
 from sqlalchemy import inspect
-from sqlalchemy.ext.declarative.clsregistry import _ModuleMarker
-from sqlalchemy.orm import RelationshipProperty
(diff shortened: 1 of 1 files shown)


plaplant (approved):
Looks great, thanks @bhazelton!

PR #549, opened 2021-04-06
Fix RTP launch scripts to work in practice

plaplant:
After deploying the RTP changes on site, some small fixes were needed to make scripts work as intended. The current version of this branch is working in production, so we should probably merge this in.


mkolopanis:
good ol' test in production

scripts/mc_launch_rtp.py
```diff
@@ -180,29 +187,49 @@ def _get_obsids(filelist):
...
+    t0 = Time.now()
+    with db.sessionmaker() as session:
+        for filename, obsid in zip(filelist, obsids):
+            try:
+                session.update_rtp_launch_record(obsid, t0)
+            except RuntimeError:
+                sys.exit(
+                    f"could not update RTP Launch Record for file {filename}; aborting"
+                )
+    session.commit()
```
mkolopanis, line 235: Two things concerning this commit:
- It is outside the session context
- I'm not sure it is necessary either.
`update_rtp_launch_record` has a `self.commit()` before returning and if this is only used to make sure the launch records are committed then I think it is redundant. Worth checking out.

Also the `__exit__` method calls `commit` by default as long as there is not an error.

mkolopanis (commented):
I noticed a small thing concerning the calls to `session.commit`. I'm actually interested in how it did not complain before because it looks like the calls  to `session.commit` are outside the scope of the context manager.

scripts/mc_launch_rtp.py
mkolopanis: Oh there is another `session.commit()` call in the other file that was edited too. Same comment there. I must not have saved my comment before posting.

plaplant:
@mkolopanis if it's not a big deal, I'd like to merge this in when we can rather than waiting to test tmux on site. I'm about to make another PR for the RTP/slurm scripts, and I'd like to have this merged in already.

scripts/mc_launch_rtp.py
plaplant: Thanks for pointing this out. This is fixed in 5d45070.

mkolopanis (approved):
thanks @plaplant for putting this together. I'm not excited about a transient CI error but uhm what can you do sometimes.

PR #550, opened 2021-04-12
Flake8 changes, pre-commit, and CI update

mkolopanis:
Adds pre-commit yaml to the repo, makes lots of style changes to conform better to flake8 standards.
Updates the CI to use the setup-miniconda action, it's easier than the `install_conda.sh` script.

Lots of things here got touched by pre-commits built in hooks. @bhazelton did a lot of by hand changes to variable names and such.

The hooks are set up for black in pre-commit, we just need to enable it at a later date.


hera_mc/__init__.py
```diff
@@ -59,7 +59,7 @@ def isclose(self, other):
 
         self_columns = self.__table__.columns
         other_columns = other.__table__.columns
-        if set(c.name for c in self_columns) != set(c.name for c in other_columns):
+        if {c.name for c in self_columns} != {c.name for c in other_columns}:
```
plaplant, line 62: Hooray! I'm a huge fan of this notation, so I'm glad flake8 makes us do it 🎉

hera_mc/tests/test_db_check.py
```diff
@@ -24,7 +24,7 @@ class ValidTestModel(Base):
         __tablename__ = "validity_check_test"
 
         #: Running counter used in foreign key references
-        id = Column(Integer, primary_key=True)
+        id_ = Column(Integer, primary_key=True)
```
plaplant, line 27: It looks like in the main code, we add the `#noqa` option, rather than redefine `id`. Is that approach available there as well? No worries if not, I don't think the `#noqa` is actually harmful.

plaplant (commented):
Thanks @mkolopanis! Lots of great changes in here. I just have a single meaningful comment.

hera_mc/tests/test_db_check.py
mkolopanis: Hm interesting. I bet this got changed to not overload the build in `id` function. I am definitely in favor of that, but also understand where you're coming from. I would propose we use a better variable name in general :laughing:

mkolopanis: probably though this is because psql like the `id` variable name when python doesn't

plaplant (approved):

plaplant:
I'm going to manually merge this in manually, as I think the reduction in coverage is not real (the lines were not covered before this PR either).

PR #551, opened 2021-04-22
Cminfo improve

david-deboer:
This changes the way the snap cm info get passed to the correlator to only use the heraNodeNSnapM format (rather than the SNP serial number).

### DIFF
--- a/hera_mc/cm_redis_corr.py
+++ b/hera_mc/cm_redis_corr.py
@@ -14,34 +14,6 @@ REDIS_CMINFO_HASH = 'cminfo'
 REDIS_CORR_HASH = 'corr:map'
 
 
-def snap_part_to_host_input(part):
-    """
-    Parse a part string for the putative hostname and adc number.
-
-    If a redis session is supplied and key available it returns the hostname, otherwise
-    it returns the SNAP name.  If None, returns the SNAP name.  An example part is
-    'e2>SNPC000008'
-
-    Parameters
-    ----------
-    part : str
-        port>snap part string as returned by cminfo
-    redis_info : None or str
-        If str and the key is available it returns the hostname, otherwise SNAP name
-        If None, returns the SNAP name
-
-    Returns
-    -------
-    hostname : str
-        hostname as parsed from input 'part' and redis
-    adc_num : str
-        port ADC number, as parsed from input 'part'
-    """
-    adc, hostname = part.split('>')
-    adc_num = int(adc[1:]) // 2  # divide by 2 because ADC is in demux 2
-    return hostname, adc_num
-
-
 def cminfo_redis_snap(cminfo):
     """
     Build a dictionary of correlator mappings.
--- a/hera_mc/cm_redis_corr.py
+++ b/hera_mc/cm_redis_corr.py
@@ -56,8 +56,9 @@ def cminfo_redis_snap(cminfo):
             all_snap_inputs[snap_hostname].append(snap_input)
             try:  # if already present make sure it agrees
                 if snap_to_serial[snap_hostname] != snap_snr[_i]:
-                    msg = "{} inconsistent: {}  !=  {}".format(
-                          snap_hostname, snap_to_serial[snap_hostname], snap_snr[_i])
+                    msg = "{} inconsistent for antpol {}{}: {}  !=  {}".format(
+                          snap_hostname, ant, pol,
+                          snap_to_serial[snap_hostname], snap_snr[_i])
                     raise ValueError(msg)
             except KeyError:  # if not present, add it
                 snap_to_serial[snap_hostname] = snap_snr[_i]
--- a/hera_mc/cm_redis_corr.py
+++ b/hera_mc/cm_redis_corr.py
@@ -7,13 +7,14 @@
 import json
 import redis
 import time
+import warnings
 from . import cm_sysutils, correlator
 
 REDIS_CMINFO_HASH = 'cminfo'
 REDIS_CORR_HASH = 'corr:map'
 
 
-def snap_part_to_host_input(part, redis_info=None):
+def snap_part_to_host_input(part):
     """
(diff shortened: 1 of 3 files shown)


mkolopanis:
I think this overall is moving in the right direction, but I'm a little concerned about these sudden API changes. @plaplant do you have strong opinions about the topic? I have yet to take a detailed look here.

plaplant:
I agree with @mkolopanis that we don't want to make API changes without a deprecation period, unless we have good reason to and know that no other subsystems are using the old method.

@david-deboer can you go through the other repos that depend on hera_mc, and make sure that none of them were calling `snap_part_to_host_input` or `cminfo_redis_snap`? Calling either would break after making this change.

david-deboer:
These _shouldn't_ have been referenced (especially snap_part_to_host_input, which really was an internal kludge), but I did a thorough grep across all repos that may be installed with hera_mc and no appearance.

hera_mc/cm_redis_corr.py
```diff
@@ -72,25 +36,28 @@ def cminfo_redis_snap(cminfo, redis_info=None):
...
+            pol = psnap.lower()[0]
+            if pol not in ['e', 'n'] or '>' not in psnap:
+                warnings.warn(f"{psnap} is not an allowed correlator input")
+                continue
+            snap_input, snap_hostname = psnap.split('>')
+            channel = int(snap_input[1:]) // 2  # divide by 2 because ADC is in demux 2
+            ant_to_snap[ant][pol] = {'host': snap_hostname, 'channel': channel}
+            snap_to_ant.setdefault(snap_hostname, [None] * 6)
+            snap_to_ant[snap_hostname][channel] = name + pol.upper()
+            snap_to_serial[snap_hostname] = snap_snr[_i]  # This can/does overwrite per pol
```
plaplant, line 55: Rather than unconditionally overwriting this value for each polarization, it might be better to check that they're the same. Maybe something like this?
```python
try:
    existing_hostname = snap_to_serial[snap_hostname]
    if existing_hostname != snap_snr[_i]:
        raise ValueError(f"The snap_hostname {snap_hostname} is different between polarizations")
except KeyError:
    # key does not yet exist in dict; add it
    snap_to_serial[snap_hostname] = snap_snr[_i]
```

david-deboer: I added the suggested check.  However, I issue a warning not raise an error since I can imagine at some point we'd want to run some tests where they aren't the same snap and we want the system to still work in that case.  I kept the warning since we do want to be alerted if that does happen (purposely or inadvertently)

hera_mc/cm_redis_corr.py
```diff
@@ -72,25 +36,33 @@ def cminfo_redis_snap(cminfo, redis_info=None):
...
+            ant_to_snap[ant][pol] = {'host': snap_hostname, 'channel': channel}
+            snap_to_ant.setdefault(snap_hostname, [None] * 6)
+            snap_to_ant[snap_hostname][channel] = name + pol.upper()
+            all_snap_inputs.setdefault(snap_hostname, [])
+            all_snap_inputs[snap_hostname].append(snap_input)
+            try:  # if present, check if hostnames are the same between polarizations.
+                if snap_to_serial[snap_hostname] != snap_snr[_i]:
+                    warnings.warn(f"snap_hostname {snap_hostname} differs between inputs")
+            except KeyError:  # if not present, add it
+                snap_to_serial[snap_hostname] = snap_snr[_i]
```
mkolopanis, line 61: Looks like now this does not overwrite if they are different. Does this matter which one is tracked?

hera_mc/cm_redis_corr.py
david-deboer: I now correctly raise an error if the hostname to serial numbers are not consistent.

hera_mc/cm_redis_corr.py
david-deboer: I now correctly raise an error if the hostname to serial numbers are not consistent.

hera_mc/cm_redis_corr.py
```diff
@@ -72,25 +36,35 @@ def cminfo_redis_snap(cminfo, redis_info=None):
...
+            ant_to_snap[ant][pol] = {'host': snap_hostname, 'channel': channel}
+            snap_to_ant.setdefault(snap_hostname, [None] * 6)
+            snap_to_ant[snap_hostname][channel] = name + pol.upper()
+            all_snap_inputs.setdefault(snap_hostname, [])
+            all_snap_inputs[snap_hostname].append(snap_input)
+            try:  # if already present make sure it agrees
+                if snap_to_serial[snap_hostname] != snap_snr[_i]:
+                    msg = "{} inconsistent: {}  !=  {}".format(
+                          snap_hostname, snap_to_serial[snap_hostname], snap_snr[_i])
+                    raise ValueError(msg)
```
mkolopanis, line 61: I'm definitely more liking this error raising. Can we maybe tweak the message to read more like
> The snap hostname {snap_hostname} differs between {type of input device}. {parameret 1} has value {one value}, {parameter 2} has value {two value} 
is it possible to even construct those parameter 1 and 2 parts? maybe just at least the first sentence and then follow up with what the two values are like what you have now.

@plaplant any additional thoughts on this?

david-deboer: I've added the other piece of information we have at hand, which is the offending antpol.  Getting the other hostname would require another dictionary etc, which seems unnecessary since the fix is going into the database and fixing it and with the information provided in the error message you have enough information to find it.

mkolopanis: I think the point I was trying to make is the error message is very curt and more difficult to parse than the one you had before even.

mkolopanis (approved):
Thanks for the error message update. It at least seems easier to parse for me. We can always come back if there are other issues.

PR #554, opened 2021-04-28
Implement Black

mkolopanis:
Implements black styling.


INSTALL.md
```diff
@@ -58,7 +60,7 @@ management tools. The options that can be passed in this way are:
 [`sqlite`, `all`, `dev`]. The `all` option will install all optional
 dependencies, `dev` adds packages required for testing
 and also requires pre-commit to check for code style.
-Make sure to run `pre-commit install` to initilized the pre-commit hooks in the git repo.
+Make sure to run `pre-commit install` to initialized the pre-commit hooks in the git repo.
```
bhazelton, line 63: I think this should be "initialize" rather than "initialized".

bhazelton (approved):
Thanks @mkolopanis! I'm not sure why codecov is confused but I don't think this decreases coverage.

PR #556, opened 2021-06-18
pin codecov action version

mkolopanis:
pins codecov at v1.5.2 until codecov/codecov-action#372 is resolved


plaplant (approved):
Looks good to me. Thanks @mkolopanis!

PR #557, opened 2021-06-25
Observing around a certain RA

Zhilei:
This is my first pass of adding an argument `center_ra_deg` in the `correlator_control.py` script.
There are several numbers hard-coded in. Please feel free to let me know if you have any suggestions.


dannyjacobs (commented):
Hi Zhilei,
This looks like it would work. Note that it assumes all necessary systems will be on and running at the relevant time. This is not necessarily the case though. If the desired time is in the day time the nodes will be switched off.

  We're aiming to keep core operations software independent from research code as much as we can so we'll have to find another option for the LST2JD function. Pretty sure the Librarian does this when the correlator loads new files, we should find out what its using.
The correlator and M&C teams have continued to discuss the long-planned revamp for next season to hopeful resolve persistent instabilities and to enable baseline dependent averaging and fringe stopping. This will include changes to how the correlator starts up.  Startup will be handled by the correlator not M&C.  What will use this PR as a reference when formulating the next startup protocol.  

~D

Zhilei:
Hi Danny,
Thank you so much for the reply! I realized the ongoing correlator work which is related to this. I have called into one correlator telecon. I will keep an eye to the progress of the correlator LST locking work. 
Do you think we should merge this PR or just leave it in the branch right now? Thanks!

bhazelton:
I think we want to leave it on a branch, I don't think we want to merge it here but it will be a good reference for the updates that are coming. We definitely do not want to take a hera_cal dependency in this code though, so a different source (astropy or pyuvdata) for that functionality will be required.

bhazelton:
@dannyjacobs do we need this still or should I close it?

dannyjacobs:
Yeah, go ahead and close it.

PR #558, opened 2021-08-18
standardize the keywords in the `get_*_metric` methods

bhazelton:
The keywords in these methods behaved very differently than the other get methods in mc_session.py, this PR standardizes them.

### DIFF
--- a/hera_mc/mc_session.py
+++ b/hera_mc/mc_session.py
@@ -4795,8 +4795,8 @@ class MCSession(Session):
                 most_recent = True
             elif most_recent is False:
                 raise ValueError(
-                    'If most_recent is set to False, at least one of ant, pol, metric, '
-                    'obsid, or starttime must be specified.'
+                    'If most_recent is set to False, at least one of metric, obsid, or '
+                    'starttime must be specified.'
                 )
 
         if most_recent is True or starttime is not None:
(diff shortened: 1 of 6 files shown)


hera_mc/mc_session.py
```diff
@@ -4764,25 +4856,35 @@ def update_metric_desc(self, metric, desc):
...
         -------
         list of MetricList objects
 
         """
-        args = []
+        query = self.query(MetricList)
+
         if metric is not None:
-            args.append(MetricList.metric.in_(get_iterable(metric)))
+            query = query.filter(MetricList.metric == metric)
```
mkolopanis, line 4882: Any particular reason this is changed to only allow a single metric return? I feel like this multiple entry method _could_ be useful, though I bet it was hardly used.

mkolopanis (commented):
This looks pretty good, I do have one small quesiton.

hera_mc/mc_session.py
bhazelton: This was done in the name of unifying all the `get` methods on MCSession. We talked about it in a telecon and decided to make the metric get methods behave the same as other methods which do not accept lists only single filter values for each column.

mkolopanis: Ah fair enough, I must have missed that conversation.!

mkolopanis (approved):

PR #559, opened 2021-08-25
Make changes to RTP tables after testing

plaplant:
After testing on site, there are some changes that needed to be made to accommodate testing with RTP. Significantly, the `RTPProcessEvent` table requires the addition of a `task_name` column to capture which step in the pipeline was being recorded for a particular obsid. (In practice, multiple tasks for the same obsid were running at the same time, so the primary key constraint was being violated.) Additionally, some of the associated scripts required changes.

Because this PR includes a schema change, I would like to do some thorough testing before merging and deploying. In particular, this includes a changing of the primary key, which I am not sure I have specified optimally in alembic.


hera_mc/rtp.py
```diff
@@ -168,6 +170,115 @@ def create(cls, time, obsid, event):
...
+            Process event type.
+
+        Returns
+        -------
+        RTPTaskProcessEvent object
+
+        """
+        if not isinstance(time, Time):
+            raise ValueError('time must be an astropy Time object')
+        time = floor(time.gps)
```
bhazelton, line 223: should we enforce the enum here? Downside is that if something else comes in it will error and not get recorded. Maybe we don't want to throw errors...

scripts/add_rtp_task_jobid.py
```diff
@@ -14,59 +14,101 @@
...
 """
 
+import h5py
+import numpy as np
 from astropy.time import Time
 
 import hera_mc.mc as mc
 
 
+def get_obsid_from_file(filename):
```
bhazelton, line 24: This logic looks duplicated from `add_rtp_process_event`. Maybe it should go in the utilities?

bhazelton (commented):
This looks good, has it been tested thoroughly at site?

hera_mc/rtp.py
plaplant: I lean toward not enforcing it, because right now everything that uses it _should_ be resecting the enum already. My guess is right now we'd get an SQLAlchemy-related error right now, which is hopefully informative enough.

scripts/add_rtp_task_jobid.py
plaplant: Okay, I can break it out and put it into the utilities.

hera_mc/rtp.py
bhazelton: That's fine, just thought it was worth thinking about.

scripts/add_rtp_task_jobid.py
plaplant: Done in 1bb677e

plaplant:
@bhazelton I checked on-site and the `RTPTaskProcessEvent` and `RTPTaskJobID` tables seem to be getting populated now. I think this is ready for another look whenever you get a chance. Thanks!

bhazelton (approved):
Looks good, thanks!

PR #560, opened 2021-09-28
Update RTP launch script for H5C

jsdillon:


plaplant:
Thanks @jsdillon! I actually cherry-picked this commit and added it to the `update_jobid_script` branch, so it'll get merged in as a part of #559. I'm going to close this here, so if there are other changes, it's best to add them to that other branch.

PR #562, opened 2021-10-20
Update rtp launch

mkolopanis:
Updates the RTP launch script to skip to the next JD waiting if there is an error. Also replaces screen with tmux because paul and I like it better than screen.

TODO:
- [ ] onsite testing


plaplant:
Merging this in. If there are follow-up issues, we can address them in future PRs.

plaplant (approved):

PR #567, opened 2021-11-19
Add a CI job to ensure tests pass with only the required dependencies installed

bhazelton:
Also added testing against 3.10, drop testing agains python 3.6.

### DIFF
--- a/ci/min_deps.yaml
+++ b/ci/min_deps.yaml
@@ -1,0 +1,19 @@
+name: min_deps
+channels:
+  - conda-forge
+  - defaults
+dependencies:
+  - alembic
+  - astropy
+  - cartopy
+  - coverage
+  - numpy
+  - psycopg2
+  - pytest
+  - pyuvdata
+  - pyyaml
+  - pytest
+  - pytest-cov
+  - redis-py
+  - setuptools_scm
+  - sqlalchemy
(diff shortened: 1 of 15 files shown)


ci/min_deps.yaml
```diff
@@ -0,0 +1,19 @@
...
+  - cartopy
+  - coverage
+  - numpy
+  - psycopg2
+  - pytest
+  - pyuvdata
+  - pyyaml
+  - pytest
+  - pytest-cov
+  - redis-py
```
mkolopanis, line 17: I'm a little confused as to why the minimum dependencies requires redis. You're not setting up a redis database in the test suite.

mkolopanis (commented):
I have one question but otherwise looks pretty good.

ci/min_deps.yaml
bhazelton: Yeah, it's a little weird and I can change that if we want to. It's imported in several files and it was already listed as required in the `install_requires` section of `setup.py`, so I left it alone. But I did want to test that you can run without a redis server running.

mkolopanis: I see now. I've looked a little more thoroughly through the rest of the code base. I would be a lot more work adding in these checks for redis imports and erroring when calling redis required functions. Like you said we already have it set as a required dependency. I'm fine with booting this to an issue at least for now.

mkolopanis (approved):

PR #568, opened 2021-11-19
change watch dog to use EmailMessage and update tests

david-deboer:
For another reason I updated the smtplib mailer to use EmailMessage instead of strings, which is finicky.

### DIFF
--- a/hera_mc/watch_dog.py
+++ b/hera_mc/watch_dog.py
@@ -25,19 +25,23 @@ def send_email(subject, msg, to_addr=None, from_addr='hera@lists.berkeley.edu', 
 
     Returns
     -------
-    str or None
-        If skip_send, will return the composed message.
+    EmailMessage
+
     """
     import smtplib
+    from email.message import EmailMessage
     if to_addr is None:  # pragma: no cover
         to_addr = read_forward_list()
-    msg_to_send = ("From: {}\nTo: {}\nSubject: {}\n{}"
-                   .format(from_addr, ', '.join(to_addr), subject, msg))
-    if skip_send:
-        return msg_to_send
-    else:  # pragma: no cover
+    email = EmailMessage()
+    email.set_content(msg)
+    email['From'] = from_addr
+    email['To'] = to_addr
+    email['Subject'] = subject
+    if not skip_send:  # pragma: no cover
         server = smtplib.SMTP('localhost')
-        server.sendmail(from_addr, to_addr, msg_to_send)
+        server.send_message(email)
+        server.quit()
+    return email
 
 
 def read_forward_list():  # pragma: no cover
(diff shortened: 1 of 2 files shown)


hera_mc/watch_dog.py
```diff
@@ -48,7 +52,8 @@ def read_forward_list():  # pragma: no cover
             fwd.append(line.strip())
 
 
-def node_verdict(age_out=3800, To=None, testing=False, redishost='redishost'):
+def node_verdict(age_out=3800, To=None, return_as='node_info',
```
bhazelton, line 55: This new keyword needs to be added to the docstring

hera_mc/watch_dog.py
```diff
@@ -86,25 +91,34 @@ def node_verdict(age_out=3800, To=None, testing=False, redishost='redishost'):
...
+                msg += s
+        if len(msg):
+            msg = subject + '\n\n' + msg
+            this_email = send_email(subject, msg, To, skip_send=testing)
+    if return_as == 'node_info':
+        return node_info
+    elif return_as == 'email':
+        return this_email
+    else:
+        raise ValueError("Must be returned as 'node_info' or 'email'")
```
bhazelton, line 121: I like that you're checking that this, but I'd do it earlier so it's a fast fail rather than only failing after it did all the work.

hera_mc/watch_dog.py
```diff
@@ -25,19 +25,23 @@ def send_email(subject, msg, to_addr=None, from_addr='hera@lists.berkeley.edu',
...
     -------
-    str or None
-        If skip_send, will return the composed message.
+    EmailMessage
+
     """
     import smtplib
+    from email.message import EmailMessage
     if to_addr is None:  # pragma: no cover
         to_addr = read_forward_list()
```
bhazelton, line 34: Where is this function defined?

bhazelton (commented):
looks pretty reasonable

hera_mc/watch_dog.py
david-deboer: Done

hera_mc/watch_dog.py
david-deboer: Done

hera_mc/watch_dog.py
david-deboer: read_forward_list is defined in watch_dog.py

bhazelton: oh, I see, it's defined below where it's used. It should be moved, it's best practice to define functions before they're used.

david-deboer: I've moved it up

bhazelton (approved):

PR #570, opened 2021-11-23
Line up process event api

mkolopanis:
The add process event calls in the script where using incorrect keywords when trying to call mc_session methods.


plaplant (approved):
Thanks @mkolopanis! Easy enough

PR #572, opened 2021-12-01
Update rtp launch command

mkolopanis:
Changes the tmux launching command to explicitly source obs's bashrc. 
this allows us to just use conda (de)-activate calls.

changed the default shell to /bin/bash in .tmux.conf file on site.


bhazelton (approved):
simple.

PR #573, opened 2021-12-07
Urgent: script to confirm that file has been uploaded to librarian and is safe to delete.

dannyjacobs:
adds a script which checks to see if files have been added to the librarian. This is urgently needed to keep the catcher disk from filling up.

Note: Assumes the M&C "lib_files" table is a faithful representation of what is in the hera librarian. This is dangerous as the primary source should be the Librarian itself. Would be better to check with it directly, but I didn't have time to reload the Librarian API into my brainstack


plaplant (approved):
Looks great, thanks @dannyjacobs!

scripts/mc_check_lib_file.py
```diff
@@ -0,0 +1,33 @@
...
+ap.add_argument("files", type=str, default=None, nargs="*", help="list of files")
+
+
+args = ap.parse_args()
+db = mc.connect_to_mc_db(args)
+
+found_files = []
+for pathname in args.files:
+    filename = os.path.basename(pathname)
+    with db.sessionmaker() as session:
```
bhazelton, line 30: Do we really want the sessionmaker to be inside the file loop? I'd lean towards making the session outside the loop.

david-deboer (approved):
Rebased.

PR #574, opened 2021-12-22
fix the fem lna power flag

david-deboer:
This makes the corresponding fix in hera_mc to accommodate the change in the fem power flag in hera_corr_f via hera_corr_cm.


bhazelton (approved):
Looks good.

PR #575, opened 2021-12-22
took out the fem_e_/fem_n_ thing to just be fem_

david-deboer:
I hadn't updated the docstring...


bhazelton (approved):
Thanks for catching that!

PR #576, opened 2021-12-23
Launch more RTP jobs

plaplant:
This is a quick PR to increase the number of jobs that RTP runs from the default (30) to 100.


bhazelton (approved):
Looks simple. I'm a little worried about how much complex logic is in this script and so has no tests. Long term we should pull most of this logic out to a utility function or MCSession method and make the script but thin.

PR #579, opened 2022-01-13
change get_astropytime bounds

david-deboer:
Make a trivial change to extend the time bounds of get_astropytime so that old queries don't issue warning


mkolopanis (approved):
more times.

PR #581, opened 2022-01-21
Get pam_id and fem_id into database

david-deboer:
This will hopefully make this more robust in terms of getting meaningful id numbers into the database.

### DIFF
--- a/hera_mc/correlator.py
+++ b/hera_mc/correlator.py
@@ -1435,14 +1435,20 @@ def _pam_fem_id_to_string(idno):
         decoded id number
 
     """
+    if isinstance(idno, list):
+        return ":".join([str(x) for x in idno])
     try:
         val = json.loads(idno)
         if isinstance(val, list):
             return ":".join([str(x) for x in val])
         else:
-            return str(idno)
-    except json.JSONDecodeError:
-        return str(idno)
+            retval = str(idno)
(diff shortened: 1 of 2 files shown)


bhazelton:
Can you explain a bit of what the fix is doing? I see that you're inserting colons between sections of a list, why is that?

david-deboer:
Previously, this assumed a very specific format of the id reported into redis in a specific way from hera_corr_f.  That underlying fundamental way has changed in hera_corr_f.  That value was converted to hex.  Now, it handles various options and instead of a hex number converts to a colon-delimited number set or other as appropriate.  This monitoring is actually currently mostly broken on the hera_corr_f side, so will further change so this new method should allow more flexibility.

bhazelton:
Ok, thanks. So should this PR wait on more changes in hera_corr_f?

david-deboer:
I think that this version will handle where we end up in hera_corr_f (ie I don't think that the new way there now needs to change, but overall that monitor on hera-snap-head is not working reliably).  So, I don't think it needs to wait.  Thanks!

bhazelton:
ok. I guess this change means that IDs for the same hardware will not match before and after.

bhazelton (approved):
I can't tell if this code is correct because the input comes from hera_corr_f. But if it's working on site I think it is ok.

PR #583, opened 2022-02-02
Modify installation instructions to match reality

plaplant:
On a recent site trip, pot1 was changed to use the netboot like most of the other pot machines. This means that it's no longer necessary to update the environment on a schema change (as it's updated with qmaster).


mkolopanis (approved):
Works for me!

PR #588, opened 2022-04-25
change dir name to docs and edit README for that

david-deboer:
For naming convenience, changed Docs sub-directory name to docs.


bhazelton:
It looks like this is somehow dropping coverage for line 1426 of `correlator.py`. I don't really understand why that would be. @david-deboer I think this is code you wrote, any insight?

david-deboer:
The tests must be using the production version of redis.  Recall we are running on a branch for hera_corr_cm with the new implementation, which must change the id format such that it doesn't produce the same list.

bhazelton (approved):
Thanks!!

PR #592, opened 2022-05-02
Fix rtp resource records

jburba:
## Description

Updated `add_rtp_task_resource_records.py` to account for RTP jobs submitted with multiple attempts.  Also fixed a bug in the SLURM query related to parsing a string with a fixed column width.

## Motivation and Context

Fixes broken `add_rtp_task_resource_records.py` script.

## Types of changes

- [ ] Version change
- [ ] Other

## Checklist:

Bug fix checklist:
- [ ] My fix includes a new test that breaks as a result of the bug (if possible).

### DIFF
--- a/scripts/add_rtp_task_resource_records.py
+++ b/scripts/add_rtp_task_resource_records.py
@@ -1,4 +1,4 @@
-#!/home/obs/anaconda/envs/hera_mc_rtp/bin/python3.1
+#! /usr/bin/env python
 # -*- mode: python; coding: utf-8 -*-
 # Copyright 2021 the HERA Collaboration
 # Licensed under the 2-clause BSD license.
--- a/scripts/add_rtp_task_resource_records.py
+++ b/scripts/add_rtp_task_resource_records.py
@@ -109,9 +109,14 @@ def query_slurm_db(jobid):
         stop_time = Time(output[3], format="isot", scale="utc")
         slurm_dict["stop_time"] = stop_time
 
-        max_memory = float(output[4][:-1])
-        # SLURM provides the MaxRSS in bytes, but we choose to record in megabytes
-        slurm_dict["max_memory"] = max_memory / 1e6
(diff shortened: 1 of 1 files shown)


scripts/add_rtp_task_resource_records.py
```diff
@@ -157,11 +240,83 @@ def query_slurm_db(jobid):
...
+                                [t[0] for t in mult_times.all()], dtype=int
+                            )
+                            closest_time_ind = np.argmin(
+                                np.abs(start_time_tpe - mult_times)
+                            )
+                            closest_time = int(mult_times[closest_time_ind])
+                            task = mult_attempts.filter(
+                                RTPTaskMultipleJobID.start_time == closest_time
+                            )
+                            task = task[0]
```
plaplant, line 317: It looks like this logic is very similar to above (lines 192-222). I wonder if it's possible to harmonize them into a single function, where the table type is passed in? It looks like the queries columns (`event`, `time`, and `start_time`) are common between the two tables.

plaplant (commented):
Thanks @jburba! I appreciate all the effort to get this working on site! I have just a couple of small comments.

More generally, I am wondering if we should put some of this logic into a module inside of the package, rather than have a lot of logic in a script. I think it's okay to keep it this way if we want, but then we should probably make an issue to move this to module-level code (so we can write tests for it).

scripts/add_rtp_task_resource_records.py
jburba: I think it might be possible to put this logic into a function.  I'll take a look at that over the next few days.

bhazelton: There's an example of a function that runs similar queries on different tables in `mc_session._time_filter` in case it's useful.

plaplant (approved):
Looks good, thanks @jburba!

PR #596, opened 2022-05-06
minor fixes and cleanup

### LINKED ISSUE #543: remove `rtp_status` table from SQL alchemy code
That table is no longer updated and is superseded by the `rtp_launch_record` table (created in #538).

We should remove it from the model but not drop the table.

### LINKED ISSUE #586: Add katportalclient as an optional dependency
We need to use katportalclient to get the weather on site. We should add it as an optional dependency.

bhazelton:
## Description

While working on adding support for handling Exception messages passed from hera_corr_cm, I found a few things that needed to be fixed/cleaned up. There has since been work to eliminate those from the output of hera_corr_cm, so this doesn't include that support.

I also added some other cleanup items from our issues:
- update the PR template to include stuff about modifications of redis structure
- add katportalclient as an optional dependency
- remove RTPStatus table related code (but do not update schema because we want to keep the table on site)

## Motivation and Context

fixes #586 
fixes #543 

## Types of changes

- [ ] Version change
- [x] Other

## Checklist:

Other:
changes when this is merged. **None required**


hera_mc/__init__.py
```diff
@@ -66,13 +66,19 @@ def isclose(self, other):
...
+            if not isinstance(other_c, type(self_c)):
+                print(
+                    f"column {c} has different types, left is {type(self_c)}, "
+                    f"right is {type(other_c)}."
+                )
+                return False
             if isinstance(self_c, int):
                 if self_c != other_c:
-                    print("column {col} is int, values are not equal".format(col=c))
+                    print(f"column {c} is int, values are not equal")
```
david-deboer, line 77: Should change the looping variable from c to, say, col.

plaplant (approved):
Looks great, thanks @bhazelton!

PR #597, opened 2022-05-17
Fix session

david-deboer:
WIP

Convert configuration management classes, methods and functions to handle sessions such that when complete they close.

## Description

Configuration management files did not explicitly close connections to the Postgres database upon completion, letting Python trash handling to clean up.  The time lag meant that many connections to the db could persist and eventually cause too many connections to be set up.  Once that exceeds a limit (currently 100, but the config file has been updated to 150 but not restarted) errors will ensue.

This entails interface changes to many classes and methods in all the changed files.  Classes like ActiveData and Hookup now require a session to be passed into them.

## Motivation and Context

As mentioned above, this explicitly closes connections to the db so that the max connections are less likely to be exceeded.

## Types of changes

- [x] New feature without schema change (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to change)
- [x] Version change
- [ ] Other

## Checklist:

Breaking change checklist:
- [x] I have updated the docstrings associated with my change using the [numpy docstring format](https://numpydoc.readthedocs.io/en/latest/format.html).
- [x] I have detailed any changes required to other HERA repos above and I will coordinate
implementing the changes across all the repos.
changes when this is merged.
- [x] I have added tests to cover my changes.

Version change checklist:
- [x] I have updated the [CHANGELOG](https://github.com/HERA-Team/hera_mc/blob/main/CHANGELOG.md) to put all the unreleased changes under the new version (leaving the unreleased section empty).


scripts/add_part_info.py
```diff
@@ -47,20 +47,20 @@ def query_args(args):
...
+        if args.verbose:
+            print("Adding info for part {}:{}".format(args.hpn, args.rev))
+        cm_partconnect.add_part_info(
+            session,
+            args.hpn,
+            args.rev,
+            args.comment,
+            at_date=at_date,
+            reference=args.reference,
+        )
```
bhazelton, line 66: May need to call `session.commit()` here.

hera_mc/cm_active.py
```diff
@@ -5,8 +5,44 @@
 
 """Methods to load all active data for a given date."""
 
-from . import cm_utils
+from . import mc, cm_utils
 from . import cm_partconnect as partconn
+from copy import copy
```
bhazelton, line 10: super minor, but this line should be at the top of the imports with an empty line before the imports from the hera_mc package.

bhazelton (commented):
I think this is looking pretty good. I like what you've done on the objects and in the scripts.
A couple of high-level comments:
- There are a fair number of functions (besides the new ones) that will make a session if you don't pass one in. I'd only do this in functions where it is very likely that users will want to call that function and nothing else that uses the DB. In general, I'd lean towards making most of them require a session to be passed in.
- There are a lot of new lines that still need test coverage, if you look at the "files changed" tab, Codecov has nicely labeled them for you.

scripts/add_part_info.py
bhazelton: never mind, the function it calls does the commit.

hera_mc/cm_hookup.py
```diff
@@ -14,6 +14,74 @@
...
+    exact_match=False,
+    hookup_type="parts_hera",
+    testing=False,
+):
+    """
+    Return a single hookup dossier.
+
+    This allows for a simple way to get a part hookup, however if more
+    transactions are needed, please use the structure to generate a session
+    and pass that session to the methods.
```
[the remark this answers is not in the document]
david-deboer: Updated and added -- structure meant session marker.

hera_mc/cm_sysutils.py
```diff
@@ -570,8 +564,14 @@ def node_antennas(source="file", session=None):
                     prefix = "HH"
                 ants_per_node[node_hpn].append("{}{}".format(prefix, ant))
     else:
+        if session is None:
```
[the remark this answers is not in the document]
david-deboer: Per your introductory remark, I'm going to take out None default on session and make it mandatory.  Also, there isn't a testing flag if I don't.  See below.

hera_mc/cm_hookup.py
```diff
@@ -14,6 +14,74 @@
...
+
+    Returns
+    -------
+    dict
+        Hookup dossier dictionary as defined in cm_dossier
+
+    """
+    if testing:
+        db = mc.connect_to_mc_testing_db()
+    else:  # pragma: no cover
```
bhazelton, line 70: we should have an onsite test that covers this line.

hera_mc/cm_redis_corr.py
```diff
@@ -96,8 +96,22 @@ def set_redis_cminfo(
     rsession = redis.Redis(connection_pool=redis_pool)
 
     # Write cminfo content into redis (cminfo)
+    if session is None:
+        from hera_mc import mc
+
+        if testing:
+            db = mc.connect_to_mc_testing_db()
+        else:  # pragma: no cover
```
bhazelton, line 104: we should have an onsite test to cover this too

hera_mc/cm_revisions.py
```diff
@@ -79,6 +87,13 @@ def get_last_revision(hpn, session=None):
         (hpn, rev, rev_query, started, ended, [hukey], [pkey])
 
     """
+    if session is None:
```
bhazelton, line 90: Does this function (and the other functions in this file) really need this? It doesn't seem like something a user is likely to run as a stand-alone db call to me, but I'm not a CM power user.

hera_mc/watch_dog.py
```diff
@@ -172,6 +172,9 @@ def node_temperature(
 
         db = mc.connect_to_mc_db(None)
         session = db.sessionmaker()
+        close_session_when_done = True
```
bhazelton, line 175: we want this to be tested on site as well.

bhazelton (commented):
good progress. I think we want to be sure that lines that we cannot cover with CI are covered on site.

hera_mc/cm_revisions.py
david-deboer: The revision stuff is my one big regret in the whole CM business and came about from some very early discussions.  These functions are used rarely to never (I think one or two scripts uses one or two).  I'd like to raise a technical debt issue to purge as much of the revision stuff as possible.  In the meantime, it might be best to leave the structure as it is to avoid the unintended.

hera_mc/watch_dog.py
david-deboer: I think I'll need some assistance in these site test comments.

scripts/add_qm_metrics.py
```diff
@@ -25,9 +25,7 @@
...
 db = mc.connect_to_mc_db(args)
-session = db.sessionmaker()
 
-for f in args.files:
-    session.ingest_metrics_file(f, args.type)
-
-session.commit()
+with db.sessionmaker() as session:
+    for f in args.files:
+        session.ingest_metrics_file(f, args.type)
```
bhazelton, line 31: I think you still need the commit, it doesn't look like that method has a commit at the end.

david-deboer: The context manager has a commit on __exit__

bhazelton (approved):
This looks pretty good, let's get it in. There are some followup improvements planned.

PR #603, opened 2022-06-02
Add snap mini corr

mkolopanis:
Adds a mini data catcher script to save snap spectra to a UVH5 file.
## Description

Reads the `status:snap:*` keys in redis and compiles them into a UVData object to save to disk.
## Motivation and Context

Danny asked for this.
## Types of changes

- [ ] Version change
- [x] Other

## Checklist:

Other:
changes when this is merged.

### DIFF
--- a/scripts/hera_mini_snap_data_catcher.py
+++ b/scripts/hera_mini_snap_data_catcher.py
@@ -4,6 +4,7 @@
 # Licensed under the 2-clause BSD license.
 """Accumulate SNAP spectra from redis and write to a UVH5 file."""
 import time
+import json
 import logging
 import argparse
 import traceback
@@ -268,6 +269,9 @@ while True:
                 uvd.phase_type = "drift"
 
                 uvd.reorder_blts("time", "baseline")
+                uvd.extra_keywords["snap_to_ant_mapping"] = json.dumps(
+                    snap_to_ant_mapping
+                )
                 fname = f"zen.{Time.now().jd:.6f}.snap_autos.uvh5"
                 logger.info(f"Writing output file {fname}.")
                 uvd.write_uvh5(fname)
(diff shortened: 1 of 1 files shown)


mkolopanis:
@bhazelton I think this is worth a look. I'm open to any ideas.

we still have a problem were sometimes while loop thinks it is ready to trigger but only after 200ms. I think this is due to how snap spectra only update on a ~10 min timescale and even then only roll out in stages.

scripts/hera_mini_snap_data_catcher.py
```diff
@@ -0,0 +1,299 @@
...
+                # Nblts
+                time_array = []
+                ant_array = []
+
+                downtime = 0
+                file_len = 0
+
+                while downtime < args.max_downtime and file_len < args.max_file_len:
+                    time.sleep(0.01)
+                    # do the thing
```
bhazelton, line 133: remove this line or make it more helpful

scripts/hera_mini_snap_data_catcher.py
```diff
@@ -0,0 +1,299 @@
...
+                                # Time has not progressed yet.
+                                # Just sit tight
+                                continue
+                            last_time_mapping[snap_hostname] = timestamp
+
+                        auto_corr = status["autocorrelation"]
+
+                        if data_array is None:
+                            data_array = np.zeros(
+                                (0, auto_corr.shape[0], 2), dtype=np.float32
```
bhazelton, line 174: Why is the first axis length 0?

bhazelton (commented):
Looking good!

scripts/hera_mini_snap_data_catcher.py
mkolopanis: The 0 in the shape acts as a placeholder axis so we can accumulate data into it until we reach either timeout condition or a hour of data. I didn't want to necessarily assume how long a data file would get so I let it grow.

bhazelton (approved):
I think this is ready. Eventually we might want to put more of the logic in functions that can be tested, but that's hard right now and we should just get this in.

PR #607, opened 2022-06-10
update rtp process record to be compatible with scm

mkolopanis:
@jsdillon noticed that RTP is failing now that we switched all the repos to `setuptools_scm`. I updated the `add_rtp_process_record` to treat everything like how pyuvdata was parsed before.

### DIFF
--- a/scripts/add_rtp_process_record.py
+++ b/scripts/add_rtp_process_record.py
@@ -95,7 +95,7 @@ if __name__ == "__main__":
             rtp_git_version=version_info["hera_opm"]["tag"],
             rtp_git_hash=version_info["hera_opm"]["hash"],
             hera_qm_git_version=version_info["hera_qm"]["tag"],
-            hera_qm_git_hash=version_info["herq_qm"]["hash"],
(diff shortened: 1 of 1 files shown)


scripts/add_rtp_process_record.py
```diff
@@ -6,15 +6,13 @@
...
 import warnings
+import importlib
 from pkg_resources import parse_version
 
 import numpy as np
 from astropy.time import Time
 
 import hera_mc.mc as mc
-import hera_qm
-import hera_opm
```
jsdillon, line 16: Unfortunately, hera_opm is not on setuptools-scm yet, so this will still break.

mkolopanis: we've merged a new hera_opm that uses setuptools scm

mkolopanis:
testing on 2459740. RTP is still failing but not because of the process record

jsdillon:
RTP failures are not M&C's problem, just something that happens when the array is switched to load/noise. I think this PR solves the issue from RTP's perspective

plaplant (approved):
Looks good, thanks @mkolopanis! Feel free to merge/install anytime.

PR #612, opened 2022-07-19
More snap status

### LINKED ISSUE #587: Grab rest of f_status keys from redis
The new hera_corr_cm plumbs through the rest of the snap status redis keys, which we'd like to capture.  These are:
- is_programmed
- adc_is_configured
- is_initialized
- dest_is_configured
- version
- sample_rate (maybe not include, since shouldn't changed)
- input - have to figure out, but the value is typically adc,adc,adc,adc,adc,adc (as opposed to digital noise?)

bhazelton:
## Description

Add more SNAP status information. Most of this is just new columns in the `snap_status` table, but the input information ("adc" or "noise-{d}" where {d} is the seed) is per SNAP channel, so that is broken out in a new table called `snap_input`. That table is keyed the same way as `snap_status` but I also added columns for the connected antpol (when available, derived from config management) so it can be lined up with the fem switch from the `antenna_status` table to determine what the signal source per antenna is.

## Motivation and Context

closes #587 

## Types of changes

- [ ] Version change
- [ ] Other

## Checklist:

Schema change:
- [x] I have tested looping upgrading and downgrading the alembic version and tests pass consistently.
changes when this is merged.
- [x] I have updated the schema documentation document with my changes (under docs/mc_definition.tex)


david-deboer (commented):
Looks pretty straight-forward.  I think that it would be helpful in the mc_def.tex doc and where the table gets defined to remind that snap_channel_number 0-5 corresponds to n0,e2,n4,e6,n8,e10 ports on the SNAP

david-deboer (approved):
Thanks

PR #618, opened 2022-08-08
Add a tag column to the hera_obs table

### LINKED ISSUE #605: Add observing tag to the hera_obs table
Can be filled by the catcher in future.

Column will be nullable because we don't have values for the past, but the function to fill it could require a value.

bhazelton:
## Description

Add a `tag` column to the `hera_obs` table.

Note: As currently written, the `create` classmethod on the `Observation` object and the `add_obs` method on `MCSession` _requires_ `tag` to be passed. So we will need to update RTP to pass it. The column is actually allowed to be null in the database, though, because all the old data do not have tags. I think long term we want RTP to set the `tag` field, but if it would be easier short term I can make that an optional parameter.

## Motivation and Context

closes #605 

## Types of changes

- [ ] Version change
- [ ] Other

## Checklist:

Schema change:
- [x] I have tested looping upgrading and downgrading the alembic version and tests pass consistently.
changes when this is merged.
- [x] I have updated the schema documentation document with my changes (under docs/mc_definition.tex)


hera_mc/observations.py
```diff
@@ -58,7 +66,7 @@ def length(self):
         return self.stoptime - self.starttime
 
     @classmethod
-    def create(cls, starttime, stoptime, obsid, hera_cofa):
+    def create(cls, starttime, stoptime, obsid, hera_cofa, tag):
```
plaplant, line 69: Do we want to explicitly check that `tag` is one of the suggested columns? I don't think we want this to be an actual `enum`, but it might be good to enforce it's in one of the delineated categories.

plaplant (commented):
Thanks @bhazelton! I think this looks great. I just have a question and found a few print statements that should be removed.

We should also change `mc_add_observation.py` to add the tag to the observation when it's first added to M&C. I can make that change as part of this PR.

hera_mc/tests/test_observations.py
```diff
@@ -60,34 +61,42 @@ def test_add_obs(mcsession):
 
     t3 = t1 + TimeDelta(10 * 60.0, format="sec")
     t4 = t2 + TimeDelta(10 * 60.0, format="sec")
-    test_session.add_obs(t3, t4, utils.calculate_obsid(t3))
+    print(utils.calculate_obsid(t3))
```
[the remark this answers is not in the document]
bhazelton: yep, thanks!

hera_mc/observations.py
bhazelton: I wasn't sure. I'm happy to add some checking on that. If it doesn't match, do we want to error or just warn?

plaplant (approved):
Looks great, thanks @bhazelton! Merging this in so we can do a rebase for the forthcoming PR.

PR #620, opened 2022-08-10
Mini snap make folders

mkolopanis:
Organizes the mini-snap catcher to write data to folders based on the JD.

Tried to sure up some logic to reduce the amount of logger warnings about data timeouts.

On site verified that data is written to the expected folders but haven't verified the logger is any less talkative yet (currently taking data at time of writing).

### DIFF
--- a/scripts/hera_mini_snap_data_catcher.py
+++ b/scripts/hera_mini_snap_data_catcher.py
@@ -299,7 +299,7 @@ while True:
 
                 # make the fold for the JD if it does not exist
                 folder = Path(args.outdir) / folder_name
-                folder.mkdir(exists_ok=True)
+                folder.mkdir(exist_ok=True)
 
                 fname = folder / f"zen.{date:.6f}.snap_autos.uvh5"
                 logger.info(f"Writing output file {fname}.")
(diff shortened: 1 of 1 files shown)


scripts/hera_mini_snap_data_catcher.py
```diff
@@ -278,9 +294,18 @@ def get_blt_index(antnum, time, ant_array, time_array):
...
                     snap_to_ant_mapping
                 )
-                fname = f"zen.{Time.now().jd:.6f}.snap_autos.uvh5"
+
+                date = Time.now().jd
+
+                folder_name = f"{int(np.floor(date)):d}"
+
+                # make the fold for the JD if it does not exist
+                folder = Path(args.outdir) / folder_name
```
bhazelton, line 303: I assume this works, I would normally use `os.path.join` myself.

bhazelton (commented):
Looks reasonable.

scripts/hera_mini_snap_data_catcher.py
mkolopanis: Yeah I like the object based interface for the pathlib.Path object. It does resolve to `os.path.join` eventually I bet.

bhazelton (approved):

PR #626, opened 2022-09-07
Fix how prefix directory is handled when adding files to RTP

plaplant:
## Description

When adding data files to the M&C tables for RTP, we need to know where to find these files on disk. In particular, there are in principle two different storage nodes for data to end up. Previously we were always running this script from the storage nodes. Now, we are calling this from the RTP execution hosts as part of the file conversion process, so the `hostname` command does not give the result it did previously. This change has been deployed to site and seems to be working.

## Motivation and Context

Fixes a bug observed onsite where the prefix was recorded as `unknown`, which then meant that RTP failed to launch correctly.

## Types of changes

- [ ] Version change
- [ ] Other

## Checklist:

Bug fix checklist:
- [ ] My fix includes a new test that breaks as a result of the bug (if possible).
changes when this is merged.

### DIFF
--- a/scripts/mc_rtp_launch_record.py
+++ b/scripts/mc_rtp_launch_record.py
@@ -46,9 +46,10 @@ for uvfile in args.files:
     # get appropriate filename and prefix information
     filename = os.path.basename(uvfile)
     hostname = socket.gethostname()
-    if "hera-sn1" in hostname:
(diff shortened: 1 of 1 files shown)


bhazelton (approved):
very straightforward.

PR #632, opened 2022-09-22
Add longer sleep to mini catcher

mkolopanis:
have the mini catcher sleep longer. Snap autos don't update very often anyway


plaplant (approved):
I approve of more sleeping, let's give the little guy a chance to rest. Thanks @mkolopanis

PR #639, opened 2022-10-27
Better time filtering to allow for getting the most recent table entry as of a time in the past

bhazelton:
## Description

Currently it is easy to get the next entry after a point in time or the most recent entry as of the current time, but it is hard to get the last entry before a point in time (the most recent as of that time). This PR improves the time filter logic that nearly all the real time db getter methods use to enable getting the most recent entry in a table as of a time in the past.

## Motivation and Context

This came up recently as an issue in the daily notebooks. I put in a work around, but this will simplify the logic there and for other users.

## Types of changes

- [x] New feature without schema change (non-breaking change which adds functionality)
- [ ] Version change
- [ ] Other

## Checklist:

New feature without schema change checklist:
changes when this is merged.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,12 @@ commanded signal source (one of "antenna", "load", "noise", "digital_same_seed",
 had an event related to normal observing (f-engine sync, x-engine integration start,
 catcher start, stop or stop identified via a timeout).
 
+
+### Changed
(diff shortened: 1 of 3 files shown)


mkolopanis (commented):
Looks good to me. Reasonable feature to have.

mkolopanis (approved):

PR #640, opened 2022-10-27
Add auto spectra

bhazelton:
## Description

Add a new table (`hera_auto_spectrum`) with the full autocorrelation spectra (not just the medians which are currently recorded in the `hera_autos` table).

This new table uses an Array type column in postgres and a character type column in SQLite.  In order to make sure this works properly I added testing related to spinning up SQLite databases, which we never did before in our tests. In the process I found and fixed some bugs and made some significant changes to `cm_gen_sqlite.py`.

I also changed the CI to use mamba instead of conda because I was annoyed with how slow it was.

## Motivation and Context

I think this closes an issue somewhere, but I can't find it.

## Types of changes

- [ ] Version change
- [ ] Other

## Checklist:

Schema change:
- [ ] I have tested looping upgrading and downgrading the alembic version and tests pass consistently.
- [x] I have updated or created monitoring scripts and daemons so any new tables will be automatically populated (if appropriate).
changes when this is merged.
- [ ] I have updated the schema documentation document with my changes (under docs/mc_definition.tex)


bhazelton:
@david-deboer do you have ideas around testing that it doesn't error with sqlite? 2 uncovered lines here associated with the support for sqlite.

hera_mc/autocorrelations.py
```diff
@@ -132,3 +133,70 @@ def create(cls, time, antenna_number, antenna_feed_pol, measurement_type, value)
...
+
+    Attributes
+    ----------
+    time : BigInteger Column
+        The time in GPS seconds of the observation, floored as an int. Part of primary_key
+    antenna_number : Integer Column
+        Antenna number. Part of primary_key.
+    antenna_feed_pol : String Column
+        Feed polarization, either 'e' or 'n'. Part of primary_key.
+    spectrum : Float Columnn
```
dannyjacobs, line 150: assumption made here about channel/frequency mapping.  I'm fine with this if you are.

dannyjacobs (approved):
I can't comment on the finer points of the sql array definition, but it looks reasonable to me. We continue here the practice of leaving channel mapping to the wits of the end consumer. But in the interest of time I suggest we test now and debate later.

hera_mc/autocorrelations.py
bhazelton: I'm not sure what you're getting at. That the frequencies aren't in their own column? That would be pretty redundant because they'd be the same for all rows.

mkolopanis: I mostly agree with bryna here, but we have done 8k and 16k channel autos before (as a test if I remember but we still did it). But overall it seems like a bit of wasted space to save it all the time. Even then we don't spit out the frequencies in redis anywhere, I'm pretty sure we'd always have to reconstruct them using a linspace and an average (which is what we'd ask a user to do anyway). We could always just write down the algorithm somewhere convenient for anyone who is interested.

bhazelton: we could add a column for frequency resolution if that helped. I don't actually know where to find that in redis...

david-deboer:
Isn't the only way to hit those two lines is to have a non-postgresql data base (e.g. a test sqlite version in data/test_data) that gets used as the database in the tests?

hera_mc/autocorrelations.py
bhazelton: I added something to the table definition to describe how to get the frequencies. The basic answer is that the bandwidth (0 -> 250 MHz) is fixed, so the frequencies can be computed from that and from the number of entries in the spectrum.

hera_mc/cm_gen_sqlite.py
```diff
@@ -152,8 +169,8 @@ def update_sqlite(self, db_file="hera_mc.db"):
             f.write(inserts)
             f.write(".save {}\n".format(dbfile_full))
         subprocess.call("sqlite3 < {}".format(sqlfile), shell=True)
-        subprocess.call(f"rm -f {schema_file}{testtag}", shell=True)
-        subprocess.call(f"rm -f {inserts_file}{testtag}", shell=True)
+        subprocess.call(f"rm -f {schema_file}", shell=True)
+        subprocess.call(f"rm -f {inserts_file}", shell=True)
         subprocess.call(f"rm -f {sqlfile}", shell=True)
```
mkolopanis, line 174: I get these were always subprocess calls. but why not use like `os.remove` or Pathlib's `unlink` just to remove a file?

mkolopanis (commented):
I have a few extra questions now

hera_mc/cm_gen_sqlite.py
```diff
@@ -114,18 +115,34 @@ def update_sqlite(self, db_file="hera_mc.db"):
...
+                interline = line + ""
+                if "[]" in line:
+                    # want to convert any array column to a character varying column.
+                    # first find the column name (must be a single word)
+                    name_match = re.search(r"[\w]+", line)
+                    column_name = name_match.group(0)
+                    # then search after the column name for the type, which can be
+                    # multiple words and ends in "[]"
+                    type_match = re.search(
+                        r"([a-zA-Z ]+)\[\]", line.split(column_name)[-1]
```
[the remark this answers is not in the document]
bhazelton: I think so? Not totally sure where to check. I tested and verified that this works for real and float arrays.

hera_mc/cm_gen_sqlite.py
bhazelton: done

hera_mc/cm_gen_sqlite.py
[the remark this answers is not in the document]
bhazelton: I guess I could add hyphen and underscore to the expression, I don't think it would cause problems because the item of interest has to end in `"[]"`.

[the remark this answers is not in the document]
mkolopanis: I'm not particularly prone to it honestly. we can just leave it as letters for now and see what happens.

mkolopanis (approved):
Thanks for the extra work on this. Looks good from my end. I don't mind a little hand waving on the sqlite coverage.

PR #649, opened 2023-02-15
Fix compatibility with sqlalchemy2.0

bhazelton:
## Description

SQLAlchemy 2.0 is out and it broke some things. This PR fixes:
- compatibility with SQLAlchemy 2.0
- Some old problems in alembic revisions that prevented upgrades and downgrades from being fully reversible.
- Adds a new CI that runs the tests with the `-W error` flag to catch warnings. As on other repos, I don't think this should be a required check, but it will prevent us missing new deprecation warnings (as we did with the changes in SQLAlchemy 2.0 which were warned about for some time apparently).
- Drop support for python 3.7.
- Fixes conftest and cm_utils.py to make it possible to run tests from another folder. Runs from another folder in the min_deps test to improve coverage in `__init__.py`

## Types of changes

- [ ] Version change
- [ ] Other

## Checklist:

Bug fix checklist:
- [x] My fix includes a new test that breaks as a result of the bug (if possible).
changes when this is merged.

### DIFF
--- a/.github/workflows/publish-to-test-pypi.yml
+++ b/.github/workflows/publish-to-test-pypi.yml
@@ -19,11 +19,12 @@ jobs:
         with:
           fetch-depth: 0
 
-      - name: Setup Miniconda
+      - name: Setup Minimamba
         uses: conda-incubator/setup-miniconda@v2.0.0
         with:
-          auto-update-conda: true
-          miniconda-version: "latest"
+          miniforge-variant: Mambaforge
+          miniforge-version: latest
+          use-mamba: true
           python-version: ${{ env.PYTHON }}
           environment-file: ci/${{ env.ENV_NAME }}.yml
           activate-environment: ${{ env.ENV_NAME }}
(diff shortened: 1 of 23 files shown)


mkolopanis:
do we need to force sqlalchemy version to be >=2?

bhazelton:
This doesn't break compatibility with older versions of sqlalchemy. I am sure of this because a python 3.7 run passed using sqlalchemy 1.4.

However, after I removed `pkg_resources` in favor of `importlib.metadata`, python 3.7 is no longer passing because `importlib.metadata` was introduced in python 3.8. I'm now wondering how important it is to keep supporting python 3.7. Are we using it anywhere onsite?

mkolopanis:
I'm not aware of anywhere on site it is used. I would argue if we have an older environment maybe we should update it. But we'll have to go check all the environments.

bhazelton:
per a conversation on Slack, I removed support for python 3.7

mkolopanis (approved):
Thanks for the updates and checking on site. I think this is working for me.

bhazelton:
I've now tested this onsite without errors, so I'm going to merge.

PR #651, opened 2023-02-23
Prevent alembic autogenerate from dropping deleted tables

### LINKED ISSUE #590: Alembic tweaks to ignore deleted tables
It'd be nice to configure Alembic not to drop tables that we've decided to delete in our code but not on site.

The following links might be useful for figuring this out:
 https://stackoverflow.com/questions/65184035/alembic-ignore-specific-tables
 https://stackoverflow.com/questions/54978681/ignoring-a-model-when-using-alembic-autogenerate

The second link may also provide some breadcrumbs for ignoring some tables when using SQLITE that have datatypes that SQLITE doesn't support (e.g. arrays).

bhazelton:
## Description

Use a solution out of the Alembic cookbook to prevent Alembic from dropping existing tables onsite.

## Motivation and Context

Fixes #590

## Types of changes

- [ ] Version change
- [x] Other

## Checklist:

Other:
changes when this is merged.

I don't think we need to test this onsite or update the changelog.

### DIFF
--- a/alembic/env.py
+++ b/alembic/env.py
@@ -36,6 +36,15 @@ target_metadata = mc.MCDeclarativeBase.metadata
 # ... etc.
 
 
+# Don't drop any tables.
+# See https://alembic.sqlalchemy.org/en/latest/cookbook.html#don-t-generate-any-drop-table-directives-with-autogenerate
+def include_object(object, name, type_, reflected, compare_to):  # noqa A002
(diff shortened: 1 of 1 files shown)


bhazelton:
With this change, if I make an autogenerated version on main I unfortunately still get a new version file, it just has "pass" in both the `upgrade` and `downgrade` functions. But I think we can live with that.

mkolopanis (approved):

PR #660, opened 2023-10-07
Add a backstop to file conversion before launching RTP

jsdillon:
This PR adds a step in the RTP launch script as a last-ditch backstop for fileconversion. Over the past week, I've noticed a least three different nights where one or a handful of files fail to convert (often the .diff or .sum but not the other). This script finds those unconverted files and tries one last time to convert them before launching RTP. 

This PR also adds deletion of .dat files that have been successfully converted. This is useful for saving space on /mnt/sn1 in the case where RTP gets backed up (since .dat files are roughly double the size of .uvh5 files)

### DIFF
--- a/scripts/mc_launch_rtp_daemon.py
+++ b/scripts/mc_launch_rtp_daemon.py
@@ -11,6 +11,7 @@ included in a launched workflow, the corresponding RTPLaunchRecord is updated.
 """
 
 import os
+import glob
 import shutil
 import subprocess
 import warnings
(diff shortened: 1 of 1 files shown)


jsdillon:
@AaronParsons this PR is mostly for you, but I'd also love @bhazelton to check if there's something else that should be done to make sure the newly converted files are properly recorded in M&C.

bhazelton:
I'm not totally sure (I didn't write this part of M&C) but I read it through and it _seems_ like it'll be ok.

PR #664, opened 2024-03-19
more properly set phase info for the snap auto UVData object

mkolopanis:
updates the snap auto catcher script to more properly set the phase center catalog. Was exploiting some old phase code before.


bhazelton (approved):
Looks good, thanks!

PR #675, opened 2025-04-02
various fixes from site

bhazelton:
## Description

Various fixes that have lived for some time on site but never been pushed up to GitHub.

## Motivation and Context

local changes on site are bad!

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] Schema change (any change to the SQL tables)
- [ ] New feature without schema change (non-breaking change which adds functionality)
- [ ] Change associated with a change in redis structure
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Version change
- [ ] Build or continuous integration change
- [x] Other

## Checklist:

Bug fix checklist:
- [x] My code follows the code style of this project.
- [x] My fix includes a new test that breaks as a result of the bug (if possible).
- [x] I understand the updates required onsite (detailed in the readme) and I will make those
changes when this is merged.
- [x] Unit tests pass **on site** (This is a critical check, CI can differ from site).
- [ ] I have updated the [CHANGELOG](https://github.com/HERA-Team/hera_mc/blob/main/CHANGELOG.md).

Other:
- [x] My code follows the code style of this project.
- [x] I understand the updates required onsite (detailed in the readme) and I will make those
changes when this is merged.
- [x] Unit tests pass **on site** (This is a critical check, CI can differ from site).
- [ ] I have updated the [CHANGELOG](https://github.com/HERA-Team/hera_mc/blob/main/CHANGELOG.md).

PR #677, opened 2025-09-19
use pyproj.toml, src layout

bhazelton:
## Description

This is a modernization/reorganization to a src layout (allows better testing) and using pyproj.toml instead of setup.py to define the package (the current setup.py approach is deprecated).

I actually did this work several months ago but was waiting for #675 to be merged to get this in.

## Types of changes

- [ ] Version change
- [x] Build or continuous integration change
- [x] Other

## Checklist:

Build or continuous integration change checklist:
- [x] If required or optional dependencies have changed (including version numbers), I have updated the readme to reflect this.

Other:
changes when this is merged.


david-deboer (dismissed):
Looks fine to me.  The only question is the url.  postgresql -> postgresql+psycopg.  I'm assuming that is correct and then not related to the CI having issues?

bhazelton:
> Looks fine to me. The only question is the url. postgresql -> postgresql+psycopg. I'm assuming that is correct and then not related to the CI having issues?

I think it was a mistake. That change should only be made for sqlalchemy access to postgres, not to the actual postgres server. I fixed it and it's making it past that step (there's a new error I'm working on). Thanks for pointing it out!

.coveragerc
```diff
@@ -1,5 +1,21 @@
 [run]
-omit = */tests/*
+omit =
+    */tests/*
+    */docs/*
+branch = false
```
bhazelton: punted turning this on, made an issue for now (#679)

